PR #10, opened 2019-10-16
Feature/ab2d 191 data kickoff

adaykin:
Implement the FHIR spec of the endpoint GET [fhir base]/Patient/$export

### Summary

Implement the method that will be used for the endpoint, as well as add several services and repositories for interacting with the data. Adding tests for the export to make sure that data is valid. Add properties file for testing so that app can run with H2 for tests.

### Checklist

- [x] Tests and linting pass

### Security

No PII or PHI involved in this part. Auth is still being figured out.


api/pom.xml
```diff
@@ -13,6 +13,7 @@
     <properties>
         <project.root>${basedir}/..</project.root>
         <springfox.version>2.9.2</springfox.version>
+        <springboot.version>2.1.6.RELEASE</springboot.version>
```
dkrylovsb: @adaykin Let's declare this property in the top level `pom`, since that is where the `spring-boot-starter-parent` is declared. The property will propagate to the modules automatically.

api/pom.xml
```diff
@@ -35,6 +40,18 @@
...
             <version>${springfox.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <version>${springboot.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
dkrylovsb: Unless this dependency is used in this PR, let's take it out and add back in once we actually need it.

api/src/main/java/gov/cms/ab2d/api/SpringBootApp.java
```diff
@@ -2,9 +2,10 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-
+import org.springframework.boot.autoconfigure.domain.EntityScan;
 
 @SpringBootApplication
+@EntityScan(basePackages = {"gov.cms.ab2d.domain"})
```
dkrylovsb: `@SpringBootApplication` alone should be sufficient to find the entity classes. is the app not working without EntityScan?

api/src/main/java/gov/cms/ab2d/api/auth/AuthService.java
```diff
@@ -0,0 +1,39 @@
+package gov.cms.ab2d.api.auth;
```
dkrylovsb: AuthN/Z is out of scope of this task; we can leave this class out.

api/src/main/java/gov/cms/ab2d/api/config/ErrorHandler.java
```diff
@@ -0,0 +1,30 @@
+package gov.cms.ab2d.api.config;
```
dkrylovsb: We should probably keep this class in `gov.cms.ab2d.api.controller`

api/pom.xml
adaykin: Ok. Would it be excluded from the domain module since I don't believe it would be needed there?

api/src/main/java/gov/cms/ab2d/api/SpringBootApp.java
adaykin: Yeah, I believe it's because the entities are in a different directory, if I don't have this it will give an error on startup like:
Caused by: java.lang.IllegalArgumentException: Not a managed type: class gov.cms.ab2d.domain.User
	at org.hibernate.metamodel.internal.MetamodelImpl.managedType(MetamodelImpl.java:552) ~[hibernate-core-5.3.10.Final.jar:5.3.10.Final]

api/pom.xml
adaykin: I thought this was needed but I guess it's not, I can remove

api/pom.xml
adaykin: Actually I guess this is just the property so it's fine.

api/src/main/java/gov/cms/ab2d/api/SpringBootApp.java
dkrylovsb: 👍

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -50,9 +68,17 @@
                     allowableValues = ALLOWABLE_OUTPUT_FORMATS, defaultValue = "application/fhir" +
                     "+ndjson"
             )
-            @RequestParam(required = false, name = "_outputFormat") String outputFormat) throws IOException {
-        //String encoded = FHIRUtil.outcomeToJSON(FHIRUtil.getSuccessfulOutcome("OK"));
-        return new ResponseEntity<>(null, null,
+            @RequestParam(required = false, name = "_outputFormat") String outputFormat) {
+        //activeRequestCheck();
+
```
dkrylovsb: Before we actually create a Job, let's apply validation to the request parameters. Let's make sure `_outputFormat` is one of the three accepted values and return 400.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -50,9 +68,17 @@
...
+        Job job = jobService.createJob(resourceTypes, ServletUriComponentsBuilder.fromCurrentRequest().toUriString());
+
+        String statusURL = ServletUriComponentsBuilder.fromCurrentRequestUri().replacePath
+                (String.format(API_PREFIX + "/status/%s", job.getJobID())).toUriString();
```
dkrylovsb: Status URL isn't correct (it _was_ in the PDP coding challenge): status check is at `/Job/{jobId}/$status`

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -108,5 +134,23 @@
...
 
+    /*private void activeRequestCheck() {
+        User user = userService.getCurrentUser();
+        if (jobService.getActiveJob(user) != null) {
+            throw new RuntimeException("The current user has an active job");
+        }
+    }*/
+
+    @ControllerAdvice
+    class ErrorHandler extends ResponseEntityExceptionHandler {
```
dkrylovsb: Let's place `ErrorHandler` in its own class file. Inner classes are kind of hard to find when looking at source tree

api/src/main/java/gov/cms/ab2d/api/service/JobService.java
```diff
@@ -0,0 +1,49 @@
...
+    public Job createJob(String resourceTypes, String url) {
+        Job job = new Job();
+        job.setResourceTypes(resourceTypes);
+        job.setJobID(UUID.randomUUID().toString());
+        job.setRequestURL(url);
+        job.setStatus(JobStatus.SUBMITTED);
+        job.setStatusMessage(INITIAL_JOB_STATUS_MESSAGE);
+        job.setCreatedAt(LocalDateTime.now());
+        job.setUser(userService.getCurrentUser());
+
```
dkrylovsb: Need to set `progress` at zero

api/src/main/java/gov/cms/ab2d/api/service/JobService.java
```diff
@@ -0,0 +1,49 @@
...
+        return jobRepository.save(job);
+    }
+
+    public Job getActiveJob(User user) {
```
dkrylovsb: Not being used for this PR; can remove.

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -0,0 +1,84 @@
...
+        Assert.assertEquals(job.getStatus(), JobStatus.SUBMITTED);
+        Assert.assertEquals(job.getStatusMessage(), INITIAL_JOB_STATUS_MESSAGE);
+        Assert.assertEquals(job.getProgress(), null);
+        Assert.assertEquals(job.getRequestURL(), "http://localhost" + API_PREFIX + PATIENT_EXPORT_PATH + typeParams);
+        Assert.assertEquals(job.getResourceTypes(), "Patient,ExplanationOfBenefits");
+        Assert.assertEquals(job.getUser(), null);
+    }
+
+    @Test
+    public void testPatientExportWithInvalidParameters() throws Exception {
```
dkrylovsb: This test case needs to verify `OperationOutcome` is returned in body

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -0,0 +1,84 @@
...
+                .andExpect(header().string("Content-Location", statusUrl));
+
+        Assert.assertEquals(job.getStatus(), JobStatus.SUBMITTED);
+        Assert.assertEquals(job.getStatusMessage(), INITIAL_JOB_STATUS_MESSAGE);
+        Assert.assertEquals(job.getProgress(), null);
+        Assert.assertEquals(job.getRequestURL(), "http://localhost" + API_PREFIX + PATIENT_EXPORT_PATH + typeParams);
+        Assert.assertEquals(job.getResourceTypes(), "Patient,ExplanationOfBenefits");
+        Assert.assertEquals(job.getUser(), null);
+    }
+
```
dkrylovsb: Will need a test case to check for invalid `_outputFormat`

api/src/main/java/gov/cms/ab2d/api/service/UserService.java
```diff
@@ -0,0 +1,23 @@
...
+import gov.cms.ab2d.domain.User;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.Authentication;
+import org.springframework.stereotype.Service;
+
+@Service
+/**
+ * Just gets the current user from the authentication context.
+ */
+public class UserService {
```
dkrylovsb: Similar comment about extracting interface

domain/src/main/java/gov/cms/ab2d/domain/Job.java
```diff
@@ -36,6 +38,22 @@
     private JobStatus status;
     private String statusMessage;
     private Integer progress;
+
+    @Pattern(regexp = "^(((Patient)|(ExplanationOfBenefits)),?)*$",
+            message = "_type should contain a string of comma-delimited FHIR resource types; "
+                    + "currently limited to Patient and ExplanationOfBenefits")
     private String resourceTypes;
 
+    @Override
```
dkrylovsb: We're using Project Lombok to avoid having to declare getters, setters, hashCode, equals, etc explicitly. Let's remove these two methods and then use `@EqualsAndHashCode(onlyExplicitlyIncluded = true)` from Lombok instead. Annotate `id` with `@EqualsAndHashCode.Include`

domain/src/main/java/gov/cms/ab2d/domain/User.java
```diff
@@ -17,11 +19,25 @@
 
     @Column(unique = true)
     private String userID;
+    private String passwordHash;
```
dkrylovsb: We won't store passwords; we can drop this field.

domain/src/main/java/gov/cms/ab2d/domain/User.java
```diff
@@ -17,11 +19,25 @@
...
     private String name;
     private String email;
     private Boolean accountDisabled;
+    private Date expire;
```
dkrylovsb: also not needed.

domain/src/main/java/gov/cms/ab2d/domain/User.java
```diff
@@ -17,11 +19,25 @@
...
 
     @ManyToOne
     private Sponsor sponsor;
 
+    @Override
```
dkrylovsb: Same Lombok comment

api/src/main/java/gov/cms/ab2d/api/service/JobService.java
```diff
@@ -0,0 +1,49 @@
...
+    private UserService userService;
+
+    @Autowired
+    private JobRepository jobRepository;
+
+    public static final String INITIAL_JOB_STATUS_MESSAGE = "0%";
+
+    public Job createJob(String resourceTypes, String url) {
+        Job job = new Job();
+        job.setResourceTypes(resourceTypes);
```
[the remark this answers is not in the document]
adaykin: That pattern is in the job class now, I copied it over from the coding challenge.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -1,31 +1,44 @@
...
 /**
  * The sole REST controller for AB2D's implementation of the FHIR Bulk Data API specification.
  */
 public class BulkDataAccessAPI {
 
-    private static final String ALLOWABLE_OUTPUT_FORMATS =
-            "application/fhir+ndjson,application/ndjson,ndjson";
+    private static final Set<String> ALLOWABLE_OUTPUT_FORMAT_SET = Set.of("application/fhir+ndjson", "application/ndjson", "ndjson");
+
+    // Since this is used in an annotation, it can't be derived from the Set, otherwise it will be an error
```
dkrylovsb: True. But you could derive the Set above from this constant to avoid duplication. Something like
```
Set.<String[]>of(ALLOWABLE_OUTPUT_FORMATS.split(","))
```

pom.xml
```diff
@@ -25,6 +25,7 @@
         <project.root>${basedir}</project.root>
         <java.version>12</java.version>
         <checkstyle.config>${project.root}/src/main/resources/checkstyle.xml</checkstyle.config>
+        <springboot.version>2.1.6.RELEASE</springboot.version>
```
dkrylovsb: The parent declaration is still specifying the same version explicitly, so we have the version number duplicated in two places. Let's see if we could change the parent declaration to use the variable instead:

```
<parent>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-starter-parent</artifactId>
        <version>2.1.6.RELEASE</version>
        <relativePath/> <!-- lookup parent from repository -->
    </parent>```

adaykin: I had tried to use that, unfortunately maven doesn't allow for properties to be used in parent definitions.

PR #16, opened 2019-10-18
Feature/ab2d 192 delete request

adaykin:
Implementing 5.2 spec for FHIR for cancelling a job

Adding code for the deleteJob API call so that a job can have the status updated to cancelled 

### Checklist

- [x] Tests and linting pass

### Security
N/A

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
@@ -2,6 +2,7 @@ package gov.cms.ab2d.api.controller;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import gov.cms.ab2d.api.service.InvalidJobStateTransition;
 import gov.cms.ab2d.api.service.ResourceNotFoundException;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.hl7.fhir.dstu3.model.OperationOutcome;
@@ -31,7 +32,7 @@ class ErrorHandler extends ResponseEntityExceptionHandler {
     }
 
     // Most errors that are the fault of the client
-    @ExceptionHandler({IllegalArgumentException.class, IllegalStateException.class})
+    @ExceptionHandler({InvalidUserInputException.class, InvalidJobStateTransition.class})
     public ResponseEntity<JsonNode> assertionException(final Exception e) throws IOException {
         return generateFHIRError(e, HttpStatus.BAD_REQUEST);
     }
--- a/api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
@@ -31,7 +31,7 @@ class ErrorHandler extends ResponseEntityExceptionHandler {
     }
 
     // Most errors that are the fault of the client
-    @ExceptionHandler(IllegalArgumentException.class)
+    @ExceptionHandler({IllegalArgumentException.class, IllegalStateException.class})
     public ResponseEntity<JsonNode> assertionException(final Exception e) throws IOException {
         return generateFHIRError(e, HttpStatus.BAD_REQUEST);
     }
--- a/api/src/main/java/gov/cms/ab2d/api/service/JobServiceImpl.java
+++ b/api/src/main/java/gov/cms/ab2d/api/service/JobServiceImpl.java
@@ -8,6 +8,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
+import javax.persistence.EntityNotFoundException;
 import java.time.LocalDateTime;
 import java.util.UUID;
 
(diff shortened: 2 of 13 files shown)


api/src/main/java/gov/cms/ab2d/api/service/JobServiceImpl.java
```diff
@@ -36,4 +37,13 @@ public Job createJob(String resourceTypes, String url) {
 
         return jobRepository.save(job);
     }
+
+    public void cancelJob(String jobId) {
+        Job job = jobRepository.findByJobID(jobId);
```
dkrylovsb: Once we have the other PR merged, we should extract job lookup and null check into a common private method, since "getJobByJobId" does the same thing

api/src/main/java/gov/cms/ab2d/api/service/JobServiceImpl.java
```diff
@@ -36,4 +37,13 @@ public Job createJob(String resourceTypes, String url) {
...
+        if (job == null) {
+            throw new EntityNotFoundException("No job with jobID " +  jobId + "  was found");
+        }
+
```
dkrylovsb: We need to throw an exception if a user is trying to cancel a job that's in FAILED or COMPLETE status.

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -98,4 +100,26 @@ public void testPatientExportWithInvalidOutputFormat() throws Exception {
...
+        Job cancelledJob = jobRepository.findByJobID(job.getJobID());
+        Assert.assertEquals(JobStatus.CANCELLED, cancelledJob.getStatus());
+    }
+
+    @Test
+    public void testDeleteNonExistentJob() throws Exception {
+        this.mockMvc.perform(delete(API_PREFIX + "/Job/NonExistentJob/$status"))
+                .andExpect(status().is(404))
+                .andExpect(content().string(JOB_NOT_FOUND_ERROR_MSG));
+    }
```
dkrylovsb: Will need test cases to cover attempts to cancel FAILED or COMPLETED job (see previous comments)

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -98,4 +100,26 @@ public void testPatientExportWithInvalidOutputFormat() throws Exception {
...
+            .andExpect(content().string(JOB_CANCELLED_MSG));
+
+        Job cancelledJob = jobRepository.findByJobID(job.getJobID());
+        Assert.assertEquals(JobStatus.CANCELLED, cancelledJob.getStatus());
+    }
+
+    @Test
+    public void testDeleteNonExistentJob() throws Exception {
+        this.mockMvc.perform(delete(API_PREFIX + "/Job/NonExistentJob/$status"))
+                .andExpect(status().is(404))
```
dkrylovsb: We are required by the spec to return FHIR OperationOutcome in case of 4xx or 5xx, so let's assert this here.

api/src/test/java/gov/cms/ab2d/api/service/JobServiceTest.java
```diff
@@ -48,4 +50,21 @@ public void createRequest() {
...
+
+        jobService.cancelJob(job.getJobID());
+
+        // Verify that it has the correct status
+        Job cancelledJob = jobRepository.findByJobID(job.getJobID());
+
+        assertEquals(JobStatus.CANCELLED, cancelledJob.getStatus());
+    }
+
+    @Test(expected = EntityNotFoundException.class)
```
dkrylovsb: Similarly, will need test cases to cover attempts to cancel FAILED or COMPLETED job (see previous comments)

api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
```diff
@@ -31,7 +31,7 @@
     }
 
     // Most errors that are the fault of the client
-    @ExceptionHandler(IllegalArgumentException.class)
+    @ExceptionHandler({IllegalArgumentException.class, IllegalStateException.class})
```
dkrylovsb: So my concern here is that both these exceptions -- `IllegalArgumentException` and `IllegalStateException` -- are core Java exceptions, which are pretty generic. They can be thrown by other libraries and by Java itself in a variety of situations, to all of which we will react with `BAD_REQUEST`, while in fact the root cause could be something else. Sounds like we need few more our _own_ exceptions.

Say, `InvalidUserInputException` and `InvalidJobStateTransition` exception?

api/src/main/java/gov/cms/ab2d/api/service/JobServiceImpl.java
```diff
@@ -37,6 +37,16 @@ public Job createJob(String resourceTypes, String url) {
         return jobRepository.save(job);
     }
 
+    public void cancelJob(String jobID) {
+        Job job = getJobByJobID(jobID);
+
+        if (job.getStatus() == JobStatus.CANCELLED || job.getStatus() == JobStatus.SUCCESSFUL || job.getStatus() == JobStatus.FAILED) {
```
dkrylovsb: We can make this cleaner by adding `isCancellable` attribute to the `JobStatus` enum and checking that instead.

api/src/test/java/gov/cms/ab2d/api/service/JobServiceTest.java
```diff
@@ -5,6 +5,7 @@
 import gov.cms.ab2d.domain.Job;
 import gov.cms.ab2d.domain.JobOutput;
 import gov.cms.ab2d.domain.JobStatus;
+import org.hibernate.collection.internal.PersistentBag;
```
dkrylovsb: Is this import needed?

adaykin: No, not sure why the build didn't catch it

PR #17, opened 2019-10-21
Adding the bulk status API call of the FHIR spec

adaykin:
Adding the bulk status API call of the FHIR spec

Adding the bulk status API call and returning appropriate status codes so that the client can have an understanding of what the status of the job is. Adding exception handling for different sorts of errors so that there will be less boilerplate code and that it's easier to map the types of exceptions to status codes. 

### Checklist

- [x] Tests and linting pass

### Security
N/A

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
@@ -134,18 +134,25 @@ public class BulkDataAccessAPI {
         jobService.updateJob(job);
 
         HttpHeaders responseHeaders = new HttpHeaders();
-        if (job.getProgress().equals(100)) {
-            responseHeaders.add("Expires", DateUtil.formatLocalDateTimeAsUTC(job.getExpires()));
-            JobCompletedResponse resp = new JobCompletedResponse();
-            resp.setTransactionTime(new DateTimeType(convertLocalDateTimeToDate(job.getCompletedAt())).toHumanDisplay());
-            resp.setRequest(job.getRequestURL());
-            resp.setOutput(job.getJobOutput().stream().filter(o -> !o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), o.getFilePath())).collect(Collectors.toList()));
-            resp.setError(job.getJobOutput().stream().filter(o -> o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), o.getFilePath())).collect(Collectors.toList()));
-            return new ResponseEntity<>(new ObjectMapper().valueToTree(resp), responseHeaders, HttpStatus.OK);
-        } else {
-            responseHeaders.add("X-Progress", job.getProgress() + "% complete");
-            responseHeaders.add("Retry-After", Integer.toString(retryAfterDelay));
-            return new ResponseEntity<>(null, responseHeaders, HttpStatus.ACCEPTED);
+        switch (job.getStatus()) {
+            case SUCCESSFUL:
+                responseHeaders.add("Expires", DateUtil.formatLocalDateTimeAsUTC(job.getExpires()));
+                JobCompletedResponse resp = new JobCompletedResponse();
+                resp.setTransactionTime(new DateTimeType(convertLocalDateTimeToDate(job.getCompletedAt())).toHumanDisplay());
+                resp.setRequest(job.getRequestURL());
+                resp.setRequiresAccessToken(true);
+                resp.setOutput(job.getJobOutput().stream().filter(o -> !o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), o.getFilePath())).collect(Collectors.toList()));
+                resp.setError(job.getJobOutput().stream().filter(o -> o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), o.getFilePath())).collect(Collectors.toList()));
+                return new ResponseEntity<>(new ObjectMapper().valueToTree(resp), responseHeaders, HttpStatus.OK);
+            case SUBMITTED:
+                IN_PROGRESS:
+                responseHeaders.add("X-Progress", job.getProgress() + "% complete");
+                responseHeaders.add("Retry-After", Integer.toString(retryAfterDelay));
+                return new ResponseEntity<>(null, responseHeaders, HttpStatus.ACCEPTED);
+            case FAILED:
+                throw new JobProcessingException("Job failed while processing");
+            default:
+                throw new RuntimeException("Unknown status of job");
         }
     }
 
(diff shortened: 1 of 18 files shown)


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -109,11 +120,33 @@
...
+            throw new TooManyRequestsException("You are polling too frequently");
+        }
+
+        job.setLastPollTime(now);
+        jobService.updateJob(job);
+
         HttpHeaders responseHeaders = new HttpHeaders();
-        responseHeaders.add("X-Progress", "0%");
-        return new ResponseEntity<>(null, responseHeaders, HttpStatus.ACCEPTED);
+        if (job.getProgress().equals(100)) {
```
dkrylovsb: We should be checking the `status` field.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -109,11 +120,33 @@
...
+            responseHeaders.add("Expires", DateUtil.formatLocalDateTimeAsUTC(job.getExpires()));
+            JobCompletedResponse resp = new JobCompletedResponse();
+            resp.setTransactionTime(new DateTimeType(convertLocalDateTimeToDate(job.getCompletedAt())).toHumanDisplay());
+            resp.setRequest(job.getRequestURL());
+            resp.setOutput(job.getJobOutput().stream().filter(o -> !o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), o.getFilePath())).collect(Collectors.toList()));
+            resp.setError(job.getJobOutput().stream().filter(o -> o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), o.getFilePath())).collect(Collectors.toList()));
+            return new ResponseEntity<>(new ObjectMapper().valueToTree(resp), responseHeaders, HttpStatus.OK);
```
dkrylovsb: We will need to set `requiresAccessToken` to `true` as well.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -109,11 +120,33 @@
...
+        } else {
```
dkrylovsb: We should return this _only_ if the job is in `SUBMITTED` or `IN_PROGRESS` status

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -109,11 +120,33 @@
...
+            responseHeaders.add("X-Progress", job.getProgress() + "% complete");
+            responseHeaders.add("Retry-After", Integer.toString(retryAfterDelay));
+            return new ResponseEntity<>(null, responseHeaders, HttpStatus.ACCEPTED);
+        }
```
dkrylovsb: Need to handle the `FAILED` job status and return 500 with an OperationOutcome as described in the spec, e.g.
```
Status: 500 Internal Server Error
Content-Type: application/json

{
 "resourceType": "OperationOutcome",
 "id": "1",
 "issue": [
  {
   "severity": "error",
   "code": "processing",
   "details": {
    "text": "An internal timeout has occurred"
   }
  }
 ]
}```

File: api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java [hunk 5]
```diff
@@ -19,12 +21,41 @@
...
+        return generateFHIRError(e, HttpStatus.INTERNAL_SERVER_ERROR);
+    }
+
+    // Most errors that are the fault of the client
+    @ExceptionHandler(IllegalArgumentException.class)
     public ResponseEntity<JsonNode> assertionException(final Exception e) throws IOException {
+        return generateFHIRError(e, HttpStatus.BAD_REQUEST);
+    }
+
+    @ExceptionHandler(EntityNotFoundException.class)
```
dkrylovsb: This is JPA-specific exception, which makes the service class a somewhat "leaky" abstraction, i.e. persistence-related stuff leaks to controllers. Let's just introduce our own exception class, just like you did with TooManyRequests: e.g. `ResourceNotFoundException`

api/src/main/java/gov/cms/ab2d/api/repository/JobRepository.java
```diff
@@ -7,4 +7,5 @@
 @Repository
```
dkrylovsb: I just realized that this repository is in the `api` module. This needs to be moved to the `domain` module, since both the API and Worker will use this repository.

api/src/main/java/gov/cms/ab2d/api/service/JobService.java
```diff
@@ -5,4 +5,8 @@
 public interface JobService {
```
dkrylovsb: Same. Needs to go into `domain`

api/src/main/java/gov/cms/ab2d/api/service/JobServiceImpl.java
```diff
@@ -36,4 +36,17 @@ public Job createJob(String resourceTypes, String url) {
```
dkrylovsb: Also into `domain`

dkrylovsb: In that case we would have a design problem. JobService won't need to access any API- or Worker-specific services.

api/src/main/java/gov/cms/ab2d/api/repository/JobRepository.java
adaykin: Discussed over slack and this can be completed in a separate jira to refactor code to go into the shared module.

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -98,4 +106,85 @@ public void testPatientExportWithInvalidOutputFormat() throws Exception {
...
+
+        jobRepository.saveAndFlush(job);
+
+        this.mockMvc.perform(get(statusUrl).contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().is(500))
+                .andExpect(jsonPath("$.resourceType", Is.is("OperationOutcome")))
+                .andExpect(jsonPath("$.issue[0].severity", Is.is("error")))
+                .andExpect(jsonPath("$.issue[0].code", Is.is("invalid")))
+                .andExpect(jsonPath("$.issue[0].details.text", Is.is("JobProcessingException: Job failed while processing")));
+    }
```
dkrylovsb: Need a test for 404 (job not found). Also need a test for an empty Job ID passed in the URl

adaykin: Added in those tests

PR #22, opened 2019-10-24
Adding 5.4 spec of FHIR for downloading a file

adaykin:
Implementing section 5.4 of FHIR spec

Implementing the method in the controller to take the file parameter and find the file on the filesystem, which will then be downloaded to the client in an ndjson format. Testing to check for whether the file exists or not and when the file is downloaded if it is what is expected.

### Checklist

- [x] Tests and linting pass

### Security

NA

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
@@ -3,6 +3,7 @@ package gov.cms.ab2d.api.controller;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import gov.cms.ab2d.api.service.JobService;
+import gov.cms.ab2d.api.service.ResourceNotFoundException;
 import gov.cms.ab2d.api.util.Constants;
 import gov.cms.ab2d.api.util.DateUtil;
 import gov.cms.ab2d.domain.Job;
@@ -11,6 +12,7 @@ import org.hl7.fhir.dstu3.model.DateTimeType;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.core.io.Resource;
+import org.springframework.core.io.UrlResource;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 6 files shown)


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -48,6 +53,9 @@
     @Value("${api.retry-after.delay}")
     private int retryAfterDelay;
 
+    @Value("${api.file-download.path}")
```
dkrylovsb: This will move to the service method. (more to follow)

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -170,8 +178,15 @@
     public ResponseEntity<Resource> downloadFile(
             @ApiParam(value = "A job identifier", required = true) @PathVariable @NotBlank String jobId,
             @ApiParam(value = "A file name", required = true) @PathVariable @NotBlank String filename) throws IOException {
-        HttpHeaders responseHeaders = new HttpHeaders();
-        responseHeaders.add("X-Progress", "0%");
-        return new ResponseEntity<>(null, responseHeaders, HttpStatus.ACCEPTED);
+        Path file = Paths.get(fileDownloadPath + filename);
```
dkrylovsb: Getting a job's output file is more involved. We will need a JobService method to encapsulate that logic: JobService.getJobOutput(jobID, fileName) returns `Resource`. Explanation to follow.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
dkrylovsb: This is a common property used by both the API and Worker, and it will point at the shared file system mount. So let's name it `efs.mount`

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
dkrylovsb: We can't really just take the `filename` provided by the user and concatenate, but the user could send us anything (security risk). We need to validate and make sure the `filename` in fact matches one the job's output precisely.

api/src/main/resources/application.properties
```diff
@@ -1 +1,4 @@
-api.retry-after.delay=30
\ No newline at end of file
+api.retry-after.delay=30
+
+# This will need to be updated when a production path is finalized
+api.file-download.path=/src/resources
```
dkrylovsb: Will go into `application.properties` under `domain`

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -242,4 +244,71 @@ public void testGetStatusWithBadUrl() throws Exception {
...
+        job.setProgress(100);
+        LocalDateTime expireDate = LocalDateTime.now().plusDays(100);
+        job.setExpires(expireDate);
+        LocalDateTime now = LocalDateTime.now();
+        job.setCompletedAt(now);
+
+        JobOutput jobOutput = new JobOutput();
+        jobOutput.setFhirResourceType("ExplanationOfBenefits");
+        jobOutput.setJob(job);
+        jobOutput.setFilePath("http://localhost" + API_PREFIX + "/Job/" + job.getJobID() + "/file/test.ndjson");
```
dkrylovsb: JobOutput will never contain a full URL by the way, it should only contain a relative path to the file (relative to the EFS mount). The controller is responsible for constructing the URL itself. So in this particular example the file name should be just `test.ndjson`.

PR #23, opened 2019-10-25
Rename user table to user_account table

spathiyil:
Rename user table to user_account table as user is a reserved keyword in PG SQL

### Summary

`user` is a reserved keyword in postgres and SQL so renaming the table to remove any ambiguity and awkwardness while querying the table.
https://www.postgresql.org/docs/10/sql-keywords-appendix.html

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/domain/src/main/java/gov/cms/ab2d/domain/User.java
+++ b/domain/src/main/java/gov/cms/ab2d/domain/User.java
@@ -7,7 +7,7 @@ import lombok.Setter;
 import javax.persistence.*;
 
 @Entity
-@Table(name = "`user`")
+@Table(name = "user_account")
 @Getter
 @Setter
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)


domain/src/main/java/gov/cms/ab2d/domain/User.java
```diff
@@ -7,7 +7,7 @@
 import javax.persistence.*;
 
 @Entity
-@Table(name = "`user`")
+@Table(name = "user_account")
 @Getter
 @Setter
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
```
spathiyil: I only renamed the table. 
@dkrylovsb, Should I have renamed the entity to `UserAccount` as well to match the new table name?

dkrylovsb: Entity can stay I guess

spathiyil: Okay.  Will leave the entity name unchanged.

PR #24, opened 2019-10-25
Feature/ab2d 270 refactor common classes

adaykin:
Moving the code for common libraries to the common module

Renamed the domain module to common, setup packages that make sense for this library, added several scans so that Spring can pickup the appropriate beans, moved common code to the common module.

### Checklist

- [x] Tests and linting pass

### Security

N/A


common/pom.xml
```diff
@@ -23,12 +23,19 @@
...
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-security</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
dkrylovsb: `web` should probably be in API only, since `common` does not deal with REST controllers.

common/src/test/java/gov/cms/ab2d/common/SpringBootApp.java
```diff
@@ -0,0 +1,16 @@
+package gov.cms.ab2d.common;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+
+@SpringBootApplication(scanBasePackages = "gov.cms.ab2d.common")
```
dkrylovsb: Can you try just the `SpringBootApplication` annotation _and_ without `scanBasePackages`? package names now match, so auto-config should work.

common/pom.xml
adaykin: Before I put this in, it was failing on the JobServiceTest that expected a TransactionSystemException, perhaps it's a transitive dependency of web, or something with how web integrates with JPA for handling some of the @Pattern validation?

common/src/test/java/gov/cms/ab2d/common/SpringBootApp.java
adaykin: Yeah the auto config works now

pom.xml
```diff
@@ -129,6 +129,12 @@
                                     </limits>
                                 </rule>
                             </rules>
+                            <excludes>
+                                <!-- These are reporting false positives that are already being tested elsewhere -->
```
dkrylovsb: Understand the domain & SpringBootApp excludes, but excluding everything under `controller` might be too much: there is code in there that needs coverage checks. Let's find out what's going on.

Also, is there a way to specify the excludes within individual Maven modules?

common/pom.xml
adaykin: Actually doing just this works:
<dependency>
      <groupId>org.springframework.boot</groupId>
      <artifactId>spring-boot-starter-validation</artifactId>
      <version>2.2.0.RELEASE</version>
      <scope>compile</scope>
      <exclusions>
        <exclusion>
          <artifactId>tomcat-embed-el</artifactId>
          <groupId>org.apache.tomcat.embed</groupId>
        </exclusion>
      </exclusions>
    </dependency>

adaykin: Github removed the brakcets of the XML from above.

pom.xml
adaykin: I think I actually just wanted the controller class itself since there's an unimplemented method in there, I'm not sure if there's a way to just exclude individual methods.

PR #27, opened 2019-10-25
Create the UserRole ManyToMany association

spathiyil:
Create the UserRole ManyToMany association between a User and their Roles

### Summary

Create the Role entity
Add a `@ManyToMany` association between `User` entity and `Role` entity.

Reference used:
https://vladmihalcea.com/the-best-way-to-use-the-manytomany-annotation-with-jpa-and-hibernate/

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


domain/src/main/java/gov/cms/ab2d/domain/User.java
```diff
@@ -27,4 +40,23 @@
...
+
+    @ManyToMany(cascade = { PERSIST, MERGE })
+    @JoinTable(
+            name = "user_role",
+            joinColumns = @JoinColumn(name = "user_account_id"),
+            inverseJoinColumns = @JoinColumn(name = "role_id"))
+    private Set<Role> roles = new HashSet<>();
+
+
+    public void addRole(Role role) {
```
dkrylovsb: Next two methods will change based on the previous comment

domain/src/main/java/gov/cms/ab2d/domain/Role.java
```diff
@@ -0,0 +1,33 @@
...
+    @Id
+    @GeneratedValue
+    private Long id;
+
+    @NaturalId
+    @EqualsAndHashCode.Include
+    @Column(nullable = false, unique = true)
+    private String name;
+
+    @ManyToMany(mappedBy = "roles")
```
dkrylovsb: I don't anticipate us needing to ever navigate from Role to Users based on the requirements. We can remove this.

spathiyil: Removed the bi-directional association back to Set<User> entity.

domain/src/main/java/gov/cms/ab2d/domain/User.java
spathiyil: done

PR #28, opened 2019-10-28
Using common library to check for job output and see if it matches fi…

adaykin:
…les, so that file can be downloaded and checked for correct output

Implementing 5.4 of FHIR spec so that users can download a file produced from job output.

Implementing the method in the controller to take the file parameter and find the file on the filesystem, which will then be downloaded to the client in an ndjson format. Testing to check for whether the file exists or not and when the file is downloaded if it is what is expected. Also test to make sure that the job output matches up with what is in our system so that we know whether or not the client is requesting output that they have permission to.

### Checklist

- [x] Tests and linting pass

### Security

Make sure that the output the user requests is something that is owned by the job being requested.


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -166,8 +167,12 @@
                 resp.setTransactionTime(jobCompletedAt.toHumanDisplay());
                 resp.setRequest(job.getRequestURL());
                 resp.setRequiresAccessToken(true);
-                resp.setOutput(job.getJobOutput().stream().filter(o -> !o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), o.getFilePath())).collect(Collectors.toList()));
-                resp.setError(job.getJobOutput().stream().filter(o -> o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), o.getFilePath())).collect(Collectors.toList()));
+
+                String requestURIString = ServletUriComponentsBuilder.fromCurrentRequestUri().toUriString();
+                String urlPath = requestURIString.substring(0, requestURIString.indexOf(API_PREFIX)) + API_PREFIX + "/Job/" + job.getJobID() + "/file/";
+
+                resp.setOutput(job.getJobOutput().stream().filter(o -> !o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), urlPath + o.getFilePath())).collect(Collectors.toList()));
```
dkrylovsb: ...then call that method from here

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -166,8 +167,12 @@
...
+                resp.setError(job.getJobOutput().stream().filter(o -> o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), urlPath + o.getFilePath())).collect(Collectors.toList()));
```
dkrylovsb: ...and here.

common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -56,7 +69,37 @@ public Job getJobByJobID(String jobID) {
...
+                jobOutputMatchesPath = true;
+                break;
+            }
+        }
+
+        if (!jobOutputMatchesPath) {
+            throw new ResourceNotFoundException("No Job Output with the file name " + fileName + " exists in our records");
+        }
+
+        Path file = Paths.get(fileDownloadPath + fileName);
```
dkrylovsb: Has to be `fileDownloadPath+ jobId + FileUtils.pathSeparator + fileName`. This is because each job will get its own sub-directory under the EFS root.

common/src/test/java/gov/cms/ab2d/common/service/JobServiceTest.java
```diff
@@ -164,11 +173,120 @@ public void updateJob() {
...
+        errorJobOutput.setFilePath("error.ndjson");
+        errorJobOutput.setJob(job);
+
+        List output = List.of(jobOutput, errorJobOutput);
+        job.setJobOutput(output);
+
+        Job updatedJob = jobService.updateJob(job);
+
+        Resource resource = jobService.getResourceForJob(updatedJob.getJobID(), "test.ndjson");
+        Assert.assertEquals("test.ndjson", resource.getFilename());
```
dkrylovsb: For completeness sake, let's add an assertion for `error.ndjson` as well

common/src/test/java/gov/cms/ab2d/common/service/JobServiceTest.java
```diff
@@ -164,11 +173,120 @@ public void updateJob() {
...
+        jobOutput.setJob(job);
+
+        JobOutput errorJobOutput = new JobOutput();
+        errorJobOutput.setError(true);
+        errorJobOutput.setFhirResourceType(OPERATION_OUTCOME);
+        errorJobOutput.setFilePath("error.ndjson");
+        errorJobOutput.setJob(job);
+
+        List output = List.of(jobOutput, errorJobOutput);
+        job.setJobOutput(output);
```
dkrylovsb: We should try and keep our tests clean, too. These lines duplicate the previous test case. Let's extract them into a common private method within this test class.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -186,6 +183,12 @@
         }
     }
 
+    private String getUrlPath(Job job, String filePath) {
+        String requestURIString = ServletUriComponentsBuilder.fromCurrentRequestUri().toUriString();
```
dkrylovsb: Use `ServletUriComponentsBuilder.fromCurrentRequestUri().replacePath` method. It's cleaner and more flexible.

common/src/main/java/gov/cms/ab2d/common/util/FileUtil.java
```diff
@@ -0,0 +1,35 @@
+package gov.cms.ab2d.common.util;
```
dkrylovsb: We should keep the code used for testing under `test`, not `main`.

common/src/main/resources/application.common.properties
```diff
@@ -1,2 +1 @@
-# Replace with actual value once we figure out what this will look like
-efs.mount=src/test/resources/
\ No newline at end of file
+efs.mount=${java.io.tmpdir}jobdownloads/
```
dkrylovsb: I think we need a slash: `${java.io.tmpdir}/jobdownloads/`

dkrylovsb:
You don't really have to delete `common/src/test/resources/test.ndjson`

common/src/main/java/gov/cms/ab2d/common/util/FileUtil.java
adaykin: I had tried putting it in the test project under common, but I don't think it's visible to the code that uses it in api if it's under there.

common/src/main/java/gov/cms/ab2d/common/util/FileUtil.java
```diff
@@ -0,0 +1,35 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.List;
+
+public class FileUtil {
+
+    @VisibleForTesting
+    public static void createTmpFileForDownload(String jobID, String fileName) throws IOException {
+        List<String> lines = List.of("{",
```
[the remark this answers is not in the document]
adaykin: I was using static files, but changed to using this just so it can be created in the directory that is based off a job id, something that can't be predicted since it will be a UUID. It seems to mimic production more closely by doing it that way.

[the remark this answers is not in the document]
dkrylovsb: You'd have that test file under `src/test/resource` and in your test you'd actually copy it into whatever path you need. We don't really to maintain the file's content within the code itself.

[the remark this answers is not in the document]
dkrylovsb: @adaykin We can probably just drop this class altogether. Instead, in each test class we could just copy the `src/test/resource/test.ndjson` to wherever we need it.

PR #29, opened 2019-10-29
Initial creation of HPMS module

adaykin:
Creating HPMS module for usage

Creating skeleton with dockerfile, dependencies, and app

### Checklist

- [x] Tests and linting pass

### Security
NA


hpms/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM openjdk:12
```
dkrylovsb: We can drop `Dockerfile`. This module will be a dependency of `api`.

hpms/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+        <springfox.version>2.9.2</springfox.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>gov.cms.ab2d</groupId>
+            <artifactId>common</artifactId>
+            <version>0.0.1-SNAPSHOT</version>
+        </dependency>
+        <dependency>
```
dkrylovsb: Since `hpms` depends on `common`, it will pull down all the Spring starters from `common` transitively. We don't need to declare this one.

hpms/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
```
dkrylovsb: `repackage` will be inherited from the parent `app`, and we don't need to override jacoco either, I think. We won't have SpringBootApp in this module.

hpms/src/main/java/gov/cms/ab2d/hpms/SpringBootApp.java
```diff
@@ -0,0 +1,17 @@
+package gov.cms.ab2d.hpms;
```
dkrylovsb: `hpms` won't be a Boot app; can drop this

hpms/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+        <artifactId>app</artifactId>
+        <groupId>gov.cms.ab2d</groupId>
+        <version>0.0.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>hpms</artifactId>
+
+    <properties>
+        <project.root>${basedir}/..</project.root>
+        <springfox.version>2.9.2</springfox.version>
```
dkrylovsb: `springfox.version` still showing up

PR #31, opened 2019-10-31
Integrate Liquibase into project

spathiyil:
Integrate Liquibase into the project

### Summary

Introduce liquibase into the project so that the database migration can be done outside hibernate.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


api/src/main/resources/application.properties
```diff
@@ -7,7 +7,21 @@ spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
...
+## -----------------------------------------------------------------------------------------------------  LOGGING LEVEL
+logging.level.root=INFO
+logging.level.gov.cms.ab2d=INFO
+
+logging.level.org.springframework=WARN
+logging.level.com.zaxxer.hikari=INFO
+logging.level.org.hibernate=WARN
+## -------------------------------------------------------------------------------------------------  LIQUIBASE LOGGING
+logging.level.liquibase.executor.jvm.JdbcExecutor=INFO
+logging.level.liquibase.integration.spring=DEBUG
```
dkrylovsb: `DEBUG` might be too verbose

api/src/test/resources/application.properties
```diff
@@ -9,4 +9,7 @@ spring.integration.jdbc.initialize-schema=always
 spring.jpa.properties.hibernate.temp.use_jdbc_metadata_defaults=false
 spring.jpa.hibernate.ddl-auto=create-drop
 
+#-- disable liquibase for junit test which uses H2 instead of postgres
```
dkrylovsb: This is probably OK for now, as long as Liquibase migrations merely follow the domain model changes (in which case both running Liquibase or using Hibernate Auto-DDL would lead to the same schema). But if we ever add a migration that adds something to the DB that is *not* expressed in the model, we might have to start running Liquibase in tests (or think of a more creative solution). Just a thought at this point

common/src/main/java/gov/cms/ab2d/common/model/Attestation.java
```diff
@@ -20,7 +23,8 @@
 public class Attestation {
 
     @Id
-    @GeneratedValue
+    @GeneratedValue(strategy = SEQUENCE, generator = "attestation_id_seq")
```
dkrylovsb: I would like to reduce the clutter introduced by having to specify annotation attributes (strategy, generator, etc). Can't we just declare `@GeneratedValue` and rely on the defaults?

common/src/main/java/gov/cms/ab2d/common/model/Attestation.java
```diff
@@ -20,7 +23,8 @@
...
+    @SequenceGenerator(name = "attestation_id_seq", sequenceName = "attestation_id_seq", allocationSize = 1)
```
dkrylovsb: Same here. Can we let JPA pick the default values?

spathiyil: I tried to use simply the `@GeneratedValue` hoping JPA would pick the defaults.

But after I got my liquibase changesets working properly, when I tried to run the spring boot app, it started giving the following error in a stack trace : 

```
missing sequence [hibernate_sequence]
```

Since the primary key column is auto-incrementing,  hibernate needs to know how to increment it from 1, 2, 3 .... 

My first attempt at overcoming this was to use `strategy = 'IDENTITY'` which seemed simpler as it was a single property to the (existing) `@GeneratedValue` annotation.

But after doing some more research on it, I found a few articles recommending `SEQUENCE`  over `IDENTITY` (especially for PostgreSQL). Here are a couple of articles
[1](https://thoughts-on-java.org/hibernate-postgresql-5-things-need-know/) 
[2](https://vladmihalcea.com/hibernate-identity-sequence-and-table-sequence-generator/)

When liquibase creates the primary key for the tables, it automatically creates the corresponding _sequence generator_ for each of the tables.  So to get the `SEQUENCE` strategy working, I needed to let hibernate know the respecgive name of the `sequence_generator` for each of the `entities`. I keyed in the name of the PostgreSQL generated sequence numbers and it started working fine.   

If there is a better way to do this, I am happy to change it.

common/src/main/java/gov/cms/ab2d/common/model/Attestation.java
spathiyil: With only `@GeneratedValue` I was getting and error when attempting to run the boot App. (details in the comment below)

api/src/main/resources/application.properties
spathiyil: oops... The above configuration worked best while testing. I forgot to change it back after testing. 
Will change it to `WARN` for all the liquibase shortly

common/src/main/java/gov/cms/ab2d/common/model/Attestation.java
dkrylovsb: @spathiyil Just stick `CREATE SEQUENCE hibernate_sequence` into your first Liquibase migration and you'll be good to go with the defaults.

api/pom.xml
```diff
@@ -43,6 +43,10 @@
             <artifactId>springfox-swagger-ui</artifactId>
             <version>${springfox.version}</version>
         </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
spathiyil: I tried to move the liquibase dependency inside `common` module. But kept getting an error. I finally got it working when I moved the maven dependency outside `common`.  I cannot remember the details, but I'll try again to move the dependency into the `common` sub-module

api/src/main/resources/application.properties
dkrylovsb: @spathiyil One more thought: I am used to the fact that _usually_ logging configuration goes into its own property file: `logging.properties`, `log4j.xml`, etc. How hard would it be to move the loggers config into a separate file so that it does not clutter `application.properties`?

karlmdavis:
I know I'm just popping up here randomly, but happened to see this float past in my inbox.

FWIW, I pretty strongly recommend against Liquibase these days -- too many painful random bugs. Flyway is an excellent alternative. Best of all, it uses raw SQL for migrations, which means you get to do **all** the things -- not just the things that the library exposes for you.

api/src/main/resources/application.properties
spathiyil: This seems to be the default way for logging in `[spring-boot](https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html#howto-logging)`.

```
Spring Boot includes a number of extensions to Logback that can help with advanced configuration. .... Because the standard logback.xml configuration file is loaded too early, you cannot use extensions in it. ... The extensions cannot be used with Logback’s configuration scanning. If you attempt to do so, making changes to the configuration file results in an error similar to one of the following being logged:
```

Doing a quick read, it seems possible to get spring-boot's customization etc while creating a separate `logback-spring.xml` file. But am unsure how much effort it would be as I had not had to do it in a spring-boot project before.

Would it be okay, if we do it in a separate story outside of this PR?
For now, I could just delete the liquibase package names from here to reduce clutter, if it helps?

dkrylovsb:
@karlmdavis Good to know; thanks for the feedback! I have to confess I have picked Liquibase "by inertia" without considering alternatives, since I've been using it for years on other projects. But we'll switch to Flyway if Liquibase indeed turns out to be an issue.

api/src/main/resources/application.properties
dkrylovsb: @spathiyil Agreed!

spathiyil: done

common/src/main/java/gov/cms/ab2d/common/model/Attestation.java
spathiyil: done. created `hibernate_sequence` via liquibase, so that hibernate can use this as the default generator for generating the primary key for all the tables.

common/src/main/java/gov/cms/ab2d/common/model/Attestation.java
spathiyil: Changed to use the default `hibernate_sequence` are advised.

api/pom.xml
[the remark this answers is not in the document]
spathiyil: done - the liquibase dependencies have been moved into the `common` sub-module

api/src/test/resources/application.properties
spathiyil: I concur

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
```diff
@@ -0,0 +1,197 @@
...
+
+
+--changeset spathiyil:AB2D-291-CreateSequence-hibernate_sequence failOnError:true dbms:postgresql
+CREATE SEQUENCE hibernate_sequence START WITH 1 INCREMENT BY 1;
+
+
+--changeset spathiyil:AB2D-291-CreateTable-beneficiary failOnError:true dbms:postgresql
+CREATE TABLE beneficiary
+(
+    id                  BIGSERIAL,
```
dkrylovsb: @spathiyil I think this needs to become `bigint` since we are using a sequence, and not relying on auto-increment.

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
```diff
@@ -0,0 +1,197 @@
...
+ALTER TABLE beneficiary ADD CONSTRAINT "uc_beneficiary_patient_id" UNIQUE (patient_id);
+
+--rollback DROP TABLE beneficiary;
+--  -------------------------------------------------------------------------------------------------------------------
+
+
+--changeset spathiyil:AB2D-291-CreateTable-contract failOnError:true dbms:postgresql
+CREATE TABLE contract
+(
+    id                  BIGSERIAL,
```
dkrylovsb: Same as above

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
```diff
@@ -0,0 +1,197 @@
...
+ALTER TABLE coverage ADD CONSTRAINT "uc_coverage_beneficiary_id_contract_id" UNIQUE (beneficiary_id, contract_id);
+
+--rollback  DROP TABLE coverage;
+--  -------------------------------------------------------------------------------------------------------------------
+
+
+--changeset spathiyil:AB2D-291-CreateTable-sponsor failOnError:true dbms:postgresql
+CREATE TABLE sponsor
+(
+    id                  BIGSERIAL,
```
dkrylovsb: Change to bigint throughout

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
```diff
@@ -0,0 +1,197 @@
...
+    hpms_id             INTEGER             NOT NULL,
+    legal_name          VARCHAR(255)        NOT NULL,
```
dkrylovsb: Should be nullable. HPMS report does not show legal names for parent orgs.

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
```diff
@@ -0,0 +1,197 @@
...
+--  -------------------------------------------------------------------------------------------------------------------
+
+
+--changeset spathiyil:AB2D-291-CreateTable-attestation failOnError:true dbms:postgresql
+CREATE TABLE attestation
+(
+    id                  BIGSERIAL,
+    sponsor_id          BIGINT              NOT NULL,
+    contract_id         BIGINT              NOT NULL,
+    attested_on         TIMESTAMPTZ         NOT NULL
```
dkrylovsb: Nullable. Null date would mean a lack of attestation or a revoked attestation.

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
```diff
@@ -0,0 +1,197 @@
...
+
+
+--changeset spathiyil:AB2D-291-CreateTable-user_account failOnError:true dbms:postgresql
+CREATE TABLE user_account
+(
+    id                  BIGSERIAL,
+    user_name           VARCHAR(64)         NOT NULL,
+    first_name          VARCHAR(64)         NOT NULL,
+    last_name           VARCHAR(64)         NOT NULL,
+    email               VARCHAR(255)        NOT NULL,
```
dkrylovsb: Let's make these 3 nullable. Unclear at this point how much of user info we will get from HPMS

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
```diff
@@ -0,0 +1,197 @@
...
+ALTER TABLE user_role ADD CONSTRAINT "fk_user_role_to_role" FOREIGN KEY (role_id) REFERENCES role (id);
+ALTER TABLE user_role ADD CONSTRAINT "uc_user_role_user_account_id_role_id" UNIQUE (user_account_id, role_id);
+
+
+--rollback DROP TABLE user_role;
+--  -------------------------------------------------------------------------------------------------------------------
+
+
+--changeset spathiyil:AB2D-291-CreateTable-job_status failOnError:true dbms:postgresql
+CREATE TABLE job_status
```
dkrylovsb: Let's drop this table and just use a `enum` in code.

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
```diff
@@ -0,0 +1,197 @@
...
+CREATE TABLE job
+(
+    id                  BIGSERIAL,
+    job_id              VARCHAR(255)        NOT NULL,
+    user_account_id     BIGINT              NOT NULL,
+    created_at          TIMESTAMPTZ         NOT NULL,
+    expires_at          TIMESTAMPTZ         NOT NULL,
+    resource_types      VARCHAR(255)        NOT NULL,
+--  job_status_id       BIGINT              NOT NULL,
+    status              INT                 NOT NULL,
```
dkrylovsb: Let's change to `varchar` and use `@Enumerated(EnumType.STRING)` on the corresponding entity's field. Using enum names instead of integer ordinals would be more readable (at a small cost of increased storage)

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
spathiyil: done

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
spathiyil: done

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
spathiyil: done

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
spathiyil: done - legalName is now nullable

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
spathiyil: done. thanks for the clarification - it makes sense now.

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
spathiyil: done. firstName, lastName, email is now nullable

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
spathiyil: done. removed this change-set

common/src/main/resources/db/changelog/v001/AB2D-291-create-initial-tables.sql
spathiyil: done

PR #33, opened 2019-11-06
Attestation report processing

adaykin:
Attestation report processing for HPMS file.

### Summary

Attestation report processing, adding support for API call from admin controller, and processing report so that attestations can be updated in the database with a new datetime.

### Checklist

- [x] Tests and linting pass

### Security

NA


common/src/main/java/gov/cms/ab2d/common/model/Attestation.java
```diff
@@ -32,5 +32,8 @@
     @Column(columnDefinition = "TIMESTAMP WITH TIME ZONE")
     private OffsetDateTime attestedOn;
 
-
+    @Override
+    public int compareTo(Attestation other) {
+        return this.getAttestedOn().compareTo(other.getAttestedOn());
+    }
```
dkrylovsb: In general an implementation of `compareTo` should be consistent with `equals`, which is not quite the case here. This `compareTo` will equate two completely different attestation records as long as they happen to have the same attestation date. (more on this [here](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html)).

I'd recommend dumping this method and using an explicit `Comparator` wherever you need it.

common/src/main/java/gov/cms/ab2d/common/repository/AttestationRepository.java
```diff
@@ -1,10 +1,12 @@
...
 
 import gov.cms.ab2d.common.model.Attestation;
+import gov.cms.ab2d.common.model.Contract;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
 @Repository
 public interface AttestationRepository extends JpaRepository<Attestation, Long> {
 
+    Attestation findOneByContractOrderByAttestedOnDesc(Contract contract);
```
dkrylovsb: Since we don't need to maintain attestation history as HPMS does, we will always have only 1 attestation record per contract. So I guess we just need `findOneByContract`

common/src/main/java/gov/cms/ab2d/common/service/AttestationService.java
```diff
@@ -0,0 +1,11 @@
+package gov.cms.ab2d.common.service;
+
+import gov.cms.ab2d.common.model.Attestation;
+import gov.cms.ab2d.common.model.Contract;
+
+public interface AttestationService {
+
+    Attestation saveAttestation(Attestation attestation);
+
+    Attestation getMostRecentAttestationFromContract(Contract contract);
```
dkrylovsb: No need: not keeping the attestation history.

common/src/main/java/gov/cms/ab2d/common/repository/AttestationRepository.java
adaykin: The contract entity has multiple attestations per contract, should that be a 1:1 instead of 1:many?

common/src/main/java/gov/cms/ab2d/common/service/AttestationService.java
```diff
@@ -0,0 +1,11 @@
+package gov.cms.ab2d.common.service;
+
+import gov.cms.ab2d.common.model.Attestation;
+import gov.cms.ab2d.common.model.Contract;
+
+public interface AttestationService {
+
+    Attestation saveAttestation(Attestation attestation);
```
dkrylovsb: I think we probably want a single method in this service class that deals with the attestation logic. It would accept a `Sponsor`, `Contract`, and a nullable `Date` as parameters and either update an existing attestation record or create a new one.

common/src/main/java/gov/cms/ab2d/common/repository/AttestationRepository.java
dkrylovsb: @adaykin Yes.

common/src/main/java/gov/cms/ab2d/common/model/Attestation.java
adaykin: Probably won't need comparisons anymore, since there will only be one contract for attestation, I can just drop this.

hpms/src/main/java/gov/cms/ab2d/hpms/processing/AttestationReportProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+
+            while (iterator.hasNext()) {
+
+                Row currentRow = iterator.next();
+                if (currentRow == null) {
+                    continue;
+                }
+
+                String contractNumber = currentRow.getCell(0).getStringCellValue();
+                // This report is ordered by contract number and date DESC, so just move on to the next row if we've already seen a contract
```
dkrylovsb: This is a somewhat fragile assumption: if HPMS changes row ordering for whatever reason on their end, we will be storing incorrect attestation info. We have to actually look at the `Date/Time of Action` value to determine the current attestation record.

hpms/src/main/java/gov/cms/ab2d/hpms/processing/AttestationReportProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+                        OffsetDateTime offsetDateTime = OffsetDateTime.parse(attestetedDateTimeCell + " " + offset,
+                                DateTimeFormatter.ofPattern(ATTESTATION_OFFSET_DATE_TIME_PATTERN));
+                        attestation.setAttestedOn(offsetDateTime);
+                    } else {
+                        attestation.setAttestedOn(null);
+                    }
+
+                    attestationService.saveAttestation(attestation);
+                } else {
+                    log.warn("Contract ID {} was not found in the database during contract report processing", contractNumber);
```
dkrylovsb: Missing contract is an error condition. Let's throw an unchecked exception here, which will cause the transaction to rollback and the uploading admin to become aware of the problem.

dkrylovsb:
@adaykin Let's supplement this PR with a Liquibase changelog that adds an `unqiue` constraint on `attestation.contract_id`.

common/src/main/java/gov/cms/ab2d/common/service/AttestationService.java
adaykin: I think in the case of the attestation report, there should already be a sponsor associated with a contract since we only process contracts currently in the system, and the contract would have an attestation, so it would just need to be updating the attestation date.

dkrylovsb: OK

PR #34, opened 2019-11-06
Renaming columns so that they are less ambiguous

adaykin:
Renaming columns so that they are less ambiguous

### Summary

Renaming columns so that they are less ambiguous

### Checklist

- [x] Tests and linting pass

### Security

NA


common/src/main/resources/db/changelog/v001/AB2D-356-rename-columns.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE job RENAME job_id TO job_uuid;
```
spathiyil: There is a unique constraint on the `job` table on `job_id` named `uc_job_job_id` both to enforce uniqueness and also for the index it creates.  I think you may need to recreate it to use the `job_uuid` column.

dkrylovsb: @adaykin Let's re-create (or rename, if possible) the constraint and this should be good to go into `master`

PR #35, opened 2019-11-06
Create message handler in worker

spathiyil:
Create message handler in worker

### Summary

Create an initial version of the WorkerService and associated classes to bootstrap the sample worker.
When a record is inserted into the job table with the status `SUBMITTED`, a JdbcPollingChannelAdapter which polls the job table every second streams the oldest job record into a Subscribable Executor channel.
This triggers the JobHandler which delegates to a WorkerService to process the job.

Currently, the process method simply sleeps for 5 seconds to simulate a long running process before updating the job table with a `SUCCESSFUL` status

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


worker/src/main/java/gov/cms/ab2d/worker/SpringBootApp.java
```diff
@@ -2,9 +2,15 @@
...
+import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
 
 
-@SpringBootApplication
+@SpringBootApplication(scanBasePackages = {"gov.cms.ab2d.common", "gov.cms.ab2d.worker"})
+@EntityScan(basePackages = {"gov.cms.ab2d.common.model"})
+@EnableJpaRepositories("gov.cms.ab2d.common.repository")
+@PropertySource("classpath:application.common.properties")
```
dkrylovsb: Wondering how this is working since `application.common.properties` does not have DB connection properties? The Worker will run in a separate container and will have its own `application.properties` (just like the API).

dkrylovsb: OK, nevermind, I guess `application.properties` is picked up by default as well

worker/src/main/java/gov/cms/ab2d/worker/config/WorkerConfig.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * sample values used for thread count and queue capacity. Review it later.
+     */
+    @Bean
+    public ThreadPoolTaskExecutor threadPoolTaskExecutor() {
+        final ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
+        taskExecutor.setCorePoolSize(5);
+        taskExecutor.setMaxPoolSize(10);
+        taskExecutor.setQueueCapacity(50);
```
dkrylovsb: Let's make `0` the default queue capacity. This will cause the executor to select `SynchronousQueue` as the queue implementation. That way the worker will simply reject new jobs when the thread pool is exhausted rather than queuing them up and waiting. More info at https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html?is-external=true

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerService.java
```diff
@@ -0,0 +1,65 @@
...
+        this.jobRepository = jobRepository;
+    }
+
+    public void process(String jobId) throws IOException {
+
+        final Job job = jobRepository.findByJobUuid(jobId);
+        if (job == null) {
+            log.warn("Job not found for job_id : {} ", jobId);
+            return;
+        }
```
dkrylovsb: A job being null is an error condition worth throwing an exception. You could replace these with a one-liner something like
`Assert.notNull(job, String.format("Job %s not found unexpectendly", jobId));`

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerService.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private final JobRepository jobRepository;
+
+    public WorkerService(JobRepository jobRepository) {
+        this.jobRepository = jobRepository;
+    }
+
+    public void process(String jobId) throws IOException {
+
+        final Job job = jobRepository.findByJobUuid(jobId);
```
dkrylovsb: Even though the `process` method itself should not be transactional as we discussed (it's long-running), _ some parts_ of the workflow still should. `process` begins by doing 3 things: find the job, verify it is still in `SUBMITTED` status, move to `IN_PROGRESS`. These 3 things should be done in a transaction; so I guess they need to be extracted into a separate _public_ method with tx annotation applied.

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerService.java
```diff
@@ -0,0 +1,65 @@
...
+            log.warn("Job not found for job_id : {} ", jobId);
+            return;
+        }
+
+        putJobInProgress(job);
+        doLongRunningWork();
+        completeJob(job);
+    }
+
+    private void putJobInProgress(Job job) {
```
dkrylovsb: This could be the method that does the "3 things" above.

worker/src/main/java/gov/cms/ab2d/worker/config/JobMessageSource.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private static final String QUERY_GET_NEXT_UNPROCESSED_JOB = "                  " +
+                                    "    SELECT id, job_uuid, status                " +
+                                    "      FROM job                                 " +
+                                    "     WHERE status = 'SUBMITTED'                " +
+                                    "       AND (SELECT count(lock_key)             " +
+                                    "              FROM int_lock                    " +
+                                    "             WHERE lock_key = job_uuid) = 0    " +
+                                    "  ORDER BY created_at;                         ";
+
```
spathiyil: I was wondering, if I should also add a `LIMIT 1` at the end of this query.
I did not add it because line 25 below has a `setMaxRows(1);` 

Thoughts?

dkrylovsb: @spathiyil I think we are good as is. I don't feel strongly either way though

worker/src/main/java/gov/cms/ab2d/worker/config/WorkerConfig.java
spathiyil: changed QueueCapacity to 0

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerService.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerService.java
spathiyil: This was a good discussion - helped me understand the scenario better.
As discussed, I have extracted a `JobProcessingService`  where I created a public method `putJobInProgress` annotated as follows:
```@Transactional(propagation = Propagation.REQUIRES_NEW, isolation = Isolation.SERIALIZABLE)```

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerService.java
spathiyil: done.

PR #37, opened 2019-11-08
Create a mock implementation of BFD API

spathiyil:
Create a mock implementation of BFD API

### Summary

Create an implementation for a mock BFD API, so that the AB2D system can retrieve benes for a Contract ID

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/BeneficiaryAdapterStub.java
```diff
@@ -0,0 +1,112 @@
...
+    private Integer extractContractSno(String contractNumber) {
+        final String contractNumberSuffix = contractNumber.substring(contractNumber.length() - 3);
+
+        final Integer sno;
+        try {
+            sno = Integer.valueOf(contractNumberSuffix);
+        } catch (NumberFormatException e) {
+            final String errMsg1 = String.format("Invalid ContractNumber : %s.  ", contractNumber);
+            final String errMsg2 = "The rightmost 3 characters of the contract number must be numeric. ";
+            throw new IllegalArgumentException(errMsg1 + errMsg2);
```
dkrylovsb: I'd say, for any contract number that does not fall into S0001--S0300 pattern, just return an empty list of benes (since this is what BFD would do). No need to throw IAE.

dkrylovsb:
@spathiyil We probably want to add Jacoco coverage exclusions for this code, since it is temporary anyway.

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/BeneficiaryAdapterStub.java
spathiyil: done

PR #40, opened 2019-11-14
AB2D-373: Liquibase, not Hibernate's auto-DDL, should run in tests as well

dkrylovsb:
Liquibase, not Hibernate's auto-DDL, should run in tests as well.

### Summary

All of the integration tests now run the actual Liquibase migrations to prep the database instead of relying on the DDL auto-generation. This ensures the tests execute against exactly the same schema that is used in the environments. The underlying database in tests is still H2 though (albeit in PostgreSQL-compatibility mode): we'll need to think about switching to Test Containers eventually.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/common/src/test/java/gov/cms/ab2d/common/service/JobServiceTest.java
+++ b/common/src/test/java/gov/cms/ab2d/common/service/JobServiceTest.java
@@ -1,10 +1,10 @@
 package gov.cms.ab2d.common.service;
 
 import gov.cms.ab2d.common.SpringBootApp;
-import gov.cms.ab2d.common.model.Job;
-import gov.cms.ab2d.common.model.JobOutput;
-import gov.cms.ab2d.common.model.JobStatus;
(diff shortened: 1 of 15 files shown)


common/src/test/java/gov/cms/ab2d/common/service/JobServiceTest.java
```diff
@@ -60,15 +89,16 @@ public void createJob() {
         assertThat(job.getId()).isNotNull();
         assertThat(job.getJobUuid()).isNotNull();
         assertEquals(job.getProgress(), Integer.valueOf(0));
-        assertEquals(job.getUser(), null); // null for now since no authentication
+        assertEquals(job.getUser(), userRepository.findByUserName("example@example.com")); // null for now since no authentication
```
spathiyil: We can remove the comment now that we have user for a job.

PR #41, opened 2019-11-19
ab2d implement job workflow process v01

spathiyil:
A draft implementation for the job workflow process.

### Summary

This PR contains work from the following JIRA tickets :
[AB2D-375](https://jira.cms.gov/browse/AB2D-375)
[AB2D-376](https://jira.cms.gov/browse/AB2D-376)
[AB2D-340](https://jira.cms.gov/browse/AB2D-340)
[AB2D-377](https://jira.cms.gov/browse/AB2D-377)
[AB2D-378](https://jira.cms.gov/browse/AB2D-378)

When a new job is created, this job processing workflow is triggerred.
Given the job id, it will identify the sponsor based on the user and then all the contracts attached to the sponsor.  It will create an output directory for the job containing the job_id in the name
For each of the contracts, it will call the `BeneficiaryAdapterStub` created in the last sprint that returns a set of patient records for the contract.
For each of the patients, it will call the mock BFD client to get a bundle of resources asynchronously.
Periodically, it will pause to get the response form the futures and for each bundle (list of resources) for a patient, writes it out to an ByteArrayOutputStream. 
And then, when all the resources for the patient has been written into the outputstream, it will write it out to the file and flush it in one shot.

For each contract, it also writes out a job_output record.

This is an early implementation and there are a few more iterations done before the workflow process is complete.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +60,136 @@ public Job putJobInProgress(String jobId) {
...
+                jobOutput = processContract(outputDir, contract);
+                jobOutput.setError(false);
+            } catch (Exception e) {
+                jobOutput.setError(true);
+                log.error("error processing contract : {} ", contract.getContractNumber(), e);
+            }
+
+            job.addJobOutput(jobOutput);
+            jobOutputRepository.save(jobOutput);
+        }
```
spathiyil: My understanding is that each contract should have a row in the `job_output` table. Is that correct?

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +60,136 @@ public Job putJobInProgress(String jobId) {
...
+            if (counter % GROUP_SIZE == 0) {
+                processResources(futureResourcesHandles, ndJsonFile);
+            }
+        }
+
+        processResources(futureResourcesHandles, ndJsonFile);
+
+        JobOutput jobOutput = new JobOutput();
+        jobOutput.setFilePath(getFilePath(ndJsonFile));
+        jobOutput.setFhirResourceType("ExplanationOfBenefits");
```
spathiyil: I do not know how to get the value for resourceType.  I set a default value of `ExplanationOfBenefits`.  Does this seem okay

worker/src/main/java/gov/cms/ab2d/worker/service/FileService.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static java.nio.file.StandardOpenOption.APPEND;
+
```
spathiyil: All I/O related activities like creating a directory or file or writing to one has been extracted from the main business logic into this class, so that the single responsibility of this class is doing I-O.

worker/pom.xml
```diff
@@ -86,14 +86,16 @@
...
                                         </limit>
                                     </limits>
                                 </rule>
                             </rules>
                             <excludes>
+                                <exclude>**/gov/cms/ab2d/worker/SpringBootApp.java</exclude>
                                 <!-- exclude the bluebutton package from coverage as it only has a stub -->
-                                <exclude>**/gov/cms/ab2d/worker/adapter/bluebutton/*</exclude>
+                                <exclude>**/gov/cms/ab2d/worker/adapter/bluebutton/**</exclude>
+                                <exclude>**/gov/cms/ab2d/worker/service/FileService.java</exclude>
```
dkrylovsb: We can test `FileService` without mocking.

worker/src/main/java/gov/cms/ab2d/worker/config/WorkerConfig.java
```diff
@@ -39,6 +43,19 @@
     @Autowired
     private JobHandler handler;
 
+    @Bean("bfd-client")
+    public Executor bfdThreadPoolTaskExecutor() {
+        final ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
+        taskExecutor.setCorePoolSize(5);
```
dkrylovsb: Let's make these 3 pool parameters configurable via `application.properties` (can be done in a sub-task)

worker/src/main/java/gov/cms/ab2d/worker/service/FileService.java
```diff
@@ -0,0 +1,49 @@
...
+
+    Path createDirectory(Path outputDir) {
+        Path outputDirectory = null;
+        try {
+            outputDirectory = Files.createDirectories(outputDir);
+        } catch (IOException e) {
+            final String errMsg = "Could not create output directory : ";
+            log.error("{} : {}", errMsg, outputDir.toAbsolutePath());
+            throw new RuntimeException(errMsg + outputDir.getFileName(), e);
```
dkrylovsb: `java.io.UncheckedIOException` could be a better fit.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +60,136 @@ public Job putJobInProgress(String jobId) {
...
-    public void completeJob(Job job) {
+    public Job processJob(final String jobId) {
+
+        final Job job = jobRepository.findByJobUuid(jobId);
+        log.info("Found job : {} - {} - {} ", job.getId(), job.getJobUuid(), job.getStatus());
+
+        final Sponsor sponsor = job.getUser().getSponsor();
+
+        final List<Contract> attestedContracts = sponsor.getAttestedContracts();
+        log.info("number of attested contracts : {}", attestedContracts.size());
```
dkrylovsb: Let's include job's UUID in this message as well. Will help traceability in Splunk.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +60,136 @@ public Job putJobInProgress(String jobId) {
...
+
+
+        final Path outputDirPath = Paths.get(efsMount, job.getJobUuid());
+        final Path outputDir = fileService.createDirectory(outputDirPath);
+        for (Contract contract : attestedContracts) {
+            log.info("contract : {} ", contract.getContractNumber());
```
dkrylovsb: Include job's UUID

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
dkrylovsb: @spathiyil For the MVP, a job might have up to **2** outputs. Consider this: you are processing a contract with 100 patients, out of which 1 failed (a non-fatal condition). You would still write a _success_ `JobOutput` with `ndjson` with EOB resources for the 99 patients, but you'd also write an _error_ `JobOutput` with `ndjson` with a single `OperationOutcome` entry that explains what happened to the failed patient.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +60,136 @@ public Job putJobInProgress(String jobId) {
...
 
         job.setStatus(IN_PROGRESS);
 
-        log.info("Job [{}] is now IN_PROGRESS", job.getId());
         return jobRepository.save(job);
     }
 
     @Override
-    public void completeJob(Job job) {
+    public Job processJob(final String jobId) {
```
dkrylovsb: Let's add `TransactionPropagation.NEVER` just in case

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
dkrylovsb: Yes, that's the only resource type we support at the moment.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +60,136 @@ public Job putJobInProgress(String jobId) {
...
+            ++counter;
+            if (counter % GROUP_SIZE == 0) {
+                processResources(futureResourcesHandles, ndJsonFile);
+            }
+        }
+
+        processResources(futureResourcesHandles, ndJsonFile);
+
+        JobOutput jobOutput = new JobOutput();
+        jobOutput.setFilePath(getFilePath(ndJsonFile));
```
dkrylovsb: The file path here is supposed to be relative to the EFS mount, not the absolute one. More portability that way.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +60,136 @@ public Job putJobInProgress(String jobId) {
...
+
+    private void processResources(List<Future<List<Resource>>> futureHandles, Path outputFile) {
+
+        final Iterator<Future<List<Resource>>> iterator = futureHandles.iterator();
+        while (iterator.hasNext()) {
+            final Future<List<Resource>> futureResources = iterator.next();
+
+            List<Resource> resources = null;
+            try {
+                resources = futureResources.get();
```
dkrylovsb: Iterating over a list of `Futures` and calling `get` on each might not scale very well and hamper parallel processing, too. The problem here is that `get` is a _blocking_ method: it will be stuck until that particular Future is done. If you encounter one particularly slow Future, you will be stuck with it while others might have finished already.

At minimum, check for `isDone()` before calling `get`. But see my other comments: you might be better off refactoring the entire thing a bit.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: Currently, the method does not have an `@Transactional` annotation.
Would I need to add an `@Transactional` on top of the `processJob()`  to set `TransactionPropagation.NEVER` property?

worker/src/test/java/gov/cms/ab2d/worker/service/JobProcessingServiceImplTest.java
```diff
@@ -0,0 +1,199 @@
...
+    private Random random = new Random();
+
+    @Autowired
+    private FhirContext fhirContext;
+
+    @Mock FileService fileService;
+    @Mock JobRepository jobRepository;
+    @Mock JobOutputRepository jobOutputRepository;
+    @Mock BeneficiaryAdapter beneficiaryAdapter;
+    @Mock BfdClientAdapter bfdClientAdapter;
```
dkrylovsb: A complete _integration_ test will be extremely valuable here. We can avoid mocking all these relatively easily:

- `FileService` does not need to be mocked. It can write to the file system during test.
- You can use the "real" BFD client as well. Take a look at the tests in the `bfd` module. It stands up a mock HTTP server and serves test data (which you can define in XML) in response to FHIR Client REST calls. We should do the same here.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: okay. After change the filepath will look something like this:

```
JOB_11/CONTRACT_S0011.ndjson
```

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/BfdClientAdapterImpl.java
```diff
@@ -0,0 +1,51 @@
...
+@Component
+@RequiredArgsConstructor
+public class BfdClientAdapterImpl implements BfdClientAdapter {
+
+    @Autowired
+    private BFDClient bfdClient;
+
+    @Async("bfd-client")
+    @Override
+    public Future<List<Resource>> getEobBundleResources(String patientId) {
```
dkrylovsb: So this is the _async_ method that deals with a single patient's EOBs. I _think_, rather than trying to sort out the Futures in the "main" job Thread, you could just do the the entire per-patient flow right here:

- Get the EOBs from BFD
- Serialize them into JSON
- Append to the `ndjson` file. Need to pay attention to the fact that we have _one_ file and _many_ threads writing to it.

So the "main" job thread would just have to wait until all Futures are done and move the job to `COMPLETED`. You should end up with a cleaner code using this approach.

spathiyil: ```Need to pay attention to the fact that we have one file and many threads writing to it.```
The way I solved the contention during writing to the file is by doing the writing to the ndjson file into the main thread which being single threaded would not have the contention issue.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
dkrylovsb: yup

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
dkrylovsb: Correct.

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/BfdClientAdapterImpl.java
dkrylovsb: Right. `java.util.concurrent.locks.ReentrantLock` should help do the same with the new approach.

worker/src/test/java/gov/cms/ab2d/worker/service/JobProcessingServiceImplTest.java
spathiyil: Changed the unit test into an integration test that does not mock any of the dependencies.  The integration test will take longer to run than the unit test.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: done

worker/pom.xml
spathiyil: Removed FileService from the excludes list here.

worker/src/main/java/gov/cms/ab2d/worker/service/FileService.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/BfdClientAdapterImpl.java
spathiyil: I've attempted to refactor as described above where the output and error files would be written to from inside the async method.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: refined this section to fit better with the refactored logic in the comment below

worker/src/main/java/gov/cms/ab2d/worker/config/WorkerConfig.java
spathiyil: ok. I will make this change in the next sub-task for this story.

spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/BfdClientAdapter.java
```diff
@@ -0,0 +1,11 @@
+package gov.cms.ab2d.worker.adapter.bluebutton;
+
+import java.nio.file.Path;
+import java.util.concurrent.Future;
+import java.util.concurrent.locks.ReentrantLock;
+
+public interface BfdClientAdapter {
```
dkrylovsb: We should probably rename this interface and the corresponding implementation class to more accurately reflect its purpose, since it does more than just "adapting" the BFD client. `PatientClaimsProcessor` or something along these lines would do it.

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/PatientClaimsProcessorImpl.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * uses re-entrant lock to lock the shared file resource
+     * @param outputFile
+     * @param byteArrayOutputStream
+     * @throws IOException
+     */
+    private void appendToFile(Path outputFile, ByteArrayOutputStream byteArrayOutputStream, ReentrantLock lock) throws IOException {
+        lock.lock();
```
dkrylovsb: We should use `tryLock(long timeout, TimeUnit unit)` instead. Imagine a different thread has acquired the lock, but got stuck writing JSON to the output file due to I/O issues (EFS is a network file system). All of the other threads that reach this line will "freeze" trying to acquire the lock. 

Let's make the timeout configurable via `application.properties` and set it to the default of 30 seconds. 

If `tryLock` returns `false`, let's consider this a non-recoverable error, throw an exception and thus terminate any further processing of the `Future`.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +58,159 @@ public Job putJobInProgress(String jobId) {
...
+    }
+
+
+    private List<Contract> getContractsOfChildrenSponsor(Sponsor sponsor) {
+        final List<Sponsor> childrenSponsors = sponsorRepository.findByParent(sponsor);
+        return childrenSponsors.stream()
+                .map(s -> s.getAttestedContracts())
+                .flatMap(Collection::stream)
+                .collect(Collectors.toList());
+    }
```
dkrylovsb: My goal was to have this logic encapsulated in the domain class itself. Asked Monica to create a tech debt ticket to address this.

spathiyil: hmm. I needed to make another query using a repository.
```sponsorRepository.findByParent(sponsor);```
which is why I ended up putting it in the service layer.

File: worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java [hunk 17]
```diff
@@ -30,28 +58,159 @@ public Job putJobInProgress(String jobId) {
...
+        final Path outputDir = fileService.createDirectory(outputDirPath);
+
+        for (Contract contract : attestedContracts) {
+            log.info("Job [{}] - contract [{}] ", job.getJobUuid(), contract.getContractNumber());
+
+            var jobOutputs = processContract(outputDir, contract);
+
+            jobOutputs.forEach(jobOutput -> job.addJobOutput(jobOutput));
+            jobOutputRepository.saveAll(jobOutputs);
+        }
```
dkrylovsb: So any uncaught exception here will terminate the job processing (which is what we need), but it won't set the job's status to `FAILED`. Let's add a try/catch block that does so.

For `FAILED` jobs we don't need to create JobOutput records. Let's write the exception message to `Job.statusMessage` though.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
dkrylovsb: Right. We might need to introduce `Sponsor.children` JPA-managed collection.

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/BfdClientAdapter.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/PatientClaimsProcessorImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: sounds good

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +59,167 @@ public Job putJobInProgress(String jobId) {
...
+            if (future.isDone()) {
+                try {
+                    var responseCount = future.get();
+                    if (responseCount > 0) {
+                        errorCount += responseCount;
+                    }
+                } catch (InterruptedException  e) {
+                    ++errorCount;
+                    log.error("interrupted exception while processing patient ", e);
+                } catch (ExecutionException e) {
```
dkrylovsb: This exception means that the Future ended with an uncaught exception, which was likely a non-recoverable condition. Let's re-throw it as RuntimeException and let the entire Job to fail.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -30,28 +59,167 @@ public Job putJobInProgress(String jobId) {
...
+                    log.error("exception while processing patient ", e.getCause());
+                }
+                iterator.remove();
+            }
+        }
+
+        return errorCount;
+    }
+
+    private JobOutput createPartialJobOutput(Path outputFile) {
```
dkrylovsb: Based on how this method is being used, the value of the `error` flag is know by the time this method is called. So you might as well just call it `createJobOutput` and pass `isError` as a parameter.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: This comment is only for the `ExecutionException` ? 
Or should I do the same for `InterruptedException` as well?

worker/src/test/java/gov/cms/ab2d/worker/service/JobProcessingServiceTest.java
```diff
@@ -0,0 +1,171 @@
...
+        var processedJob = cut.processJob("S001");
+
+        assertThat(processedJob.getStatus(), is(JobStatus.SUCCESSFUL));
+        assertThat(processedJob.getStatusMessage(), is("100%"));
+        assertThat(processedJob.getExpiresAt(), notNullValue());
+    }
+
+    private Sponsor createSponsor() {
+        Sponsor sponsor = new Sponsor();
+        sponsor.setId(Long.valueOf(random.nextInt()));
```
dkrylovsb: Why do we need to generate a primary key? Hibernate's sequence will take care of that.

worker/src/test/java/gov/cms/ab2d/worker/service/JobProcessingServiceTest.java
```diff
@@ -0,0 +1,171 @@
...
+        sponsor.setOrgName("Hogwarts School of Wizardry");
+        sponsor.setLegalName("Hogwarts School of Wizardry LLC");
+        sponsor.setHpmsId(random.nextInt());
+        return sponsorRepository.save(sponsor);
+    }
+
+    private User createUser(Sponsor sponsor) {
+        User user = new User();
+        user.setId(Long.valueOf(random.nextInt()));
```
dkrylovsb: Same as above.

worker/src/test/java/gov/cms/ab2d/worker/service/WorkerServiceTest.java
```diff
@@ -88,10 +88,17 @@ private User createUser() {
     }
 
     private Sponsor createSponsor() {
+        final Sponsor parentSponsor = new Sponsor();
+        parentSponsor.setId((long) getIntRandom());
```
dkrylovsb: Same

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
dkrylovsb: Good question. Let's do both.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: done

worker/src/test/java/gov/cms/ab2d/worker/service/JobProcessingServiceTest.java
spathiyil: done

worker/src/test/java/gov/cms/ab2d/worker/service/JobProcessingServiceTest.java
spathiyil: done

worker/src/test/java/gov/cms/ab2d/worker/service/WorkerServiceTest.java
spathiyil: done

PR #42, opened 2019-11-19
Feature/ab2d 76

adaykin:
Add authentication and exception handling

Add authentication with Spring Security so that we can receive a JWT token and then validate it with Okta. Once we have validated the token, we will then check the user to see if it exists in our database and the user is enabled. If everything checks out, set the Spring Security context.

### Checklist

- [x] Tests and linting pass

### Security

This effects the authentication to the system, only users that are in our database should be able to login and use our system. They must be able to be validated by okta with a JWT token.


api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
```diff
@@ -37,6 +42,16 @@
...
     }
 
+    @ExceptionHandler({MissingTokenException.class, InvalidAuthHeaderException.class})
+    public ResponseEntity<JsonNode> handleUnauthorizedAccessExceptions(final RuntimeException e) throws IOException {
+        return generateFHIRError(e, HttpStatus.UNAUTHORIZED);
+    }
+
+    @ExceptionHandler({BadJWTTokenException.class, UserNotFoundException.class, UserNotEnabledException.class})
+    public ResponseEntity<JsonNode> handleForbiddenAccessExceptions(final RuntimeException e) throws IOException {
+        return generateFHIRError(e, HttpStatus.FORBIDDEN);
```
dkrylovsb: Since AuthN/Z in our case is out of scope of the Bulk API spec, we don't need to return anything other than the HTTP code itself.

api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
```diff
@@ -37,6 +42,16 @@
         return generateFHIRError(e, HttpStatus.BAD_REQUEST);
     }
 
+    @ExceptionHandler({MissingTokenException.class, InvalidAuthHeaderException.class})
+    public ResponseEntity<JsonNode> handleUnauthorizedAccessExceptions(final RuntimeException e) throws IOException {
+        return generateFHIRError(e, HttpStatus.UNAUTHORIZED);
```
dkrylovsb: Same. Just kick it back with the HTTP error code only.

api/src/main/java/gov/cms/ab2d/api/security/JwtConfig.java
```diff
@@ -0,0 +1,25 @@
+package gov.cms.ab2d.api.security;
+
+import lombok.Getter;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+@Getter
+@Configuration
+public class JwtConfig {
```
dkrylovsb: Let's keep only the values we actually need. I am not sure we need `secret` or `expiration` (the latter is set in the token generated by Okta). `uri` might not be needed, either.

api/src/main/java/gov/cms/ab2d/api/security/JwtTokenAuthenticationFilter.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Autowired
+    private JwtConfig jwtConfig;
+
+    @Autowired
+    private UserService userService;
+
+    public JwtTokenAuthenticationFilter() {
+        jwtVerifier = JwtVerifiers.accessTokenVerifierBuilder()
+                .setIssuer("https://dev-418212.okta.com/oauth2/default")
```
dkrylovsb: Issuer URL must come from app properties file

api/src/main/java/gov/cms/ab2d/api/security/JwtTokenAuthenticationFilter.java
```diff
@@ -0,0 +1,98 @@
...
+                .setAudience("api://default")                // defaults to 'api://default'
+                .setConnectionTimeout(Duration.ofSeconds(1)) // defaults to 1s
+                .setReadTimeout(Duration.ofSeconds(1))       // defaults to 1s
```
dkrylovsb: Let's make these two configurable properties as well. 1 second is likely too aggresive; 5 seconds might be a more conservative choice.

api/src/test/java/gov/cms/ab2d/api/controller/AuthenticationTests.java
```diff
@@ -0,0 +1,143 @@
...
+    @Test
+    public void testNoAuthHeader() throws Exception {
+        this.mockMvc.perform(get(API_PREFIX + "/Patient/$export")
+                .contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().is(401))
+                .andExpect(jsonPath("$.resourceType", Is.is("OperationOutcome")))
+                .andExpect(jsonPath("$.issue[0].severity", Is.is("error")))
+                .andExpect(jsonPath("$.issue[0].code", Is.is("invalid")))
+                .andExpect(jsonPath("$.issue[0].details.text",
+                        Is.is("InvalidAuthHeaderException: Authorization header for token was not present")));
```
dkrylovsb: These won't be needed anymore (see my previous comments).

api/src/test/java/gov/cms/ab2d/api/controller/TestUtil.java
```diff
@@ -0,0 +1,111 @@
...
+
+        var values = new HashMap<String, String>() {{
+            put("grant_type", "password");
+            put("username", TEST_USER);
+            put("password", "Te$t2019");
+            put("scope", "openid");
+        }};
+
+        HttpClient client = HttpClient.newHttpClient();
+        HttpRequest request = HttpRequest.newBuilder()
```
dkrylovsb: The tests can't get tokens from Okta. Making tests dependent on an _external_ system makes them fragile. We have to make our own self-signed JWT here.

api/src/main/java/gov/cms/ab2d/api/security/JwtTokenAuthenticationFilter.java
```diff
@@ -0,0 +1,98 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.time.Duration;
+
+@Component
+public class JwtTokenAuthenticationFilter extends OncePerRequestFilter {
+
+    private final AccessTokenVerifier jwtVerifier;
```
dkrylovsb: We'll have to mock the verifier in the tests given my other comment about the fact we now have to "cook" our own JWT tokens.

api/src/main/java/gov/cms/ab2d/api/security/SecurityConfig.java
```diff
@@ -0,0 +1,87 @@
...
+        security
+            .csrf().disable()
+            .sessionManagement().sessionCreationPolicy(SessionCreationPolicy.STATELESS)
+            .and()
+            // Setup filter exception handling
+            .addFilterBefore(filterChainExceptionHandler, LogoutFilter.class)
+            // Add a filter to validate the tokens with every request. Since class is instantiated outside of DI, need to pass autowired, and value objects to it directly
+            .addFilterAfter(jwtTokenAuthenticationFilter, UsernamePasswordAuthenticationFilter.class)
+            .authorizeRequests()
+            .antMatchers(HttpMethod.POST, jwtConfig.getUri()).permitAll()
```
dkrylovsb: I don't think we need this line since we don't expose any `auth` endpoints. Can remove the corresponding JwtConfig property too.

api/src/test/java/gov/cms/ab2d/api/controller/RoleTests.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Map;
+
+import static gov.cms.ab2d.api.util.Constants.*;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = SpringBootApp.class, webEnvironment = SpringBootTest.WebEnvironment.MOCK)
+@AutoConfigureMockMvc
+public class RoleTests {
```
dkrylovsb: Need a test case to verify that an _existent_ user but with *no* roles cannot access FHIR or ADMIN.

api/src/test/java/gov/cms/ab2d/api/controller/AuthenticationTests.java
```diff
@@ -0,0 +1,108 @@
...
+
+import static gov.cms.ab2d.api.controller.TestUtil.TEST_USER;
+import static gov.cms.ab2d.api.util.Constants.*;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = SpringBootApp.class, webEnvironment = SpringBootTest.WebEnvironment.MOCK)
+@AutoConfigureMockMvc
+public class AuthenticationTests {
```
dkrylovsb: We are missing a test case that covers `JwtVerificationException`, i.e. when a token fails validation. So I guess `TestUtils` needs a `setupInvalidToken` method that will configure the verifier mock to throw `JwtVerificationException`.

api/src/test/resources/application.properties
```diff
@@ -6,3 +6,7 @@ spring.jpa.hibernate.ddl-auto=none
 spring.liquibase.enabled=true
 
 api.retry-after.delay=30
+
+api.okta-url=https://dev-418212.okta.com/oauth2/default
```
dkrylovsb: let's set this to `example.com` just to make sure we don't have tests that try to connect to Okta without us realizing that.

common/src/main/java/gov/cms/ab2d/common/model/User.java
```diff
@@ -62,4 +56,43 @@ public void removeRole(Role role) {
...
+    }
+
+    @Override
+    public String getUsername() {
+        return username;
+    }
+
+    @Override
+    public boolean isAccountNonExpired() {
+        return false;
```
dkrylovsb: We don't expire accounts, so should return `true`.

common/src/main/java/gov/cms/ab2d/common/model/User.java
```diff
@@ -62,4 +56,43 @@ public void removeRole(Role role) {
...
+    }
+
+    @Override
+    public boolean isAccountNonLocked() {
+        return false;
```
dkrylovsb: Should return `enabled`

common/src/main/java/gov/cms/ab2d/common/model/User.java
```diff
@@ -62,4 +56,43 @@ public void removeRole(Role role) {
...
+    }
+
+    @Override
+    public boolean isCredentialsNonExpired() {
+        return false;
```
dkrylovsb: we don't manage the credentials, so should return `true`

PR #44, opened 2019-11-22
jacoco config

adaykin:
Moving jacoco config to top level POM

Moving jacoco config to top level POM, overriding only files that are in 

### Checklist

- [x] Tests and linting pass

### Security
NA


adaykin:
I think the issue was I forgot to remove the SpringBootApp in the top level POM, so once I removed it and added some required lines in the child POM, the effective POM files were able to get the correct values.

PR #56, opened 2019-11-27
create new module -optout- with quartz dependency

spathiyil:
create new module named optout. Add the quartz dependency in it

### Summary

create new module named optout. Add the quartz dependency in it

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


optout/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+                <artifactId>spring-boot-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>repackage</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <executable>true</executable>
```
dkrylovsb: Opt-out wouldn't be an executable Spring Boot app.

optout/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
```
dkrylovsb: We can remove this entire section.

optout/src/main/java/gov/cms/ab2d/optout/SpringBootApp.java
```diff
@@ -0,0 +1,18 @@
+package gov.cms.ab2d.optout;
```
dkrylovsb: Take a look at the `bfd` or `hpms` modules. They don't have `SpringBootApp` under `src/main` because those modules are technically not executable Boot apps, but they _do_ have it under `src/test` to help with wiring in tests. We should probably follow the same approach.

optout/pom.xml
spathiyil: removed

optout/pom.xml
spathiyil: removed

optout/src/main/java/gov/cms/ab2d/optout/SpringBootApp.java
spathiyil: moved `SpringBootApp` to `src/test/java`

PR #60, opened 2019-12-02
Feature/ab2d 505 filter data

lailasharshar:
### Summary

This is used to filter from a Explanation of Benefit object to one that is filtered to only provide the relevant data to the part D providers.


filter/src/main/java/gov/cms/ab2d/filter/EOBToAB2DEOB.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Load the AB2dExplanationOfBenefit object from a JSON file
+     *
+     * @param path - the file name and path in the classpath containing the
+     *             Explanation of Benefit data retrieved from Blue
+     * @return the AB2D object
+     */
+    public static AB2DExplanationOfBenefit fromFile(String path) {
```
dkrylovsb: The worker will have an EOB resource in-memory, so we need an extra method, such as `fromReader(Reader eobReader`).

filter/src/main/java/gov/cms/ab2d/filter/EOBToAB2DEOB.java
```diff
@@ -0,0 +1,77 @@
...
+        System.out.println("Loading file " + path);
```
dkrylovsb: Let's use `slf4j` to make sure this message flows through the centralized logging.

filter/src/main/java/gov/cms/ab2d/filter/EOBToAB2DEOB.java
dkrylovsb: Let's rename this method to clearly communicate the fact that it really loads only from a _classpath_ resource, not an arbitrary file. `fromClasspath` or something along these lines.

filter/src/main/java/gov/cms/ab2d/filter/EOBToAB2DEOB.java
lailasharshar: This was actually a debug message that I forgot to delete while testing. Should I make it a log.debug message or just get rid of it?

dkrylovsb: Getting rid of it works for me 👍

PR #63, opened 2019-12-03
Feature/ab2d 520 delete files

adaykin:
A cron job that will delete files that are older than a configurable time period.

Adding in a configurable cron job that will run by default every hour. The job will look for files older than 24 hours by default to delete to avoid taking up too much space

### Checklist

- [x] Tests and linting pass

### Security

NA


audit/pom.xml
```diff
@@ -22,9 +22,9 @@
             <version>0.0.1-SNAPSHOT</version>
         </dependency>
         <dependency>
-            <groupId>org.quartz-scheduler</groupId>
-            <artifactId>quartz</artifactId>
-            <version>${quartz.version}</version>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-quartz</artifactId>
+            <version>2.2.1.RELEASE</version>
```
dkrylovsb: We should probably drop this to make sure Spring Boot version is inherited from the top-level POM. This line says `2.2.1` while the top-level POM still refers to `2.2.0`.

audit/src/main/java/gov/cms/ab2d/audit/job/FileDeletionJob.java
```diff
@@ -0,0 +1,47 @@
...
+    public static final String EFS_MOUNT = "efsMount";
+
+    public static final String AUDIT_FILES_TTL_HOURS = "auditFilesTTLHours";
+
+    /**
+     * Delete all files that are in the efs mount that are older than the TTL variable
+     * @param jobExecutionContext
+     */
+    @Override
+    public void execute(JobExecutionContext jobExecutionContext) {
```
dkrylovsb: Similar to how @spathiyil implemented the opt-out job, the `execute` method should simply delegate to a Spring Service Bean, e.g. `CleanupServiceBean` implementing `CleanupService` interface. That way you can easily avoid dealing with job data maps and achieve better separation of concerns.

audit/src/main/java/gov/cms/ab2d/audit/job/FileDeletionJob.java
```diff
@@ -0,0 +1,47 @@
...
+        try (Stream<Path> walk = Files.walk(Paths.get(jobExecutionContext.getJobDetail().getJobDataMap().getString(EFS_MOUNT)))) {
```
dkrylovsb: Let's build in some basic safety checks prior to/during the "walk". This is powerful code: imagine a developer sets `efs.mount` property to `/home/username` by mistake and then this job kicks in on their workstation...

Let's add code to ensure that:

- `efs.mount` value begins with `/` and is at least 5 characters long
- `efs.mount` does **not** start with any of the following: `/bin`,`/boot`,`/dev`,`/etc`,`/home`,`/lib`,`/opt`,`/root`,`/sbin`,`/sys`,`/usr`,`/Applications`,`/Library`,`/Network`,`/System`,`/Users`,`/Volumes`
- The walk _only_ deletes the files with `.ndjson` extension.

worker/src/main/java/gov/cms/ab2d/worker/config/AuditJobSetup.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public static final String JOB_NAME = "DeleteAuditFilesJob";
+
+    public static final String CRON_GROUP = "FileDeletion";
+
+    public static final String TRIGGER_NAME = "DeleteAuditFilesTrigger";
+
+    @PostConstruct
+    public void setupAuditJob() throws SchedulerException {
+        SchedulerFactory sf = new StdSchedulerFactory();
```
dkrylovsb: You can avoid dealing with low-level Quartz classes if you use Spring's facilities. Take a look at `QuartzSetup` that @spathiyil wrote.

worker/src/main/resources/application.properties
```diff
@@ -13,6 +13,18 @@ spring.jpa.properties.hibernate.temp.use_jdbc_metadata_defaults=false
 ## ------------------------------------------------------------------------------------  SPRING INTEGRATION CONFIG
 spring.integration.jdbc.initialize-schema=always
 
+## ------------------------------------------------------------------------------------  DELETE AUDIT FILES CONFIG
+# Run every hour http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html
+audit-files-cron=0 0 * * * ?
```
dkrylovsb: Let's move these to `application.audit.properties` within the `audit` module and load them up via `@PropertySource`. See `QuartzSetup.java` for an example.

worker/src/main/java/gov/cms/ab2d/worker/config/AuditJobSetup.java
```diff
@@ -0,0 +1,57 @@
...
+
+import javax.annotation.PostConstruct;
+
+import static org.quartz.CronScheduleBuilder.cronSchedule;
+import static org.quartz.JobBuilder.newJob;
+import static org.quartz.TriggerBuilder.newTrigger;
+
+@Slf4j
+@Component
+public class AuditJobSetup {
```
dkrylovsb: You could probably get rid of this class and just add your job config to `QuartzSetup`

worker/src/main/resources/application.properties
dkrylovsb: Also let's follow our convention of dot-separating parts of property names, i.e. `audit.files.cron`

audit/src/main/java/gov/cms/ab2d/audit/cleanup/FileDeletionServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    private static Set<String> disallowedDirectories = Set.of("/bin", "/boot", "/dev", "/etc", "/home", "/lib",
+            "/opt", "/root", "/sbin", "/sys", "/usr", "/Applications", "/Library", "/Network", "/System", "/Users", "/Volumes");
+
+    /**
+     * Delete all files that are in the efs mount with the .ndjson extension that are older than the TTL variable
+     */
+    @Override
+    public void deleteFiles() {
+        if (!efsMount.startsWith("/")) {
+            throw new EFSMountFormatException("EFS Mount must be start with a /");
```
dkrylovsb: typo: _must start_.

PR #64, opened 2019-12-04
Feature/ab2d 423 incorporate filtering module into worker

lailasharshar:
### Summary
This is actually about my 4th refactoring of the filtering code.

- My first attempt, I tried to just do an ModelMapper. I created an object that was a subset of a 
the ExplanationOfBenefit object and copied the full object into the smaller object. This in theory
should have worked, but because the ExplanationOfBenefit object is so recursively nested and
complicated, it gave a stack overflow error.
- I tried using XPath to filter data, but really, it's not made for what I was trying to use it for. It's
more for searching for values in the data, not so much removing object within other objects, or at least not easily.
- I eventually tried just creating an object that mirrored the existing object and copied the data I wanted into it. This seemed to work well, but when I tried to use it with the worker, I realized that all the tools to serialize/deserialize objects requires it to be a real FHIR object, not only in structure, but in name. You can't even inherit from an object. So, this time..
- I created a filter object that was subtractive. It copies the object and then nulls out all fields we don't want. In some ways, I really like this method. It keeps all the underlying structure of the FHIR object, including inherited data that we probably want, but also operates exactly like the ExplanationOfBenefit object in any context. One little snag is that whenever you call a get method on any FHIR object, it returns an empty object, instead of null. For example:
```
/**
 * @return {@link #enterer} (The person who created the explanation of benefit.)
 */
public Reference getEnterer() { 
  if (this.enterer == null)
    if (Configuration.errorOnAutoCreate())
        throw new Error("Attempt to auto-create ExplanationOfBenefit.enterer");
    else if (Configuration.doAutoCreate())
       this.enterer = new Reference(); // cc
    return this.enterer;
}
```
...which is frustrating because it makes it a pain to test because:

```
val.setEnterer(null);
assertNull(val.getEnterer())
```

...fails. I just did some spot checks to make sure it that it looked empty. I don't completely love this method because when you are subtracting data, when there is a bug, it will most likely be that you include data you shouldn't instead of forgetting data you should provide which isn't as secure but I didn't see an easy way around it.

### Checklist

- [ ] Verify an inherited members should be included
- [ ] Update Data Dictionary with any changes


filter/src/main/java/gov/cms/ab2d/filter/EOBLoadUtilities.java
```diff
@@ -0,0 +1,61 @@
...
+    public static ExplanationOfBenefit getEOBFromReader(Reader reader, FhirContext context) throws IOException {
+        if (reader == null) {
+            return null;
+        }
+        String inputStr;
+        StringBuilder responseStrBuilder = new StringBuilder();
+        BufferedReader bf = new BufferedReader(reader);
+        while ((inputStr = bf.readLine()) != null) {
+            responseStrBuilder.append(inputStr);
+        }
```
dkrylovsb: You could probably reduce these to a one-liner by using `IOUtils.toString(Reader reader)`

filter/src/main/java/gov/cms/ab2d/filter/ExplanationOfBenefitsTrimmer.java
```diff
@@ -0,0 +1,147 @@
...
+     * @param benefit - the original ExplanationOfBenefit
+     * @return the cleaned up copy
+     */
+    public static ExplanationOfBenefit getBenefit(ExplanationOfBenefit benefit) {
+        if (benefit == null) {
+            return null;
+        }
+        // Copy it so we don't destroy the original
+        ExplanationOfBenefit newBenfit = benefit.copy();
+        // Remove the unauthorized ata
```
dkrylovsb: typo

filter/src/main/java/gov/cms/ab2d/filter/ExplanationOfBenefitsTrimmer.java
```diff
@@ -0,0 +1,147 @@
...
+     *
+     * @param benefit - the original ExplanationOfBenefit
+     * @return the cleaned up copy
+     */
+    public static ExplanationOfBenefit getBenefit(ExplanationOfBenefit benefit) {
+        if (benefit == null) {
+            return null;
+        }
+        // Copy it so we don't destroy the original
+        ExplanationOfBenefit newBenfit = benefit.copy();
```
dkrylovsb: typo in variable's name

dkrylovsb:
Wondering what if we flip this switch: `Configuration.doAutoCreate()` ?

filter/src/main/java/gov/cms/ab2d/filter/EOBLoadUtilities.java
lailasharshar: That seems much easier :)

PR #66, opened 2019-12-05
Delete the ndjson file after it has been sent to the client

adaykin:
Delete the ndjson file after it has been sent to the client

Delete the ndjson file after it has been sent to the client using IOUtils and change the response type to void, but still send data.

### Checklist

- [x] Tests and linting pass

### Security

NA


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -222,19 +229,26 @@ private String getUrlPath(Job job, String filePath) {
     )
     @ResponseStatus(value = HttpStatus.OK)
     @GetMapping(value = "/Job/{jobUuid}/file/{filename}")
-    public ResponseEntity<Resource> downloadFile(
+    public void downloadFile(
```
dkrylovsb: This will likely break Swagger again. Let's try `ResponseEntity<Void>`. See if Swagger definition stays correct.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -222,19 +229,26 @@ private String getUrlPath(Job job, String filePath) {
...
+        OutputStream out = response.getOutputStream();
+        FileInputStream in = new FileInputStream(downloadResource.getFile());
+
+        IOUtils.copy(in, out);
 
         log.info("Sending file to client");
 
-        return new ResponseEntity<>(downloadResource, headers, HttpStatus.OK);
+        out.close();
+        in.close();
```
dkrylovsb: Let's use a `try-with-resources` statement to make sure both streams get closed in case of an exception. Helps to avoid having to call `close` explicitly, too.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -222,19 +229,26 @@ private String getUrlPath(Job job, String filePath) {
...
-        headers.set(HttpHeaders.CONTENT_TYPE, NDJSON_FIRE_CONTENT_TYPE);
+        RequestAttributes requestAttributes = RequestContextHolder.getRequestAttributes();
+        HttpServletResponse response = ((ServletRequestAttributes) requestAttributes).getResponse();
+        response.setContentType(NDJSON_FIRE_CONTENT_TYPE);
+        OutputStream out = response.getOutputStream();
+        FileInputStream in = new FileInputStream(downloadResource.getFile());
+
+        IOUtils.copy(in, out);
 
         log.info("Sending file to client");
```
dkrylovsb: The logging statement should precede the copy.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -222,19 +229,26 @@ private String getUrlPath(Job job, String filePath) {
...
 
-        return new ResponseEntity<>(downloadResource, headers, HttpStatus.OK);
+        out.close();
+        in.close();
+        downloadResource.getFile().delete();
```
dkrylovsb: Let's delegate to `jobService.deleteResourceForJob(jobUuid, filename)` here. It's best to free the controller from dealing with file management etc and let the Service layer encapsulate all that logic.

`deleteResourceForJob(jobUuid, filename)` would just be a one-liner such as `this.getResourceForJob(jobUuid, filename).getFile().delete()`.

adaykin: I moved to the service with the file, I think I have to perform the delete after I've gathered the FileInputStream since it's used in copying the contents of the file.

dkrylovsb: Correct. `deleteResourceForJob(jobUuid, filename)` would be the last line in the controller's method. It won't be reached if copying errors out.

PR #72, opened 2019-12-06
Feature/ab2d 498 secret scanning

casey-erdmann:
### Adds a github action to automatically scan for leaked secrets

Adds a github action to automatically scan for leaked secrets in repo and then notify the security team of any findings

### Checklist

-  [X] Code checked for PHI/PII exposure

### Security

This change does not actively modify or change any security related functionality/data handling within the application. It does however improve our security posture over all by locating any potentially leaked sensitive keys that should not exist within the repository.

AB2D-492
AB2D-496


casey-erdmann:
This will pull excludes for tuning out of master, so currently the Action build is "broken" until this PR is merged. Otherwise it should only fail if a finding is observed. There is no need to make this a required status check, and at first there will be a lot of false positives. Future PRs will be submitted to update the excludes list as needed.

PR #77, opened 2019-12-09
Feature/ab2d 403 test containers

adaykin:
Adding in test containers

Add in a test container that gets built prior to the test running and has the same version of postgres running that is running in prod. The tests will use the container and testcontainers annotations to utilize the container. Upgrading some parts to junit 5 (jupiter)

### Checklist

- [x] Tests and linting pass

### Security

NA


common/src/test/java/gov/cms/ab2d/common/util/AB2DPostgresqlContainer.java
```diff
@@ -0,0 +1,34 @@
...
+    public void start() {
+        super.start();
+        System.setProperty("DB_URL", container.getJdbcUrl());
+        System.setProperty("DB_USERNAME", container.getUsername());
+        System.setProperty("DB_PASSWORD", container.getPassword());
+    }
+
+    @Override
+    public void stop() {
+        //do nothing, JVM handles shut down
```
dkrylovsb: We should set `container` to null here.

dkrylovsb: We should call `super.stop()` I think.

common/src/test/java/gov/cms/ab2d/common/util/AB2DPostgresqlContainer.java
```diff
@@ -0,0 +1,34 @@
...
+    private AB2DPostgresqlContainer() {
+        super(IMAGE_VERSION);
+    }
+
+    public static AB2DPostgresqlContainer getInstance() {
+        if (container == null) {
+            container = new AB2DPostgresqlContainer();
+        }
+        return container;
+    }
```
dkrylovsb: Do we need to reuse the container across tests? If we use `new AB2DPostgresqlContainer()` instead in the tests, we will make sure each test starts with a clean DB.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -34,7 +34,7 @@
 import static net.logstash.logback.argument.StructuredArguments.keyValue;
 
 @Slf4j
-@Service
+@Service()
```
dkrylovsb: What's this?

adaykin: I think I was playing around with that when I was trying to get the worker module to work with JUnit5, I can revert

PR #78, opened 2019-12-10
AB2D-498 Filter out all authorized_keys entries with regex

casey-erdmann:
### Filter out all authorized_keys entries with regex

New false positive was reported by a commit to the branch:
`feature/ab2d-355-automate-aws-waf-and-shield-configuration-deployment`
This PR filters out all authorized_keys entries with regex and filters additional false positives that were identified during analysis of the results. We won't be able to exclude every file that comes along, but this one highlighted a key area where we can drastically reduce false positives that flag on authorized keys and is worth updating our excludes for.

### Checklist

### Security
This improves the security posture of the code base by reducing false positives for our secrets scanner. Public keys like authorized_keys are high entropy strings, but are not secret and do not need to be alerted on.

### DIFF
--- a/trufflehog-excludes.txt
+++ b/trufflehog-excludes.txt
@@ -1,11 +1,6 @@
 .travis.yml
-Deploy/terraform/environments/ab2d-dev/authorized_keys
-Deploy/terraform/environments/ab2d-impl/authorized_keys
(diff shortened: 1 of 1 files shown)


casey-erdmann:
For some reason that regex did not work as intended...I'm going to do some more testing on a new repo and see what the issue is.

PR #79, opened 2019-12-10
AB2D 498 Regex Secrets Filter

casey-erdmann:
### Fixes regular expression account for literally all authorized_keys files

The original fix included with PR #78 had a regular expression that only tested for directories nested 1 folder deep. This new regex contains no directory constraints and checks for all instances of the authorized keys file.

### Checklist

### Security
Fixes the security regression caused by a bad regex in PR #78, impact is the same as the original PR #78

### DIFF
--- a/trufflehog-excludes.txt
+++ b/trufflehog-excludes.txt
@@ -1,9 +1,5 @@
 .travis.yml
-<<<<<<< HEAD
 (.*/)?authorized_keys
-=======
-(.*/)?/authorized_keys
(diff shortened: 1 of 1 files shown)


casey-erdmann:
Sanity check gif before I merge showing output is all clear. (Just in case this one comes back to bite me too!)

trufflehog_test

casey-erdmann:
All green! 
Screen Shot 2019-12-10 at 12 09 55 PM

PR #82, opened 2019-12-12
Feature/ab2d 610 endpoint

adaykin:
Adding an endpoint for kicking off a job based on the contract number

Adding in an endpoint to kick off a job based on contract number, adding some testing utilities, adding a migration so that contract can be tracked in the job table if it's used. Put in checks to ensure that a contract is attested on and connected to the user's sponsor.

### Checklist

- [x] Tests and linting pass

### Security

NA


common/src/main/java/gov/cms/ab2d/common/model/Job.java
```diff
@@ -25,6 +19,7 @@
 @Getter
 @Setter
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
+@Configurable(autowire = Autowire.BY_TYPE)
```
spathiyil: This seems strange to me. 
IMHO, an entity should almost never be polluted with spring package names.
Spring beans are singleton unlike entities.
I am not sure what is being autowired here, but I do not think that entities should be aware of Spring (or any dependency injection framework used).

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIBadData.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Test
+    public void testPatientExportWithContractNotTiedToUser() throws Exception {
+        String token = testUtil.setupBadSponsorUserData(List.of(SPONSOR_ROLE));
+        Optional<Contract> contractOptional = contractRepository.findContractByContractNumber(BAD_CONTRACT_NUMBER);
+        Contract contract = contractOptional.get();
+        this.mockMvc.perform(get(API_PREFIX + FHIR_PREFIX + "/Group/" + contract.getContractNumber() + "/$export")
+                .contentType(MediaType.APPLICATION_JSON)
+                .header("Authorization", "Bearer " + token))
+                .andExpect(status().is(500))
```
dkrylovsb: This should be `403` since this is a lack of authorization issue. 500 is an internal server error.

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIBadData.java
```diff
@@ -0,0 +1,100 @@
...
+    @Test
+    public void testPatientExportWithNoAttestation() throws Exception {
+        // Valid contract number for sponsor, but no attestation
+        String token = testUtil.setupContractWithNoAttestation(List.of(SPONSOR_ROLE));
+        Optional<Contract> contractOptional = contractRepository.findContractByContractNumber(VALID_CONTRACT_NUMBER);
+        Contract contract = contractOptional.get();
+        this.mockMvc.perform(get(API_PREFIX + FHIR_PREFIX + "/Group/" + contract.getContractNumber() + "/$export")
+                .contentType(MediaType.APPLICATION_JSON)
+                .header("Authorization", "Bearer " + token))
+                .andExpect(status().is(500))
```
dkrylovsb: Also a 403.

common/src/main/java/gov/cms/ab2d/common/model/Job.java
adaykin: I was playing around with trying to put in some validation here, I forgot to remove this.

dkrylovsb: JPA classes should not have this.

common/src/main/java/gov/cms/ab2d/common/model/Job.java
```diff
@@ -69,6 +64,10 @@
     @Pattern(regexp = "ExplanationOfBenefits", message = "_type should be ExplanationOfBenefits")
     private String resourceTypes; // for now just limited to ExplanationOfBenefits
 
+    @OneToOne()
```
dkrylovsb: Is this not a One-to-Many from Contract to Job? Technically, multiple `Job` records may refer to the same `Contract` as long as only one of those Jobs is in ACTIVE status.

common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -48,6 +59,28 @@ public Job createJob(String resourceTypes, String url) {
         job.setProgress(0);
         job.setUser(userService.getCurrentUser());
 
+        if (contractNumber != null) {
+            contractRepository.findContractByContractNumber(contractNumber).ifPresentOrElse(contractFound -> {
+                User user = userService.getCurrentUser();
+                Sponsor userSponsor = user.getSponsor();
+                Sponsor contractSponsor = contractFound.getSponsor();
+                if (!userSponsor.getHpmsId().equals(contractSponsor.getHpmsId())) {
```
dkrylovsb: This will not work if the contract is assigned to a "child" organization,  but the user is from a "parent" organization, which is a different Sponsor record, but still has access to that contract.

common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -48,6 +59,28 @@ public Job createJob(String resourceTypes, String url) {
...
+                    throw new ContractHasNotBeenAttestedException("Contract " + contractFound.getContractNumber() + " has not been attested for");
+                }
+
+                job.setContract(contractFound);
+            }, () -> {
+                log.error("Contract {} was not found", contractNumber);
+                throw new ResourceNotFoundException("Contract " + contractNumber + " was not found");
+            });
+        }
+
```
dkrylovsb: You can greatly simplify this if you use `userSponsor.getAggregatedAttestedContracts()`.  This will give all of the attested contracts for the Sponsor taking care of parent/child stuff. 

Of course you will not be able to tell whether the contract number if wrong, or whether it exists but is unattested. I think that's fine and actually makes our error handling simpler. Just throw a general `No attested contract with contract number NNNN found for the user`.

common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -48,6 +59,28 @@ public Job createJob(String resourceTypes, String url) {
...
         job.setUser(userService.getCurrentUser());
 
+        if (contractNumber != null) {
+            contractRepository.findContractByContractNumber(contractNumber).ifPresentOrElse(contractFound -> {
+                User user = userService.getCurrentUser();
+                Sponsor userSponsor = user.getSponsor();
+                Sponsor contractSponsor = contractFound.getSponsor();
+                if (!userSponsor.getHpmsId().equals(contractSponsor.getHpmsId())) {
+                    log.error("User {} does not have permissions for contract {}", user.getUsername(), contractFound.getContractNumber());
+                    throw new UserPermissionsException("User " + user.getUsername() + " does not have permissions for contract " + contractFound.getContractNumber());
```
dkrylovsb: `Username` could be an email address, which is PII: let's suppressed it from the exception message.

common/src/test/java/gov/cms/ab2d/common/service/JobServiceTest.java
```diff
@@ -56,6 +61,15 @@
...
     SponsorRepository sponsorRepository;
 
+    @Autowired
+    ContractRepository contractRepository;
+
+    @Autowired
+    DataSetup dataSetup;
+
+    //@MockBean
+    //UserService userService;
```
dkrylovsb: Let's remove if not needed.

common/src/main/java/gov/cms/ab2d/common/model/Job.java
adaykin: Yes it should be one to many from contract to job since those old job records would still be persisted.

common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -48,6 +61,25 @@ public Job createJob(String resourceTypes, String url) {
         job.setProgress(0);
         job.setUser(userService.getCurrentUser());
 
+        if (contractNumber != null) {
+            contractRepository.findContractByContractNumber(contractNumber).ifPresentOrElse(contractFound -> {
+                User user = userService.getCurrentUser();
+                Sponsor userSponsor = user.getSponsor();
+
+                List<Contract> contracts = userSponsor.getAggregatedAttestedContracts();
+                if (!contracts.contains(contractFound)) {
```
dkrylovsb: Since `contains` method will check for object equality, we need to add both `equals` and `hashCode` to `Contract`. Let's do so using Lombok: the only field that should be included into equality check is `id`.

PR #84, opened 2019-12-13
AB2D 609 handle job cancellation in worker

spathiyil:
Add functionality to the worker module to enable it to stop processing an `IN_PROGRESS` job mid-way if the job gets cancelled

### Summary

While the working is processing a submitted job, if the job gets cancelled, the worker needs to be aware of it and attempt to cancel the  processing of the job, if possible.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
@@ -1,14 +1,14 @@
 package gov.cms.ab2d.worker.service;
 
-import gov.cms.ab2d.common.model.OptOut;
 import gov.cms.ab2d.common.model.Contract;
 import gov.cms.ab2d.common.model.Job;
 import gov.cms.ab2d.common.model.JobOutput;
 import gov.cms.ab2d.common.model.JobStatus;
+import gov.cms.ab2d.common.model.OptOut;
 import gov.cms.ab2d.common.model.Sponsor;
-import gov.cms.ab2d.common.repository.OptOutRepository;
 import gov.cms.ab2d.common.repository.JobOutputRepository;
 import gov.cms.ab2d.common.repository.JobRepository;
+import gov.cms.ab2d.common.repository.OptOutRepository;
 import gov.cms.ab2d.worker.adapter.bluebutton.BeneficiaryAdapter;
 import gov.cms.ab2d.worker.adapter.bluebutton.GetPatientsByContractResponse;
 import gov.cms.ab2d.worker.adapter.bluebutton.PatientClaimsProcessor;
@@ -109,10 +109,6 @@ public class JobProcessingServiceImpl implements JobProcessingService {
             completeJob(job);
 
         } catch (JobCancelledException e) {
-            job.setStatus(CANCELLED);
-            job.setStatusMessage(e.getMessage());
-            job.setExpiresAt(OffsetDateTime.now().plusDays(1));
-            jobRepository.save(job);
             log.warn("Job: [{}] CANCELLED", jobUuid);
         } catch (Exception e) {
             job.setStatus(JobStatus.FAILED);
@@ -159,7 +155,7 @@ public class JobProcessingServiceImpl implements JobProcessingService {
             if (recordsProcessedCount % cancellationCheckFrequency == 0) {
                 errorCount += processHandles(futureResourcesHandles);
 
-                // A Job could run for a long time perhaps hours.
+                // A Job could run for a long time, perhaps hours.
                 // While the job is in progress, the job could be cancelled.
                 // So the worker needs to periodically check the job status to ensure it has not been cancelled.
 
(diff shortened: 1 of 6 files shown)


worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -79,44 +83,51 @@ public Job putJobInProgress(String jobId) {
...
                 jobOutputRepository.saveAll(jobOutputs);
             }
 
             completeJob(job);
 
+        } catch (JobCancelledException e) {
+            job.setStatus(CANCELLED);
+            job.setStatusMessage(e.getMessage());
+            job.setExpiresAt(OffsetDateTime.now().plusDays(1));
+            jobRepository.save(job);
```
dkrylovsb: We don't really need these. API has already updated the job record.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -128,19 +139,52 @@ public Job processJob(final String jobId) {
         // A mutex lock that all threads for a contract uses while writing into the shared files
         var lock = new ReentrantLock();
 
-        final List<Future<Integer>> futureResourcesHandles = patients.stream()
-                .map(patient -> patient.getPatientId())
-                .filter(patientId -> !isOptOutPatient(patientId))
-                .map(patientId -> patientClaimsProcessor.process(patientId, lock, outputFile, errorFile))
-                .collect(Collectors.toList());
```
dkrylovsb: Sad to see this nice stream-oriented code go

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -128,19 +139,52 @@ public Job processJob(final String jobId) {
...
+
+            }
+        }
 
 
-        int errorCount = processHandles(futureResourcesHandles);
         while (!futureResourcesHandles.isEmpty()) {
             sleep();
             errorCount += processHandles(futureResourcesHandles);
         }
```
dkrylovsb: Even If the job has been cancelled, the flow still goes into this `while` loop that's going to _wait_ for all the Futures to get done. That's the opposite of what we are trying to achieve.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -128,19 +139,52 @@ public Job processJob(final String jobId) {
...
 
+        if (jobHasBeenCancelled(jobStatus)) {
+            final String errMsg = "Job was cancelled while it was being processed";
+            log.warn("{} - JobUuid :[{}]", errMsg, jobUuid);
+            throw new JobCancelledException(errMsg);
```
dkrylovsb: An exception in this method will not really stop all the Futures from being processed by the Thread Pool. They will continue to be executed in the background.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -128,19 +139,52 @@ public Job processJob(final String jobId) {
         // A mutex lock that all threads for a contract uses while writing into the shared files
         var lock = new ReentrantLock();
```
dkrylovsb: Based on the other comments, it appears we need to communicate the job cancellation "event" from the "main" worker thread to the `PatientClaimsProcessor`'s Futures.

We could use something as simple as this:

- Create an `AtomicBoolean jobStillActive` variable right next to the `lock` and set it to `true`
- `gov.cms.ab2d.worker.adapter.bluebutton.PatientClaimsProcessorImpl#process` will need a new parameter to take it
- Whenever the Worker detects a job cancellation, it should flip the value to `false` in addition to throwing an exception as it does now.
- The first thing `gov.cms.ab2d.worker.adapter.bluebutton.PatientClaimsProcessorImpl#process` should do is check the value of the flag. If `false`, return silently with an empty Future, do nothing else.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: Here is what I was thinking.

Assume the contract had 10_000 patients.
The process adds the first `n` patient records (say 10 records = cancellationCheckFrequency) 
to the futures list. 
Every 10 records, it will check the job table to see if it has been cancelled.
If not, it releases whichever future is already done and proceeds with the next 10 records...
check for cancellation again, then proceed with the next 10 records etc.

So, if a cancellation happens, say after the 50th record was processed, there could be about 10 outstanding Futures (give or take a few)  at that point.
So it lets the already issued threads drain, checks their isDone and completes the Futures gracefully.
It would not process any new patient records as it is already in the JobCancellation logic flow.

*TLDR* 
In the `while (!futureResourcesHandles.isEmpty())` loop,  I expect approximately 10 (~= cancellationCheckFrequency, give or take a few) futures to be present - which I figured may be alright to wait on.

worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
spathiyil: removed

PR #85, opened 2019-12-13
Adding logging configuration in order to rely on an environment varia…

adaykin:
Adding logging configuration in order to rely on an environment variable (json-logging) to remove the logback appenders in JSON format and add a new console appender.

Adding logging configuration in order to rely on an environment variable (json-logging) to remove the logback appenders in JSON format and add a new console appender.

### Checklist

- [x] Tests and linting pass

### Security

NA

### DIFF
--- a/common/src/main/java/gov/cms/ab2d/common/config/LoggingConfig.java
+++ b/common/src/main/java/gov/cms/ab2d/common/config/LoggingConfig.java
@@ -29,7 +29,7 @@ public class LoggingConfig {
         String jsonLogging = env.getProperty("json-logging");
 
         // Developers must add the json-logging env variable to their configuration and set it to false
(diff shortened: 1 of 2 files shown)


common/src/main/java/gov/cms/ab2d/common/config/LoggingConfig.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Autowired
+    private Environment env;
+
+    @PostConstruct
+    public void init() {
+        String jsonLogging = env.getProperty("json-logging");
+
+        // Developers must add the json-logging env variable to their configuration and set it to true
+        if(jsonLogging != null && jsonLogging.equals("true")) {
```
dkrylovsb: Should be `false` since want to _disable_ JSON logging.

PR #87, opened 2019-12-16
enhance worker to process export job for a specific contract

spathiyil:
enhance worker to process export job for a specific contract

### Summary

When a job is submitted for a specific contract, enhance the worker to process the export file for that contract only

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
@@ -24,6 +24,7 @@ import java.nio.file.Paths;
 import java.time.LocalDate;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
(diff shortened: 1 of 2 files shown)


worker/src/main/java/gov/cms/ab2d/worker/service/JobProcessingServiceImpl.java
```diff
@@ -115,6 +113,24 @@ public Job processJob(final String jobId) {
...
     }
 
+    private List<Contract> getAttestedContracts(Job job) {
+
+        // when the job is submitted for a specific contract, process the export for only that contract.
+        final Contract jobSpecificContract = job.getContract();
+        if (jobSpecificContract != null && jobSpecificContract.getAttestedOn() != null) {
+            log.info("Job [{}] submitted for a specific attested contract [{}] ", jobSpecificContract.getContractNumber());
+            return Collections.singletonList(jobSpecificContract);
+        }
```
spathiyil: What needs to happen if the contract specified in the job is not attested?
Should we return an error indicating the specified contract is not attested and fail the job?

dkrylovsb: Attestation check isn't needed here: it is being done at job creation time.

PR #90, opened 2019-12-18
prevent concurrent execution of quartz job in worker

spathiyil:
prevent concurrent execution of quartz job in worker

### Summary

Occasionally, when the worker is started, it appeared as if the opt-out was being triggered twice concurrently on start-up for the first run. Adding an extra configuration to prevent it.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


dkrylovsb:
@adaykin @spathiyil Can we make sure all our Quartz jobs have this?

spathiyil:
> @adaykin @spathiyil Can we make sure all our Quartz jobs have this?

I will push out a commit with the same changes to the `audit` sub-module shortly.

PR #92, opened 2019-12-18
Feature/ab2d 553 create

adaykin:
Adding ability to create a user via the API.

### Summary

Adding an endpoint to the Admin controller so that users with the Admin role can update users via an API call. Adding in ModelMapper to help map DTO objects to entities, to ensure we send/receive the correct data.

### Checklist

- [x] Tests and linting pass

### Security

NA


api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIUserTests.java
```diff
@@ -0,0 +1,159 @@
+package gov.cms.ab2d.api.controller;
```
dkrylovsb: Need a test that ensures only Admins can create users.

common/src/main/java/gov/cms/ab2d/common/dto/Mapping.java
```diff
@@ -0,0 +1,16 @@
+package gov.cms.ab2d.common.dto;
```
dkrylovsb: This class makes sense only if `ModelMapper` is expensive (slow) to create. Furthermore, `ModelMapper` must be thread-safe to be used as a Singleton.

Unless instantiating `ModelMapper` is really slow, I'd say let's drop this class and use `new ModelMapper` wherever it is needed.

common/src/main/java/gov/cms/ab2d/common/dto/RoleDTO.java
```diff
@@ -0,0 +1,12 @@
...
+
+import lombok.Data;
+
+import javax.validation.constraints.NotNull;
+
+@Data
+public class RoleDTO {
+
+    @NotNull
+    private Long id;
```
dkrylovsb: This means the API caller will have to know the database ID of the role, which requires a database lookup prior to using the API. Let's just use the role name.

common/src/main/java/gov/cms/ab2d/common/dto/UserDTO.java
```diff
@@ -0,0 +1,30 @@
...
+    private String username;
+
+    @NotNull
+    private String firstName;
+
+    @NotNull
+    private String lastName;
+
+    @NotNull
+    private String email;
```
dkrylovsb: These should be nullable in case of a system-to-system user.

common/src/main/java/gov/cms/ab2d/common/dto/UserDTO.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @NotNull
+    private Boolean enabled;
+
+    @NotNull
+    private Long sponsorId;
```
dkrylovsb: Again, this requires a DB lookup prior to using the API. Can we use Sponsor HPMS ID, since it is readily available in the HPMS report? I guess not, since there were collisions between  HMPS ID for Parent & Child orgs, if I remember right. How about a combination of HPMS ID and Organization Name then?

common/src/main/java/gov/cms/ab2d/common/dto/UserDTO.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private Set<RoleDTO> roles;
```
dkrylovsb: We won't have users with more than one role. Should we change this to just `String role`?

adaykin: Is it that only users created from the API won't have more than one role or all users in general? The User model now has users - roles as a many to many

dkrylovsb: All users. Many-to-many covers possible future changes

common/src/main/java/gov/cms/ab2d/common/dto/UserDTO.java
adaykin: Child and parent hpms was what I used before so that should work.

common/src/main/java/gov/cms/ab2d/common/dto/RoleDTO.java
adaykin: Yeah I think I'll have to look into using some sort of provider or converter in model mapper to accomplish the lookup

common/src/main/java/gov/cms/ab2d/common/dto/UserDTO.java
dkrylovsb: Could you check whether the combination of HPMS ID and Organization Name is unique? I'd like to avoid having to pass parent/child flags to this API.

adaykin: Oh yeah the hpms/org name is unique.

common/src/main/java/gov/cms/ab2d/common/dto/Mapping.java
adaykin: I think with the recent changes that require mappings and converters it's a little more expensive to create so it makes sense to have this class be a bean since it requires other components that we will use. It's thread safe since it's using concurrent hash maps where it keeps state and other places that keep state are done just at the object creation time.

PR #94, opened 2019-12-20
Added additional environment variables to support Dev/Sandbox roll-out

dkrylovsb:
Added additional environment variables to support Dev/Sandbox roll-out

### Summary

Added environment variables that will hold EFS mount and opt-out file S3 location information with default fallback to values that would work in Dex or Sandbox. Update Dockerfile to call them out explicitly. 

### Checklist

- [x] Tests and linting pass

### Security
N/A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ services:
       - AB2D_DB_DATABASE=ab2d
       - AB2D_DB_USER=ab2d
       - AB2D_DB_PASSWORD=ab2d
+      - AB2D_EFS_MOUNT=/tmp/ab2d_efs_mount
     ports:
       - "8080:8080"
     depends_on:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -25,6 +25,7 @@ services:
       - AB2D_DB_DATABASE=ab2d
       - AB2D_DB_USER=ab2d
       - AB2D_DB_PASSWORD=ab2d
+      - AB2D_EFS_MOUNT=/tmp/ab2d_efs_mount
```
adaykin: Is there some other script responsible for creating the ab2d_efs_mount directory or does it get created automatically somehow?

dkrylovsb: @adaykin Worker does I believe.

PR #96, opened 2019-12-23
ab2d 651 refactor job processing service

spathiyil:
Refactor job processing service 

### Summary

Refactor job processing service

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


worker/src/main/java/gov/cms/ab2d/worker/service/FileServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+
+    @Override
+    public Path createFile(final Path outputDir, final String filename) {
+        final Path filePath = Path.of(outputDir.toString(), filename);
+        Path outputFile = null;
+        try {
+            //Delete file from previous run - Allows for a job can be restarted on failure.
+            Files.deleteIfExists(filePath);
```
adaykin: You should check the result of this since it returns a boolean and then handle the error if it's false

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    private final JobRepository jobRepository;
+
+    @Override
+    @Transactional(propagation = Propagation.REQUIRES_NEW, isolation = Isolation.SERIALIZABLE)
+    public Job preprocess(String jobUuid) {
+
+        final Job job = jobRepository.findByJobUuid(jobUuid);
+        if (job == null) {
+            log.error("Job was not found");
+            throw new IllegalArgumentException("Job " + jobUuid + " was not found");
```
adaykin: I think this should actually be a ResourceNotFoundException found in the common module instead, IllegalArgumentException could conflict with other exceptions thrown from the JDK or other libraries.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -60,31 +59,9 @@
...
-        return jobRepository.save(job);
-    }
-
     @Override
     @Transactional(propagation = Propagation.NEVER)
-    public Job processJob(final String jobUuid) {
+    public Job process(final String jobUuid) {
 
         final Job job = jobRepository.findByJobUuid(jobUuid);
         log.info("Found job : {}", jobUuid);
```
adaykin: You shouldn't need to put the jobUuid here since it's already coming from JobHandler line 46 with MDC.put

worker/src/test/java/gov/cms/ab2d/worker/processor/JobPreProcessorIntegrationTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+        var exceptionThrown = assertThrows(
+                IllegalArgumentException.class,
+                () -> cut.preprocess("S001"));
+
+        assertThat(exceptionThrown.getMessage(), is("Job S001 is not in SUBMITTED status"));
+    }
+
+
+    private Sponsor createSponsor() {
```
adaykin: Just an FYI there's already code in the common module in TestUtil to do some of this stuff. You can inject it into classes like this.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobPreProcessorUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+        var processedJob = cut.preprocess(jobUuid);
+
+        assertThat(processedJob.getStatus(), is(JobStatus.IN_PROGRESS));
+        verify(jobRepository).findByJobUuid(Mockito.any());
+        verify(jobRepository).save(Mockito.any());
+    }
+
+
+    private Job createJob() {
```
adaykin: If some of this stuff is the same from class to class you can put it in DataSetup in common module.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
spathiyil: I do not know if I agree with this comment.
If fact, in `Effective Java -  2nd ed`, *Joshua Bloch* recommends using the it.  _Item 60 - Favor the user of standard exceptions_  (page 248-9).
https://github.com/tatsuyaoiw/effective-java/blob/master/chapter-9.md

Over here, the scenario is that the method was provided a parameter which was bad data and hence inappropriate. 

Also, it is fine to call it is a Resource in the context of a Rest API so that the exception handling can handle the exception as needed when a resource is not found. However, since the worker does not provide a rest endpoint, and is mostly batch oriented, I think it is better to keep it generic and not call it a resource in this context.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobPreProcessorIntegrationTest.java
spathiyil: I considered using the object-mother with factory methods for creating the test data, but in the end, I prefer to set up my test data in the test class so that it can be specific to the test case instead of using a generic test data. Also, I feel it is a bit odd to create a spring bean just for creating test data. In addition to being unexpected, it also has the potential to become a god object over time if one keeps injecting all the repositories into it. Am going to leave this as is for now, but am happy to revisit this at a later date, if needed.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobPreProcessorUnitTest.java
spathiyil: replied to the comment in my previous response. thanks

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
spathiyil: removed jobUuid from the above log statement.

worker/src/main/java/gov/cms/ab2d/worker/service/FileServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+            final String errMsg = "Could not create output directory : ";
+            log.error("{} : {}", errMsg, outputDir.toAbsolutePath());
+            throw new UncheckedIOException(errMsg + outputDir.getFileName(), e);
+        }
+        return outputDirectory;
+    }
+
+
+    @Override
+    public Path createFile(final Path outputDir, final String filename) {
```
lailasharshar: Since I didn't realize you'd delete the file if it existed from the name, maybe call it overwriteFile? And then give an error for not being able to replace file or create file.

spathiyil: So, my understanding of the scenario is this. This should ordinarily not happen.

At the beginning of the batch process, it will create a directory (using the jobUuid) and an output file inside it.

If the batch were to fail due to some reason and is re-run, the file would exist from the _failed run_. 
To cover this scenario, Denis suggested that I put the `Files.deleteIfExists`  line. It currently does throw an `UncheckedIOException`  if it could not create the output file

lailasharshar: I have no problem with the logic, just the name. The default behavior of "create" is to fail if the file already exists. This is more of a createOrReplace. If anyone else wants to use this method, the method name is misleading.

spathiyil: I see what you mean. I've renamed the method to `createOrReplaceFile`

lailasharshar: Thanks.

PR #97, opened 2019-12-23
Upgrading to java 13

adaykin:
Upgrading to java 13

### Summary

Changing several configuration files to point to Java 13 (13.0.1) which is the LTS version of java that we want to support. Did not include the PMD plugin because that is not yet compatible with Java 13.

### Checklist

- [x] Tests and linting pass

### Security

NA


lhanekam:
@adaykin: I updated the "Uninstall any existing versions of Java" section of the "Deploy/documentation/Mac_Developer_SemanticBits.md" document in your branch. With that change, it looks good to me.

casey-erdmann:
After further inspection for the LGTM failures the commit `ce8470b3c749d6baccdc12fee216215d3d1540e8` actually has passed LGTM build checks, but is failing on the merged commit and master commit (see screenshot below). 

LGTM's biggest complaint is that the master branch is still looking for Java 12. This PR upgrades us to Java 13, and because this is actually passing on the travis-ci builds I am manually merging this in to see if it will resolve the build errors for LGTM.
Screen Shot 2019-12-30 at 10 51 41 AM

PR #98, opened 2019-12-23
Feature/ab2d 554 modify user v2

adaykin:
### Checklist

- [ ] Tests and linting pass

### Security


common/src/main/java/gov/cms/ab2d/common/service/UserServiceImpl.java
```diff
@@ -34,12 +36,30 @@ public User getCurrentUser() {
...
+            String userNotPresentMsg = "User is not present in our database";
+            log.error(userNotPresentMsg);
+            throw new ResourceNotFoundException(userNotPresentMsg);
+        } else {
+            return user;
+        }
     }
 
     @Override
-    public User createUser(UserDTO userDTO) {
```
spathiyil: Should both of these public methods (create & update) have an `@Transactional` annotation to make sure all the queries happen within the same transaction?

common/src/main/java/gov/cms/ab2d/common/service/UserServiceImpl.java
```diff
@@ -34,12 +36,30 @@ public User getCurrentUser() {
...
+        return mapping.getModelMapper().map(createdUser, UserDTO.class);
+    }
+
+    @Override
+    public UserDTO updateUser(UserDTO userDTO) {
+        // When updating there needs to be verification that the user exists
+        User user = getUserByUsername(userDTO.getUsername());
+        userDTO.setId(user.getId());
+        User mappedUser = mapping.getModelMapper().map(userDTO, User.class);
+        User updatedUser = userRepository.saveAndFlush(mappedUser);
```
spathiyil: Are you creating a new user entity here?
It seems to me that a brand new User entity is being created by the mapping library from userDTO and being saved in line 61 & 62. 
Since the user object was already fetched by hibernate in line 59 above, shouldn't the properties in the managed entity be getting updated from the DTO instead?

adaykin: No, it uses ID of the user it has found

common/src/main/java/gov/cms/ab2d/common/service/UserServiceImpl.java
adaykin: They happen during separate API calls.

adaykin: Actually I thought the service had one, I can put it there

PR #99, opened 2019-12-26
AB2D-660 Upgrade lgtm config to build with Java 13

casey-erdmann:
### Upgrade lgtm config to build with Java 13

The AB2D project is upgrading to Java 13, and as such the LGTM config needs to be reconfigured to use Java 13 instead of Java 12. This PR updates the lgtm config to build with Java 13.

### Checklist

### Security
This helps maintain our current security posture by insuring LGTM is configured correctly so it can continue to build the project and perform SAST scans.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,4 +1,4 @@
 extraction:
   java:
     index:
-      java_version: 12
+      java_version: 13


casey-erdmann:
Looks like the LGTM Java Analysis is only a partial failure on an strange commit I can't locate (hash prefix is 953a1e3).

Otherwise it is working as intended:
Screen Shot 2019-12-26 at 11 48 51 AM

PR #100, opened 2019-12-26
AB2D-657 Implement Gitleaks for Secret Scanning

casey-erdmann:
### This PR replaces Trufflehog with Gitleaks for secret scanning

This PR replaces Trufflehog with Gitleaks for secret scanning. It also implements a more de-coupled exclusion system allowing security engineers to more dynamically modify exclusions without interrupting the SDLC cycle. This is done by simply hosting the `gitleaks.toml` configuration file externally (in this case at https://github.com/casey-erdmann/ab2d-gitleaks-automation). This can be changed any time by modifying the github action should the project receive new maintainers in the future, and is as simple as pointing to a new location that hosts the exclusion list.

### Checklist

### Security

This tooling switch from Trufflehog to Gitleaks improves our security posture by drastically increasing the accuracy and focus of secret scans, and reducing the number of false positives produced significantly. This also allows for quicker action to remediate future false positives in a more effective manner. This does now add a external dependency due to the external `gitleaks.toml` configuration file being hosted externally, but this is mitigated by leaving full control of what exclusions are used in the hands of the primary repo by being able to reconfigure the github action itself at anytime.


casey-erdmann:
Sample passing build (on this commit) using new scanner!
Screen Shot 2019-12-26 at 12 52 42 PM

PR #102, opened 2019-12-30
lgtm config to fix java and maven build versions for jdk13

casey-erdmann:
### LGTM config to fix java and maven build versions for jdk13

LGTM config to fix java and maven build versions for jdk13. Our previous builds were failing due to what `appeared` to be a version mismatch on the LGTM service with JDK. It ended up being resolved by specifying the correct maven version that was working when building locally which at the time of this PR should be the latest `3.6.3`.

### Checklist

- [X] Tests and linting pass

### Security
This improves the security posture by fixing a breaking build bug that was revealed during our JDK13 upgrade preventing automated SAST. Now SAST will be able to resume as part of the regular SDLC.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -2,3 +2,8 @@ extraction:
   java:
     index:
       java_version: 13
+      maven:
+        version: 3.6.3
+      build_command:
+      - "java -version"
+      - "mvn -version"


casey-erdmann:
Screenshot from running the configuration against `master` as a test analysis. Java passed with no issues.
Screen Shot 2019-12-30 at 2 53 14 PM

casey-erdmann:
The LGTM Java build is only a partial fail on JUST `master`. This is expected since master is using the old config. This PR should fix this issue. Documenting the build success and noting it here since this is an admin override PR.
Screen Shot 2019-12-30 at 2 56 37 PM

PR #107, opened 2020-01-02
ab2d 686 delete files on job cancellation

spathiyil:
When a job is cancelled, delete the output directory created by the cancelled job and the partially created import files 

### Summary

When a job is cancelled, the worker would stop the import process in an orderly fashion. It was concluded that the (partial) import files created should also be cleaned up after stopping the processing. This PR refines the process to also delete the output directory created by the cancelled job along the files created within it.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
@@ -70,8 +70,9 @@ public class JobProcessorImpl implements JobProcessor {
         final Job job = jobRepository.findByJobUuid(jobUuid);
         log.info("Found job");
 
+        Path outputDirPath = null;
         try {
-            final Path outputDirPath = Paths.get(efsMount, jobUuid);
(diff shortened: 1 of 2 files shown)


worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -200,36 +205,25 @@ private void deleteExistingDirectory(Path outputDirPath) {
...
-        }
-
         while (!futureHandles.isEmpty()) {
             sleep();
             errorCount += processHandles(futureHandles);
         }
 
+        if (isCancelled) {
+            final String errMsg = "Job was cancelled while it was being processed";
+            log.warn("{} - JobUuid :[{}]", errMsg, jobUuid);
```
adaykin: jobUUid should already be present from MDC.put in JobHandler

spathiyil: done - removed.

PR #108, opened 2020-01-02
Feature/ab2d 669 fhir capability

adaykin:
Adding the capability API call from the FHIR spec

Add the capability statement found here http://hl7.org/implement/standards/fhir/capabilitystatement.html, mostly using the existing BCDA example for how to do it. 

### Checklist

- [x] Tests and linting pass

### Security

NA

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
@@ -351,27 +351,15 @@ public class BulkDataAccessAPI {
         }
     }
 
-    @ApiOperation(value = "Downloads a file produced by an export job.", response = String.class,
-            produces = "application/fhir+ndjson",
+    @ApiOperation(value = "A request for the FHIR capability statement", response = String.class,
+            produces = "application/json",
             authorizations = {
                     @Authorization(value = "Authorization", scopes = {
-                            @AuthorizationScope(description = "Downloads Export File", scope = "Authorization") })
+                            @AuthorizationScope(description = "Returns the FHIR capability statement", scope = "Authorization") })
             })
-    @ApiImplicitParams(value = {
-            @ApiImplicitParam(name = "Accept", required = false, paramType = "header", value =
-                    "application/fhir+json")}
-    )
     @ApiResponses(value = {
-            @ApiResponse(code = 200, message = "Returns the requested file as " +
-                    "application/fhir+ndjson.", responseHeaders = {
-                    @ResponseHeader(name = "Content-Type", description =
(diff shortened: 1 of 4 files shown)


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -349,4 +350,22 @@ private String getUrlPath(Job job, String filePath) {
...
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "Returns the FHIR capability statement", response =
+                    String.class)}
+    )
+    @ResponseStatus(value = HttpStatus.OK)
+    @GetMapping(value = "/metadata")
+    public ResponseEntity<String> capabilityStatement() {
+        CapabilityStatement capabilityStatement = new CapabilityStatement();
+        String json = new Gson().toJson(capabilityStatement);
+        return new ResponseEntity<>(json, null, HttpStatus.OK);
```
spathiyil: Would it make sense to return `ResponseEntity<CapabilityStatement>`  here instead of converting it into a string?
If we could, we could remove the reference to Gson() as well.

api/src/main/java/gov/cms/ab2d/api/controller/CapabilityStatement.java
```diff
@@ -0,0 +1,133 @@
...
+        private final String url = BASE_SANDBOX_URL;
+    }
+
+    private final class Rest {
+
+        private final String mode = "server";
+
+        private final Security security = new Security();
+
+        private final List<Interaction> interaction = new ArrayList<>();
```
spathiyil: Should this need be plural? `interactions`

adaykin: It's kind of weird, but it has to match up with the FHIR standard they use singular names http://hl7.org/implement/standards/fhir/capabilitystatement.html
https://bcda.cms.gov/sandbox/technical-user-guide/#examples

spathiyil: Thanks for clarifying...  did not realize that earlier.

spathiyil: Thanks for clarifying... did not realize that earlier.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
adaykin: I'm not sure if it would serialize to JSON properly (or in a predictable way) if it's returning the class? I did it this way since it's easy to test and verify with the Gson conversion in the tests.

PR #110, opened 2020-01-03
Feature/ab2d 673 automated test v2

adaykin:
e2e tests for the whole application

e2e tests that make several API requests, and then allow the worker to process the data. After the worker has processed the data verify that the data is correct. Adds a new docker compose so that the application can be built with a different environment and allow e2e specific variables to take place and share a volume between worker and API.

### Checklist

- [x] Tests and linting pass

### Security

dev account data that is intentionally exposed


e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,434 @@
...
+            put("password", password);
+            put("scope", "openid");
+        }};
+
+        HttpRequest jwtRequest = HttpRequest.newBuilder()
+                .uri(URI.create(oktaUrl))
+                .timeout(Duration.ofSeconds(DEFAULT_TIMEOUT))
+                .header("Content-Type", "application/x-www-form-urlencoded")
+                .header("Accept", "application/json")
+                .header("Authorization", "Basic MG9hMXB4YXB6ZE9XaUhmOXUzNTc6WGIxQnpDb1ktS0Y4Z1FhTF9JMmVzVmxsdkpZZ2VzX2d1QTVHZTlpaQ==")
```
lailasharshar: It may be more sustainable to pass in the clientId and clientPassword in the yaml and then construct the Basic Auth value "Basic Base64(clientId:clientPassword). I have those values if you need them (I have decoded them)

docker-e2e-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+  db:
+    image: postgres:11
+    environment:
+      - POSTGRES_DB=ab2d
+      - POSTGRES_USER=ab2d
+      - POSTGRES_PASSWORD=ab2d
+    ports:
+      - "5432:5432"
+  build:
+    image: maven:3-jdk-12
```
spathiyil: shouldn't this be JDK 13?

docker-compose.yml
```diff
@@ -40,5 +41,6 @@ services:
       - AB2D_DB_USER=ab2d
       - AB2D_DB_PASSWORD=ab2d
       - AB2D_EFS_MOUNT=/tmp/ab2d_efs_mount
+      - SPRING_PROFILE=prod
```
spathiyil: It is a bit odd to default the value of spring profile to `PROD`. 
I wonder if we should default it to some other environment? like `dev` or `qa` or something which is _not prod_ ?

docker-e2e-compose.yml
adaykin: Yeah it should be 13, I copied it before I merged the latest master which now has 13.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -274,8 +274,8 @@ private void logSuccessfulJobCreation(Job job) {
                 resp.setTransactionTime(jobCompletedAt.toHumanDisplay());
                 resp.setRequest(job.getRequestUrl());
                 resp.setRequiresAccessToken(true);
-                resp.setOutput(job.getJobOutputs().stream().filter(o -> !o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), getUrlPath(job, o.getFilePath()))).collect(Collectors.toList()));
-                resp.setError(job.getJobOutputs().stream().filter(o -> o.isError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), getUrlPath(job, o.getFilePath()))).collect(Collectors.toList()));
+                resp.setOutput(job.getJobOutputs().stream().filter(o -> !o .getError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), getUrlPath(job, o.getFilePath()))).collect(Collectors.toList()));
+                resp.setError(job.getJobOutputs().stream().filter(o -> o.getError()).map(o -> new JobCompletedResponse.Output(o.getFhirResourceType(), getUrlPath(job, o.getFilePath()))).collect(Collectors.toList()));
```
spathiyil: These long lines may benefit from a refactor - perhaps extract a method or two?

docker-compose.yml
adaykin: Hmm I'm not sure if we have those environments up at the moment, currently there's still not any true prod environment so I think this can be revisited once some more decisions come. In the future we might have a docker-compose-qa.yml or some other way of changing environments.

common/src/main/resources/db/changelog/v001/schema.sql
```diff
@@ -40,14 +40,14 @@ ALTER TABLE sponsor ADD CONSTRAINT "fk_sponsor_to_sponsor_parent" FOREIGN KEY (p
 --changeset spathiyil:AB2D-291-CreateTable-contract failOnError:true 
 CREATE TABLE contract
 (
-    id                  BIGINT              NOT NULL,
+    id                  BIGSERIAL              NOT NULL,
```
dkrylovsb: Unsure of the purpose of this change. We should keep the primary keys consistent across, i.e. use `BIGINT` coupled with a sequence for all.

common/src/main/resources/db/changelog/v001/seed_data.sql
```diff
@@ -15,6 +15,6 @@ INSERT INTO user_account (id, username, first_name, last_name, email, sponsor_id
 INSERT INTO user_role (user_account_id, role_id) VALUES ((SELECT id FROM user_account WHERE username='EileenCFrierson@example.com'),
                                                          (SELECT id FROM role WHERE name='SPONSOR'));
 
-INSERT INTO contract (id, contract_number, contract_name, sponsor_id, attested_on) VALUES ((select nextval('hibernate_sequence')),'S0001', 'S0001',
+INSERT INTO contract (contract_number, contract_name, sponsor_id, attested_on) VALUES ('S0001', 'S0001',
```
dkrylovsb: I am guessing you are looking for predictable primary key values, but should never have a need to know those, really.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
dkrylovsb: 👍

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,543 @@
...
+import java.sql.Statement;
+import java.time.Duration;
+import java.time.OffsetDateTime;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+
+@Testcontainers
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class TestRunner {
```
dkrylovsb: The challenge is that this e2e test will also need to act as a "smoke test" against the existing environments, i.e. Dev and Sbx so that we don't have to do manual QA each time the environments redeploy. In those scenarios the test will *not* have access to Docker or to the database. It will act purely as a HTTP client.

Therefore, we need to neatly separate the container setup code (which will only run in Travis CI), and the actual test code.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,543 @@
...
+
+    private Integer sponsorId = null;
+
+
+    @BeforeAll
+    public void init() throws IOException, InterruptedException, JSONException {
+        container.start();
+
+        Yaml yaml = new Yaml();
+        InputStream inputStream = getClass().getResourceAsStream("/default-config.yml");
```
dkrylovsb: We'll need a way to specify different yaml files for different environments.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,543 @@
...
+        HttpResponse<String> jwtResponse = httpClient.send(jwtRequest, HttpResponse.BodyHandlers.ofString());
+        String responseJwtString = jwtResponse.body();
+        JSONObject responseJsonObject = new JSONObject(responseJwtString);
+        jwtStr = responseJsonObject.getString("access_token");
+
+        dataSource = new DriverManagerDataSource();
+        dataSource.setDriverClassName("org.postgresql.Driver");
+        dataSource.setUrl("jdbc:postgresql://localhost:5432/ab2d");
+        dataSource.setUsername("ab2d");
+        dataSource.setPassword("ab2d");
```
dkrylovsb: Based on the earlier comment, the e2e test will not have access to the database when run against the environments. We could still manipulate the DB for the Travis CI of course, but unsure if we really need to.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,543 @@
...
+
+        uploadOrgStructureReport();
+        uploadAttestationReport();
```
dkrylovsb: This is going to be complicated because the test user (EileenCFrierson@example.com) does not have the admin rights.

I think initially the e2e test might need to assume the only data in the database is whatever is populated by the `seed.sql`.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,543 @@
...
+
+    @BeforeEach
+    public void initEach() throws SQLException {
+        try (Connection connection = dataSource.getConnection(); Statement statement = connection.createStatement()) {
+            statement.execute("DELETE FROM job_output");
+        }
+        try (Connection connection = dataSource.getConnection(); Statement statement = connection.createStatement()) {
+            statement.execute("DELETE FROM job");
+        }
+    }
```
dkrylovsb: No need for these. The test should not be affected by any preexisting job & outputs, because it is going to create new jobs anyway.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,571 @@
...
+                .uri(URI.create(AB2D_ADMIN_URL + "uploadAttestationReport"))
+                .timeout(Duration.ofSeconds(DEFAULT_TIMEOUT))
+                .header("Content-Type", "application/json")
+                .header("Authorization", "Bearer " + jwtStr)
+                .POST(HttpRequest.BodyPublishers.ofInputStream(() -> {
+                    return getClass().getResourceAsStream("/Attestation_Report_Sample.xlsx");
+                }))
+                .build();
+
+        return httpClient.send(uploadRequest, HttpResponse.BodyHandlers.ofString());
```
dkrylovsb: See previous comments. Might need to comment these out for the time being

common/src/main/resources/db/changelog/v001/schema.sql
adaykin: So in https://github.com/CMSgov/ab2d/pull/110/files#diff-d72ce5f81287c43e6ea1c18ac61182c2 after removing the contract ID sequence and using serial for the ID in contract table, I’m not seeing the issue that @Satheesh Pathiyil saw earlier with the missing_sequence mentioned here https://github.com/CMSgov/ab2d/pull/31#discussion_r341251018 and my e2e tests run through the app quite a bit. I’m guessing something changed in the application/db so that it’s no longer an issue. 

I was having issues with setting up contracts manually with SQL inserts, the stress testing is some of the new stuff I'm currently working on while we're looking for new tasks.

dkrylovsb: Stress testing is a whole different beast; it will likely need to go into a separate module.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,571 @@
...
+        String uuid = null;
+        try (Connection connection = dataSource.getConnection(); Statement statement = connection.createStatement()) {
+            ResultSet resultSet = statement.executeQuery("SELECT job_uuid FROM job WHERE job_status != 'CANCELLED'");
+            while (resultSet.next()) {
+                uuid = resultSet.getString("job_uuid");
+            }
+        }
+
+        return uuid;
+    }
```
dkrylovsb: Should not need this. Job UUID is returned from the `export` call.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,571 @@
...
+
+    private Set<String> getAllJobUuids() throws SQLException {
```
dkrylovsb: I think any code that touches the DB directly might need to be commented out and refactored to be using HTTP interactions.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,571 @@
...
+        List<String> xProgressList = statusResponse.headers().map().get("x-progress");
+        Assert.assertEquals(xProgressList.iterator().next(), "0% complete");
+
+        HttpResponse<String> retryStatusResponse = statusRequest(contentLocationList.iterator().next());
+
+        Assert.assertEquals(429, retryStatusResponse.statusCode());
+        List<String> retryAfterListRepeat = retryStatusResponse.headers().map().get("retry-after");
+        Assert.assertEquals(retryAfterListRepeat.iterator().next(), "5");
+
+        String jobUUid = getJobUuid();
```
dkrylovsb: Job UUID can be obtained from the status check URL.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,571 @@
...
+
+        HttpResponse<String> deleteResponse = cancelJobRequest(jobUUid);
+        Assert.assertEquals(202, deleteResponse.statusCode());
```
dkrylovsb: I think testing the "Delete" operation should be a separate test method on its own.

dkrylovsb: That way you don't need the "second try" below.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,571 @@
...
+
+        HttpResponse<String> exportResponseSecondTry = exportRequest();
+
+        Assert.assertEquals(202, exportResponseSecondTry.statusCode());
+        List<String> contentLocationListSecondTry = exportResponseSecondTry.headers().map().get("content-location");
+
+        Thread.sleep(THREAD_DELAY);
```
dkrylovsb: We cannot sleep for a fixed number of seconds and then expect the job to be completed since we don't know for sure for how long the job will actually run, especially when testing the external environments.

Instead, we need a _loop_ that will be checking the status periodically sleeping for the `retry-after` number of seconds in between (just to avoid 429s). The loop should exit once the status turns to done (200). The loop should fail the test if the job has not completed within a given timeframe (i.e. a timeout). For now, say, 5 minutes.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,571 @@
...
+        Assert.assertEquals(JobStatus.SUCCESSFUL.name(), jobData.get("status"));
+        Assert.assertEquals("100%", jobData.get("status_message"));
+        Assert.assertEquals(100, jobData.get("progress"));
+        Assert.assertEquals(null, jobData.get("resource_types"));
+        Assert.assertEquals(null, jobData.get("fhir_resource_type"));
+        Assert.assertEquals(0, jobData.get("contract_id")); // In JDBC SQL, null comes back as 0
+    }
+
+    @Test
+    public void runContractNumberExport() throws IOException, InterruptedException, SQLException, JSONException {
```
dkrylovsb: A lot of code in this test method repeats code from the previous one; let's refactor to extract the common code into a reusable method.

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerServiceImpl.java
```diff
@@ -17,16 +18,26 @@
...
+        // so that the job doesn't get processed and finishes. This way incomplete statuses can be tested. This should never
+        // run in any other environment
+        if (delayProcessing.equals("true")) {
+            try {
+                Thread.sleep(10000);
+            } catch (InterruptedException e) {
+                log.error("Thread was interrupted while sleeping for e2e tests", e);
+            }
+        }
+
```
dkrylovsb: We should address this by increasing the number of patients per contract, so that job processing takes longer.

worker/src/main/resources/application-e2e.properties
```diff
@@ -0,0 +1 @@
+delay-processing=true
```
dkrylovsb: This entire file could be removed altogether.

api/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM openjdk:13
 WORKDIR /usr/src/ab2d-api
 ADD target /usr/src/ab2d-api
-CMD java -jar api-*-SNAPSHOT.jar
+CMD java -Dspring.profiles.active=${SPRING_PROFILE} -jar api-*-SNAPSHOT.jar
```
dkrylovsb: Based on the comments I have left so far, we might not even need the `e2d` application.properties file, which essentially eliminates the need for profiles.

docker-e2e-compose.yml
```diff
@@ -0,0 +1,52 @@
+version: '3'
```
dkrylovsb: We should not introduce a new Docker Compose YAML file. Instead, we should add the missing piece -- the shared volume -- to the existing one. The existing one is _broken_ anyway because of that missing piece.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,512 @@
...
+        this.environment = environment;
+        init();
+    }
+
+    public void init() throws IOException, InterruptedException, JSONException {
+        if(environment.isUsesDockerCompose()) {
+            DockerComposeContainer container = new DockerComposeContainer(
+                    new File("../docker-e2e-compose.yml"))
+                    //.withScaledService("api", 2) // failing now since it's not changing ports
+                    //.withScaledService("worker", 2)
```
dkrylovsb: Can we still scale the worker though, since it is not using a port? If yes, let's definitely do so.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,512 @@
...
+                .version(HttpClient.Version.HTTP_1_1)
+                .followRedirects(HttpClient.Redirect.NORMAL)
+                .build();
+
+        var jwtRequestParms = new HashMap<>() {{
+            put("grant_type", "password");
+            put("username", TEST_USER);
+            put("password", password);
+            put("scope", "openid");
+        }};
```
dkrylovsb: I think you can transition the Okta piece to using client id/secret only, i.e. no human user accounts anymore. Eileen C Frierson will have to go. You can use one of the "PDP-nnnn" accounts. Ping me for help with setting those up.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,512 @@
...
+        Assert.assertEquals(429, secondExportResponse.statusCode());
+
+        HttpResponse<String> statusResponse = statusRequest(contentLocationList.iterator().next());
+
+        Assert.assertEquals(202, statusResponse.statusCode());
+        List<String> retryAfterList = statusResponse.headers().map().get("retry-after");
+        Assert.assertEquals(retryAfterList.iterator().next(), String.valueOf(DELAY));
+        List<String> xProgressList = statusResponse.headers().map().get("x-progress");
+        Assert.assertEquals(xProgressList.iterator().next(), "0% complete");
+
```
dkrylovsb: There is no way we can expect a certain completion percentage value. Let's just verify the header matches a regexp.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,512 @@
...
+        HttpResponse<String> retryStatusResponse = statusRequest(contentLocationList.iterator().next());
+
+        Assert.assertEquals(429, retryStatusResponse.statusCode());
+        List<String> retryAfterListRepeat = retryStatusResponse.headers().map().get("retry-after");
+        Assert.assertEquals(retryAfterListRepeat.iterator().next(), String.valueOf(DELAY));
+
+        HttpResponse<String> statusResponseAgain = pollForStatusResponse(contentLocationList.iterator().next());
```
dkrylovsb: A lot of repetition with the previous method. Can we extract a common private method and re-use?

e2e-test/src/test/resources/prod-config.yml
```diff
@@ -0,0 +1,8 @@
+# Placeholders for now
```
dkrylovsb: We need a way for these YAML files to refer to environment variables in certain cases, such as in Prod.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
adaykin: Yes the worker can be changed to 2

docker-e2e-compose.yml
dkrylovsb: @adaykin Let's delete this file.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -0,0 +1,491 @@
...
+        AB2D_API_URL = yamlMap.get("ab2d-api-url");
+        AB2D_ADMIN_URL = yamlMap.get("ab2d-admin-url");
+
+        var jwtRequestParms = new HashMap<>() {{
+            put("grant_type", "client_credentials");
+            put("scope", "clientCreds");
+        }};
+
+        String oktaClientId = System.getenv(environment.name() + "_OKTA_CLIENT_ID");
+        String oktaPassword = System.getenv(environment.name() + "_OKTA_CLIENT_PASSWORD");
```
dkrylovsb: The names of the environment variables should _not_ change across tiers -- it is the values that change. Making variable _names_ tier-dependent will make Jenkins setup more difficult.

e2e-test/src/test/resources/imp-config.yml
```diff
@@ -0,0 +1,3 @@
+okta-url: 'https://dev-418212.okta.com/oauth2/default/v1/token'
```
dkrylovsb: Use Okta Test here too.

PR #117, opened 2020-01-08
ab2d 667 shutdown cleanup reset jobs

spathiyil:
When a shutdown command is issued, worker needs to put active jobs into submitted status

### Summary

When a shutdown command is issued, worker needs to put active jobs into submitted status so that the job can be processed by another instance.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/service/WorkerServiceImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/service/WorkerServiceImpl.java
@@ -23,21 +23,23 @@ public class WorkerServiceImpl implements WorkerService {
     private final JobProcessor jobProcessor;
     private final ShutDownService shutDownService;
 
-    private static List<String> activeJobs = Collections.synchronizedList(new ArrayList<>());
+    private List<String> activeJobs = Collections.synchronizedList(new ArrayList<>());
 
 
     @Override
     public void process(String jobUuid) {
 
         activeJobs.add(jobUuid);
+        try {
+            jobPreprocessor.preprocess(jobUuid);
+            log.info("Job was put in progress");
 
-        jobPreprocessor.preprocess(jobUuid);
-        log.info("Job was put in progress");
+            jobProcessor.process(jobUuid);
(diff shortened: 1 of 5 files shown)


worker/src/main/java/gov/cms/ab2d/worker/service/WorkerServiceImpl.java
```diff
@@ -16,17 +21,37 @@
 
     private final JobPreProcessor jobPreprocessor;
     private final JobProcessor jobProcessor;
+    private final ShutDownService shutDownService;
+
+    private static List<String> activeJobs = Collections.synchronizedList(new ArrayList<>());
```
dkrylovsb: Does it need to be `static`? I _think_ the Worker bean is a singleton, but even if it is not, each instance would maintain the jobs it is processing anyway.

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerServiceImpl.java
```diff
@@ -16,17 +21,37 @@
...
+
+        activeJobs.add(jobUuid);
+
         jobPreprocessor.preprocess(jobUuid);
         log.info("Job was put in progress");
 
         jobProcessor.process(jobUuid);
         log.info("Job was processed");
+
+        activeJobs.remove(jobUuid);
```
dkrylovsb: We need a `try/finally` block here and remove the `jobUuid` from the list in the `finally` clause.

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerServiceImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/WorkerServiceImpl.java
spathiyil: done

PR #120, opened 2020-01-13
Feature/ab2d 450 create automation for cms sandbox

lhanekam:
Create automation for sandbox

### Summary

- Create automation for sandbox
- Refactor dev and sandbox so that dev acts as the keeper of the API and worker ECR repos (note that ECR repos will be migrated to management environment when the VPC is available)
- Destroy and deploy scripts for testing
- Documentation updates

### Checklist

- [x] Tests and linting pass

### Security

_N/A_


lhanekam:
@dkrylovsb: EFS and ECS issues have been fixed. I have deployed the latest code to both dev and sandbox and verified that the file download API works for both environments.

PR #121, opened 2020-01-14
Removed JsonNode

lailasharshar:
### Summary

When clicking on the status endpoint, it would give an error about JsonNode. 

### Checklist

- [x] Tests and linting pass

### Security
No impact

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/config/SwaggerConfig.java
+++ b/api/src/main/java/gov/cms/ab2d/api/config/SwaggerConfig.java
@@ -2,7 +2,6 @@ package gov.cms.ab2d.api.config;
 
 import com.fasterxml.classmate.TypeResolver;
 import com.fasterxml.jackson.annotation.*;
-import com.fasterxml.jackson.databind.JsonNode;
 import gov.cms.ab2d.api.util.Constants;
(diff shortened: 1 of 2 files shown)


api/src/main/java/gov/cms/ab2d/api/config/SwaggerConfig.java
```diff
@@ -42,7 +41,6 @@ public Docket api() {
                 .paths(PathSelectors.ant(API_PREFIX + FHIR_PREFIX + "/**"))
                 .build()
                 .directModelSubstitute(Resource.class, String.class)
-                .directModelSubstitute(JsonNode.class, Void.class)
```
dkrylovsb: I originally did this to prevent `JsonNode` from appearing within the "Model" section of the Swagger doc. Let's make sure that behavior has not re-occurred with this change.

lailasharshar: I did not see it (and I looked for it). The problem is that it was looking for it and that's why it threw an error. I changed the return type. While I'm not technically returning a "JobCompletedResponse", if it's not completed, it's just populated with null which seems like a reasonable response.

PR #123, opened 2020-01-15
AB2D 779 - Modify "GetPatientsByContract" stub to return variable number of patients

spathiyil:
Modify "Getting Patients by Contract" mock to return variable number of patients based on contract number

### Summary

Modify "Getting Patients by Contract" mock to return variable number of patients based on contract number

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterStub.java
```diff
@@ -19,65 +18,117 @@
...
+            allRows.addAll(createMoreRows(numberOfRows, rowsToRetrieve, patientIdRows));
+        }
+
+        return allRows;
+    }
+
+    private int determineNumberOfRows(int contractSno) {
+        if (contractSno == 0) {
+//            return 100;   // temporary change till BFD calls mocked
+            return 10;
```
spathiyil: Temporary change for now, till the BFD calls are mocked - so that only 10 patient records are returned for `S0000`

worker/src/test/java/gov/cms/ab2d/worker/processor/JobProcessorIntegrationTest.java
```diff
@@ -73,14 +73,15 @@ void processJob() {
...
+        var processedJob = cut.process("S00000");
 
         assertThat(processedJob.getStatus(), is(JobStatus.SUCCESSFUL));
         assertThat(processedJob.getStatusMessage(), is("100%"));
         assertThat(processedJob.getExpiresAt(), notNullValue());
         assertThat(processedJob.getCompletedAt(), notNullValue());
     }
 
+    @Disabled("disable till the BFD calls are mocked out.")
     @Test
```
spathiyil: This test is disabled till the BDF calls are mocked.

worker/src/test/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterStubTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+
+
+    @Test
+    @Disabled("Takes too long to run. But could be run locally as needed")
+    @DisplayName("when contractNumber is greater than 9999, returns 9999000 patient records")
+    void when_S9999_returns_9999000() {
+        var patients = cut.getPatients("S9999").getPatients();
+        assertThat(patients.size(), is(9999000));
+    }
```
dkrylovsb: Let's delete. Probably no point in having a disabled test.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobPreProcessorIntegrationTest.java
```diff
@@ -61,7 +61,7 @@ void setUp() {
     @Test
     @DisplayName("When a job is in submitted status, it can be put into progress upon starting processing")
     void whenJobIsInSubmittedStatus_ThenJobShouldBePutInProgress() {
-        var processedJob = cut.preprocess("S001");
+        var processedJob = cut.preprocess("S00000");
```
dkrylovsb: An extra `0` here and onward.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobProcessorIntegrationTest.java
dkrylovsb: How long is this one taking?

dkrylovsb: Also, why is it taking long if we have S0000 set to 10 patients only?

spathiyil: I do not have the time, but it ran long enough and filled the log and failed the build in CI.
https://travis-ci.org/CMSgov/ab2d/builds/637433959?utm_source=github_status&utm_medium=notification

spathiyil: You are right, this should only return 10 patients, but I just realized that I had a randon number appended to the contractNumber in this test case - due to which it was getting a bunch of patient records. Will have a commit shortly to fix it.

worker/src/test/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterStubTest.java
spathiyil: removed this disabled test

worker/src/test/java/gov/cms/ab2d/worker/processor/JobPreProcessorIntegrationTest.java
spathiyil: changed it to have 4 zeroes instead of 5 zeroes

worker/src/test/java/gov/cms/ab2d/worker/processor/JobProcessorIntegrationTest.java
spathiyil: Once I removed the random number suffix to the contractNumber in this test, it started finishing fast as it was 10 patient records only. So I went back and enabled this test.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobProcessorIntegrationTest.java
```diff
@@ -135,10 +133,9 @@ private User createUser(Sponsor sponsor) {
     }
 
     private Contract createContract(Sponsor sponsor) {
-        final int anInt = random.nextInt(299);
         Contract contract = new Contract();
-        contract.setContractName("CONTRACT_0000" + anInt);
-        contract.setContractNumber("CONTRACT_0000" + anInt);
```
spathiyil: This line in the test case was causing the contract number to not end in `0000` as a random number was being appended to the end of the contract number. I've removed this and now, it processes 10 patient records as expected

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/PatientClaimsProcessorImpl.java
```diff
@@ -115,7 +115,7 @@ private void appendToFile(Path outputFile, ByteArrayOutputStream byteArrayOutput
         tryLock(lock);
 
         try {
-            log.info("Attempting to append to file", keyValue(FILE_LOG, outputFile.getFileName()));
+            log.info("Attempting to append to file", keyValue(FILE_LOG, outputFile.toFile().getName()));
             fileService.appendToFile(outputFile, byteArrayOutputStream);
```
spathiyil: We found this logging statement was logging the path in addition to the filename while we were prepping for the demo during PI planning/roadmap session. I saw this while testing this change and fixed it.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobProcessorIntegrationTest.java
dkrylovsb: The CI log is flooded with "Attempting to append to file" and "finished writing" messages. Intensive logging was _likely_ the reason for slowness. If we drop these two log entries as too verbose, then 100 benes would not be a problem as well.

spathiyil: I could change the logging level to DEBUG... especially the first one `Attempting to append to file` as it is happening for each `appendToFile`.
The 2nd line you mentioned (`finished writing`) only happens once per patient record. But I could lower that to DEBUG as well.

dkrylovsb: `DEBUG` is fine

spathiyil: done - reduced logging level to DEBUG and now it returns 100 patient records for contract `0000` for this test.

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterStub.java
spathiyil: reverted, so that it now returns 100 patient records for contractNo = 0000

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterStub.java
```diff
@@ -19,65 +18,116 @@
...
+    private List<String> createMoreRows(int numberOfRows, int rowsToRetrieve, List<String> patientIdRows) {
+        final int diff = numberOfRows - rowsToRetrieve;
+        if (diff < MAX_ROWS) {
+            return patientIdRows.stream().limit(diff).collect(Collectors.toList());
+        } else {
+            List<String> accumulator = new ArrayList<>();
+            accumulator.addAll(patientIdRows);
+
+            final List<String> moreRows = createMoreRows(diff, rowsToRetrieve, patientIdRows);
+            accumulator.addAll(moreRows);
```
dkrylovsb: Just realized this is a recursion. We are risking `StackOverflowError` in case of large contract numbers. A simple loop would probably do here.

spathiyil: done - replaced recursion with a simple loop.

PR #124, opened 2020-01-16
Feature/ab2d 616 refactor worker tests

adaykin:
Change the tests so that they don't call the real BFD Client.

Modify the setup of the tests so that they mock up the dependencies correctly and don't call the BFD API. Adding a parent sponsor in so contracts get processed properly. Disable the integration from running since processing is getting called manually.

### Checklist

- [x] Tests and linting pass

### Security

NA

### DIFF
--- a/worker/src/test/java/gov/cms/ab2d/worker/processor/JobProcessorIntegrationTest.java
+++ b/worker/src/test/java/gov/cms/ab2d/worker/processor/JobProcessorIntegrationTest.java
@@ -1,26 +1,34 @@
 package gov.cms.ab2d.worker.processor;
 
+import ca.uhn.fhir.context.FhirContext;
+import gov.cms.ab2d.bfd.client.BFDClient;
 import gov.cms.ab2d.common.model.Contract;
 import gov.cms.ab2d.common.model.Job;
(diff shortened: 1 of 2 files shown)


worker/src/test/java/gov/cms/ab2d/worker/processor/JobProcessorIntegrationTest.java
```diff
@@ -1,42 +1,55 @@
...
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertThat;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.when;
 
 @SpringBootTest
 @Testcontainers
+@SpringIntegrationTest(noAutoStartup = {"inboundChannelAdapter", "*Source*"})
+@Transactional
```
spathiyil: I do not think this `@Transactional` annotation should be here.

adaykin: It's needed since there's a lazy loading hibernate error when it's not present.

PR #125, opened 2020-01-16
increase size of username column in user_account table

spathiyil:
increase size of username column in user_account table

### Summary

increase size of username column in user_account table

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIUserTests.java
+++ b/api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIUserTests.java
@@ -138,6 +138,8 @@ public class AdminAPIUserTests {
                         .contentType(MediaType.APPLICATION_JSON).content(mapper.writeValueAsString(userDTO))
                         .header("Authorization", "Bearer " + token));
 
+        userDTO.setEmail("anotherEmail@test.com");
+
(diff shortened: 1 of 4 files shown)


api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIUserTests.java
```diff
@@ -138,6 +138,8 @@ public void testCreateDuplicateUser() throws Exception {
                         .contentType(MediaType.APPLICATION_JSON).content(mapper.writeValueAsString(userDTO))
                         .header("Authorization", "Bearer " + token));
 
+        userDTO.setEmail("anotherEmail@test.com");
+
```
dkrylovsb: Curious how this is related to the `username` change we are making?

spathiyil: I agree. It was unexpected. 
But after reviewing the unit tests, it was expected to give a duplicate error on username and the next test the validation error on email, but before change, somehow it seems to have always got the validation error on username first. 
Now after the column size was altered, there is no such guarantee and it was giving the email as the validation failure.

PR #126, opened 2020-01-16
Feature/ab2d 725 sbx datasets

adaykin:
Adding in sandbox data

Creating contracts, user accounts, and sponsors so that the sandbox data can be accessed. This data will be tied to various amounts of data, the higher the contract number, the more data.

### Checklist

- [x] Tests and linting pass

### Security

Roles, User Accounts, Sponsors tied to data.

### DIFF
--- a/common/src/main/resources/db/changelog/v001/add_sbx_accounts.sql
+++ b/common/src/main/resources/db/changelog/v001/add_sbx_accounts.sql
@@ -1,58 +1,52 @@
 INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 2, 'PDP-100', null, null);
 INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 3, 'PDP-1000', null, null);
 INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 4, 'PDP-2000', null, null);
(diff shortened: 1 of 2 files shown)


common/src/main/resources/db/changelog/v001/add_sbx_accounts.sql
```diff
@@ -0,0 +1,52 @@
+INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 2, 'PDP-100', null, null);
+INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 3, 'PDP-1000', null, null);
+INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 4, 'PDP-2000', null, null);
+INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 5, 'PDP-5000', null, null);
+INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 6, 'PDP-10000', null, null);
+INSERT INTO sponsor (id, hpms_id, org_name, legal_name, parent_id) VALUES ((select nextval('hibernate_sequence')), 7, 'PDP-30000', null, null);
```
dkrylovsb: These are all "parent" organizations (due to parent_id being null), and in our design parent orgs cannot have direct links to contracts. Let's create a single parent org called `PDP-SBX` and make all these children on that. See how Eileen F. was set up.

PR #127, opened 2020-01-17
AB2D 832 - Handle efs.mount path without trailing slash

spathiyil:
Fix code so that it is not sensitive to presence/absence of trailing slash in the efs.mount property

### Summary

Fix code so that it is not sensitive to presence/absence of trailing slash in the efs.mount property

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/audit/src/test/java/gov/cms/ab2d/audit/cleanup/FileDeletionServiceTest.java
+++ b/audit/src/test/java/gov/cms/ab2d/audit/cleanup/FileDeletionServiceTest.java
@@ -2,13 +2,11 @@ package gov.cms.ab2d.audit.cleanup;
 
 import gov.cms.ab2d.audit.SpringBootApp;
 import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
-import org.junit.jupiter.api.MethodOrderer;
-import org.junit.jupiter.api.Order;
-import org.junit.jupiter.api.TestMethodOrder;
+import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.context.junit4.SpringRunner;
@@ -19,7 +17,10 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.nio.file.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.BasicFileAttributeView;
 import java.nio.file.attribute.FileTime;
 import java.time.LocalDate;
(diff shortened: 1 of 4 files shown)


common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -130,7 +130,7 @@ public Resource getResourceForJob(String jobUuid, String fileName) throws Malfor
             throw new ResourceNotFoundException("No Job Output with the file name " + fileName + " exists in our records");
         }
 
-        Path file = Paths.get(fileDownloadPath + job.getJobUuid() + File.separator +  fileName);
+        Path file = Paths.get(fileDownloadPath, job.getJobUuid(), fileName);
```
dkrylovsb: Was this the only place in the non-test code where we had this issue?

audit/src/test/java/gov/cms/ab2d/audit/cleanup/FileDeletionServiceTest.java
```diff
@@ -63,62 +60,66 @@ private void changeFileCreationDate(Path path) throws IOException {
...
+
         // other tests set this value, so set it to the correct one, JUnit ordering annotations don't seem to be respected
         ReflectionTestUtils.setField(fileDeletionService, "efsMount", efsMount);
 
         // Don't change the creation date on this file, but do so on the next ones
         Path destinationNotDeleted = Paths.get(efsMount, TEST_FILE_NOT_DELETED);
         URL urlNotDeletedFile = this.getClass().getResource("/" + TEST_FILE_NOT_DELETED);
         Path sourceNotDeleted = Paths.get(urlNotDeletedFile.toURI());
-        Files.move(sourceNotDeleted, destinationNotDeleted, StandardCopyOption.REPLACE_EXISTING);
+        Files.copy(sourceNotDeleted, destinationNotDeleted, StandardCopyOption.REPLACE_EXISTING);
```
dkrylovsb: Curious what caused the change from `move` to `copy`? CC @adaykin .

common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
spathiyil: That was the only one I could find. 
I did `search in path`  in the codebase for the string "efs.mount" and updated the ones that showed up

audit/src/test/java/gov/cms/ab2d/audit/cleanup/FileDeletionServiceTest.java
spathiyil: One thing I found strange was that when I was executing this test file from IntelliJ, it was deleting the file from `src/test/resources` so the next run it will fail saying `file not found`.
Had a chat with @adaykin  yesterday and his suggestion was to try copy instead of move and see if it works. I was able to get it running, so went with that solution.

adaykin: I believe there were issues running the test in the IDE where the file was being removed with the move.

dkrylovsb: 👍

PR #128, opened 2020-01-17
Feature/ab2d 841 jwt env vars v2

adaykin:
Adding in environment variables for the JWTs that will be used for auth

Adding in an issuer, audience, and timeout variables for JWTs so that they can be configured for different environments

### Checklist

- [x] Tests and linting pass

### Security

Changing env for auth

### DIFF
--- a/api/src/main/resources/application.properties
+++ b/api/src/main/resources/application.properties
@@ -11,6 +11,7 @@ spring.integration.jdbc.initialize-schema=always
 api.retry-after.delay=30
 
 api.okta-jwt-issuer=${AB2D_OKTA_JWT_ISSUER}
+api.okta-jwt-audience=${AB2D_OKTA_JWT_AUDIENCE}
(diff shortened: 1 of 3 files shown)


api/src/main/resources/application.properties
```diff
@@ -10,9 +10,10 @@ spring.integration.jdbc.initialize-schema=always
 
 api.retry-after.delay=30
 
-api.okta-url=https://dev-418212.okta.com/oauth2/default
-api.okta-connection-timeout=5
-api.okta-read-timeout=5
+api.okta-jwt-issuer=${AB2D_OKTA_JWT_ISSUER:#{'https://test.idp.idm.cms.gov/oauth2/aus2r7y3gdaFMKBol297'}}
+api.okta-jwt-audience=${AB2D_OKTA_JWT_AUDIENCE:#{'AB2D'}}
```
dkrylovsb: Does `AB2D` work with Okta Test?

adaykin: Yes

PR #130, opened 2020-01-21
Feature/ab2d 839 update job progress during batch processing

spathiyil:
update job progress during batch processing

### Summary

update job progress during batch processing

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


common/src/main/java/gov/cms/ab2d/common/repository/JobRepository.java
```diff
@@ -38,4 +40,9 @@
     @Query("UPDATE Job j SET j.status = 'SUBMITTED' WHERE j.jobUuid IN :jobUuids ")
     void resetJobsToSubmittedStatus(List<String> jobUuids);
 
+
+    @Modifying
```
dkrylovsb: Does the method above -- `resetJobsToSubmittedStatus` need to be `Modifying` too then?

common/src/main/java/gov/cms/ab2d/common/repository/JobRepository.java
```diff
@@ -38,4 +40,9 @@
...
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    @Query(nativeQuery = true, value = "UPDATE job SET progress = :percentageCompleted WHERE job_uuid = :jobUuid ")
```
dkrylovsb: Curious why we're having to use a native, non HQL, query here. Seems to be a simple query.

common/src/main/java/gov/cms/ab2d/common/repository/JobRepository.java
spathiyil: Yes. It is also annotated with `@Modifying`

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -306,10 +365,27 @@ private int processHandles(List<Future<Integer>> futureHandles) {
...
+        var jobUuid = progressTracker.getJobUuid();
+        var processedCount = progressTracker.getProcessedCount();
+        var totalCount = progressTracker.getTotalCount();
+
+        final int percentageCompleted = (processedCount * 100) / totalCount;
+
+        if (processedCount - progressTracker.getLastUpdatedCount() >= reportProgressFrequency) {
+            progressTracker.setLastUpdatedCount(processedCount);
+            jobRepository.updatePercentageCompleted(jobUuid, percentageCompleted);
+            log.info("[{}/{}] records processed = [{}% completed]", processedCount, totalCount, percentageCompleted);
```
dkrylovsb: We should update the job record if the _percentage_ has actually changed. Otherwise we may be issuing many `UPDATE` statements that actually don't change anything.

Should be simply a matter of adding `lastPercentageCompleted` to the tracker.

worker/src/main/resources/application.properties
```diff
@@ -16,6 +16,7 @@ spring.integration.jdbc.initialize-schema=always
 ## --------------------------------------------------------------------------------------------------  MISC CONFIG
 file.try.lock.timeout=30
 cancellation.check.frequency=10
+report.progress.frequency=500
```
dkrylovsb: Our smallest contract so far is 13K, which means the completion percentage can change as often as every _130_ patients processed. Let's set this property to `100`.

dkrylovsb:
Tests for the new feature are missing.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
spathiyil: done.
Added a `lastPercentageCompleted` & a separate `reportProgressLogFrequency` for logging the progress in the log.

worker/src/main/resources/application.properties
spathiyil: done

common/src/main/java/gov/cms/ab2d/common/repository/JobRepository.java
spathiyil: Changed to HQL.

PR #131, opened 2020-01-22
Feature/ab2d 777 opt out resolve ben id mbi

lailasharshar:
### Summary

Looks up the different ids for a patient from the BB server. The search is performed during the import of the Opt Out file. After the search is performed, it saves the data in the opt_out table in the database. It allows for multiple patients to come back with the same HICN id and saves all values. This does not change the worker to use the new ID.


optout/src/test/java/gov/cms/ab2d/optout/OptOutClientServiceTest.java
```diff
@@ -0,0 +1,124 @@
...
+    public static void setupBFDClient() throws IOException {
+        mockServer = ClientAndServer.startClientAndServer(mockServerPort);
+
+        createMockServerExpectation("/v1/fhir/metadata",
+                200,
+                getRawXML("test-data/meta.xml"),
+                List.of());
+
+        createMockServerExpectation(
+                "/v1/fhir/Patient",
```
adaykin: "/v1/fhir/Patient" is repeated a lot, it could be extracted to a constant.

bfd/src/test/java/gov/cms/ab2d/bfd/client/BlueButtonClientTest.java
```diff
@@ -198,6 +205,13 @@ public void shouldReturnBundleContainingOnlyEOBs() {
         ));
     }
 
+    @Test
+    public void testPersonIds() {
+        Bundle response = bbc.requestPatientFromServer("11111");
+        assertNotNull(response);
+        assertEquals(response.getEntry().size(), 2);
```
adaykin: Are you able to look at values in each entry and verify those?

bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClientImpl.java
```diff
@@ -62,43 +80,67 @@ public BFDClientImpl(IGenericClient bfdFhirRestClient) {
...
     public Bundle requestEOBFromServer(String patientID) {
-        return
-                fetchBundle(ExplanationOfBenefit.class,
-                        ExplanationOfBenefit.PATIENT.hasId(patientID));
+        return client.search()
+                .forResource(ExplanationOfBenefit.class)
+                .where(ExplanationOfBenefit.PATIENT.hasId(patientID))
+                .count(pageSize)
+                .returnBundle(Bundle.class)
+                .execute();
```
dkrylovsb: Curious of the reason for this change. This method does not seem to be related to the story.

bfd/src/main/resources/application.bfd.properties
```diff
@@ -1,11 +1,15 @@
...
 bfd.keystore.location=/bb.keystore
 bfd.serverBaseUrl=https://fhir.backend.bluebutton.hhsdevcloud.us/v1/fhir/
+bfd.hicn.hash=https://bluebutton.cms.gov/resources/identifier/hicn-hash
 bfd.connectionTimeout=5000
 bfd.socketTimeout=5000
 bfd.requestTimeout=5000
 bfd.pagesize=100
 
+bfd.hash.pepper=6E6F747468657265616C706570706572
+bfd.hash.iter=2
```
dkrylovsb: These two should refer to the environment variables -- AB2D_HICN_HASH_PEPPER & AB2D_HICN_HASH_ITER -- with a fallback to the defaults listed here. (I assume these work with BFD Sandbox).

optout/src/test/java/gov/cms/ab2d/optout/OptOutClientServiceTest.java
lailasharshar: I ended up extracting the whole methods out to reuse

bfd/src/main/resources/application.bfd.properties
lailasharshar: Okay. Sounds good. Yes, those work with the BFD sandbox although I think they recommend different values.

common/src/main/resources/db/changelog/v001/ab2d-777-resolve_diff_ids.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE opt_out ADD COLUMN mbi VARCHAR(255);
```
spathiyil: liquibase changelogs using sql format may need to begin with 

```
--liquibase formatted sql
```
and every change-set needs to begin with a line that looks something like this:

```
--changeset <author_id>:<change_set_id> failOnError:true
```
This is used by liquibase during the database migration process.

https://www.liquibase.org/documentation/sql_format.html

common/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -12,7 +12,9 @@ databaseChangeLog:
       file: db/changelog/v001/remove_not_null_job_output.sql
   - include:
       file: db/changelog/v001/ab2d-828-increase-username-column-length.sql
+  - include:
+      file: db/changelog/v001/ab2d-777-resolve_diff_ids.sql
```
spathiyil: Does a new change-log (and hence the change-sets defined inside it) need to be appended at the end?
Not sure if it is okay to insert them in the midst of existing change-logs like this.

lailasharshar: I'm not sure what you mean. I've never used liquidbase before so I just went with what looked right and it worked.

spathiyil: sounds good. 
I am used to seeing new change-logs being appended to the bottom of the list of existing change-logs and wasn't sure if adding it in the middle would cause any conflicts.  
Thanks for clarifying that it works fine.

optout/src/test/java/gov/cms/ab2d/optout/OptOutProcessorIntegrationTest.java
```diff
@@ -46,9 +51,33 @@
...
+    @BeforeAll
+    public static void setupBFDClient() throws IOException {
+        mockServer = ClientAndServer.startClientAndServer(mockServerPort);
+
+        MockBfdServiceUtils.createMockServerMetaExpectation("test-data/meta.xml", mockServerPort);
+        MockBfdServiceUtils.createMockServerPatientExpectation( TEST_DIR + "patientbundle.xml",
+                mockServerPort, List.of());
+    }
+
+    @Before
```
spathiyil: This may need to be `@BeforeEach`

PR #132, opened 2020-01-23
Feature/ab2d 747 filter out claims on dates

lailasharshar:
### Summary

Filter out explanation of benefit objects by dates subscribing and attestation date. I have it commented where it should go but still have questions about where the parameters should come from in the worker.

### Checklist

- [ ] Tests and linting pass

### Security
No security issues except to make sure the objects don't make it in that shouldn't


worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/PatientClaimsProcessorImpl.java
```diff
@@ -154,6 +154,12 @@ private void tryLock(Lock lock) {
     }
 
     private List<Resource> extractResources(List<BundleEntryComponent> entries) {
+        /**
+         * To filter out by date: (with params List<Integer> validMonths, year, attestationDate
+         *
+         *    List<FilterOutByDate> dateRanges = FilterOutByDate.getDateRanges(validMonths, year);
+         *    List<ExplanationOfBenefit> benefits = FilterOutByDate.filterByDate(resources, attestationDate, dateRanges);
+         */
```
dkrylovsb: `PatientDTO.monthsUnderContract` will have the coverage months you need.

dkrylovsb: ...and `Contract.attestedOn` is the attestation date.

bfd/src/test/java/gov/cms/ab2d/bfd/client/BFDClientConfigurationTest.java
```diff
@@ -1,38 +0,0 @@
-package gov.cms.ab2d.bfd.client;
```
dkrylovsb: I guess the deletion was unintentional?

lailasharshar: I have no memory of this

dkrylovsb: must have been accidental. Satheesh restored the file.

PR #133, opened 2020-01-23
Load the keystore from the classpath if available, otherwise load it …

adaykin:
Load keystore from fs or classpath, use env vars for keystore and password

### Summary

Load the keystore from the classpath if available, otherwise load it from the filesystem. In the properties file, first see if there is an env variable set, otherwise use a default value for password and keystore location.

### Checklist

- [x] Tests and linting pass

### Security

Keystore loading, setting passwords from an env variable

### DIFF
--- a/bfd/src/main/resources/application.bfd.properties
+++ b/bfd/src/main/resources/application.bfd.properties
@@ -1,5 +1,5 @@
-bfd.keystore.password=changeit
-bfd.keystore.location=/bb.keystore
+bfd.keystore.password=${BFD_KEYSTORE_PASSWORD:#{'changeit'}}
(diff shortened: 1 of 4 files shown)


bfd/src/main/resources/application.bfd.properties
```diff
@@ -1,5 +1,5 @@
-bfd.keystore.password=changeit
-bfd.keystore.location=/bb.keystore
+bfd.keystore.password=${BFD_KEYSTORE_PASSWORD:#{'changeit'}}
+bfd.keystore.location=${BFD_KEYSTORE_LOCATION:#{'/bb.keystore'}}
```
dkrylovsb: Our convention is to prefix all AB2D environment variables with `AB2D_`.

PR #134, opened 2020-01-24
Feature/ab2d 855 on failure cleanup threads in queue

spathiyil:
on failure, cleanup threads in queue so that the job comes to a stop sooner

### Summary

When an exception is thrown in the PatientClaimsProcessorImpl resulting in the job ending in a FAILED status,  ensure that the threads in queue are cleaned up so that the job comes to a stop sooner

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
@@ -18,6 +18,7 @@ import gov.cms.ab2d.worker.processor.domainmodel.ProgressTracker;
 import gov.cms.ab2d.worker.service.FileService;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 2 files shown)


worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -355,13 +355,23 @@ private int processHandles(List<Future<Integer>> futureHandles, ProgressTracker
...
-                    log.error(errMsg, e);
+                    cancelFuturesInQueue(futureHandles);
+                    log.error("exception while processing patient ", e);
+
+                    final String errMsg;
+                    if (e.getCause() != null) {
+                        errMsg = e.getCause().getMessage();
+                    } else {
+                        errMsg = e.getMessage();
+                    }
```
dkrylovsb: What if the cause has yet another cause? :)

Try https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/exception/ExceptionUtils.html#getRootCauseMessage-java.lang.Throwable-

spathiyil: done - updated code to use the ExceptionUtils to get the root cause and message

PR #137, opened 2020-01-28
ab2d 753 multiple output files per contract - draft - preview

spathiyil:
Enhance Worker  to create multiple output files when the contract has a huge number of patient records instead of creating one massive output file.

### Summary

Enhance Worker  to create multiple output files when the contract has a huge number of patient records instead of creating one massive output file. This is a draft version. Just wanted to give an early preview.

### Checklist

- [ ] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/common/src/main/java/gov/cms/ab2d/common/model/JobProgress.java
+++ b/common/src/main/java/gov/cms/ab2d/common/model/JobProgress.java
@@ -33,7 +33,7 @@ public class JobProgress {
     @EqualsAndHashCode.Include
     private Integer sliceNumber;
 
-    private Integer recordCount;
-
-    private Integer progress;
+    private Integer recordsProcessed;
+    private Integer sliceTotal;
+    private Integer percentageComplete;
 }
--- a/common/src/main/java/gov/cms/ab2d/common/model/JobProgress.java
+++ b/common/src/main/java/gov/cms/ab2d/common/model/JobProgress.java
@@ -35,5 +35,5 @@ public class JobProgress {
 
     private Integer recordsProcessed;
     private Integer sliceTotal;
-    private Integer percentageComplete;
+
 }
--- a/common/src/main/java/gov/cms/ab2d/common/model/JobProgress.java
+++ b/common/src/main/java/gov/cms/ab2d/common/model/JobProgress.java
@@ -33,6 +33,7 @@ public class JobProgress {
     @EqualsAndHashCode.Include
     private Integer sliceNumber;
 
+    private Integer recordCount;
 
     private Integer progress;
 }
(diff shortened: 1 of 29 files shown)


common/src/main/java/gov/cms/ab2d/common/model/JobProgress.java
```diff
@@ -0,0 +1,39 @@
+package gov.cms.ab2d.common.model;
```
dkrylovsb: Introducing "job slicing" adds considerable complexity to the domain & processing logic with unclear benefits. We are already processing jobs with the maximum degree of concurrency possible (each patient is an async call to BFD), so slices aren't needed here. Creating slices upfront and them managing them is not really needed to create multiple output files, either; I think there are simpler ways to achieve that.

See how logging frameworks roll log files when they reach a certain threshold. They do so transparently in the background without us knowing it, and of course without having us to "slice" streams of log information. This is exactly what I had in mind. We'd need a `JobOutputWriter` class or similar that encapsulates the file writing/rolling logic.

common/src/main/resources/db/changelog/v001/ab2d-753-create-table-job_progress.sql
```diff
@@ -0,0 +1,23 @@
...
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset spathiyil:AB2D-753-CreateTable-job_progress failOnError:true
+CREATE TABLE job_progress
+(
+    id                      BIGINT              NOT NULL,
+    job_id                  BIGINT              NOT NULL,
+    contract_id             BIGINT              NOT NULL,
+    slice_number            INTEGER             NOT NULL,
+    records_processed       INTEGER             NOT NULL,
```
dkrylovsb: Slicing makes tracking the overall job progress harder.

worker/src/main/java/gov/cms/ab2d/worker/processor/contract/ContractSliceCreatorImpl.java
```diff
@@ -0,0 +1,90 @@
+package gov.cms.ab2d.worker.processor.contract;
```
dkrylovsb: This class would be a good example of the "added complexity" I mentioned above. All the math logic here is pretty hard to follow.

dkrylovsb:
What I would recommend is, creating a new branch on which we could collaborate on designing the `JobOutputWriter` class and see if we could enable multiple output files per job using simpler ways?

spathiyil:
> What I would recommend is, create a new branch on which we'd collaborate on designing the `JobOutputWriter` class and see if we could enable multiple output files per job using simpler ways?

Okay. Declining this branch/pr

PR #140, opened 2020-01-29
Feature/ab2d 785 load testing

adaykin:
Adding in load testing so that load tests can be run from jmeter and show stats on how long each section took.

### Summary

Refactoring code into new class APIClient so that it can be reusable, adding JMeter tests that can be run from a JMeter command line or GUI, load tests will keep track of each request's time that it took. The tests are configurable so that they can hit different environments. Instructions are provided on how to run the tests from JMeter.

### Checklist

- [x] Tests and linting pass

### Security

NA


common/src/main/java/gov/cms/ab2d/common/httpclient/APIClient.java
```diff
@@ -0,0 +1,173 @@
+package gov.cms.ab2d.common.httpclient;
```
dkrylovsb: Reusable `APIClient` makes sense, but can we keep it under `e2e` and make the load test import it?

common/src/main/java/gov/cms/ab2d/common/httpclient/APIClient.java
```diff
@@ -0,0 +1,173 @@
...
+                .header("Authorization", "Bearer " + jwtStr)
+                .GET()
+                .build();
+    }
+
+    public CompletableFuture<HttpResponse<String>> exportByContractRequestAsync(String contractNumber) {
+        HttpRequest exportRequest = buildExportByContractRequest(contractNumber);
+
+        return httpClient.sendAsync(exportRequest, HttpResponse.BodyHandlers.ofString());
+    }
```
dkrylovsb: Are we using this?

common/src/main/java/gov/cms/ab2d/common/httpclient/APIClient.java
```diff
@@ -0,0 +1,173 @@
...
+                .header("Authorization", "Bearer " + jwtStr)
+                .GET()
+                .build();
+    }
+
+    public CompletableFuture<HttpResponse<String>> statusRequestAsync(String url) {
+        HttpRequest statusRequest = buildStatusRequest(url);
+
+        return httpClient.sendAsync(statusRequest, HttpResponse.BodyHandlers.ofString());
+    }
```
dkrylovsb: ...and this?

common/src/main/java/gov/cms/ab2d/common/httpclient/APIClient.java
dkrylovsb: You could move it under 'src/test` in e2e, and then use https://maven.apache.org/plugins/maven-jar-plugin/examples/create-test-jar.html to produce a "test jar" so that the load test could import it as a dependency.

load-test/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+    <properties>
+        <project.root>${basedir}/..</project.root>
+        <jmeter.version>5.2.1</jmeter.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>gov.cms.ab2d</groupId>
+            <artifactId>common</artifactId>
+            <version>0.0.1-SNAPSHOT</version>
```
dkrylovsb: So this one would import that test jar instead.

common/src/main/java/gov/cms/ab2d/common/httpclient/APIClient.java
adaykin: Yeah I think that makes sense.

common/src/main/java/gov/cms/ab2d/common/httpclient/APIClient.java
adaykin: I was using that in some of the e2e testing before but I can delete now.

load-test/src/main/java/gov/cms/ab2d/loadtest/TestRunner.java
```diff
@@ -0,0 +1,198 @@
...
+                statusResult.sampleStart();
+
+                // Eventually back off, otherwise the server will infinitely send a 429
+                int i = 0;
+                int delay = 50;
+                boolean finishedStatus = false;
+                while (!finishedStatus) {
+                    Thread.sleep(delay);
+                    statusResponse = apiClient.statusRequest(url);
+                    status = statusResponse.statusCode();
```
dkrylovsb: We'd also want to make sure we produce a metric to indicate how each _individual_ status request is taking in addition to the overall job completion metric that you already have.

common/pom.xml
```diff
@@ -69,6 +69,12 @@
             <artifactId>modelmapper</artifactId>
             <version>2.3.5</version>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin.external.google</groupId>
+            <artifactId>android-json</artifactId>
+            <version>0.0.20131108.vaadin1</version>
+            <scope>compile</scope>
+        </dependency>
```
dkrylovsb: I assume we can now drop this.

common/src/main/java/gov/cms/ab2d/common/util/JobUtil.java
```diff
@@ -0,0 +1,8 @@
+package gov.cms.ab2d.common.util;
+
+public class JobUtil {
+
+    public static String getJobUuid(String url) {
+        return url.substring(url.indexOf("/Job/") + 5, url.indexOf("/$status"));
+    }
+}
```
dkrylovsb: And this can also be moved out of "commons"

PR #141, opened 2020-01-29
Setting max concurrent jobs to 3

adaykin:
Setting max concurrent jobs to 3

### Summary

Add a column in the user_account table that will set the maximum concurrent jobs a user can have to 1 by default, and adding a migration that will change it to 3 for all users. The exports will enforce that a user can only have a max number of jobs to whatever their limit is set to.

### Checklist

- [x] Tests and linting pass

### Security

NA

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
@@ -99,11 +99,7 @@ public class BulkDataAccessAPI {
             @RequestParam(required = false, name = "_outputFormat") String outputFormat) {
         log.info("Received request to export");
 
-        if (jobService.checkIfCurrentUserHasActiveJob()) {
-            String errorMsg = "User already has an active or submitted job";
-            log.error(errorMsg);
-            throw new TooManyRequestsException(errorMsg);
-        }
+        checkIfCurrentUserCanAddJob();
 
         checkResourceTypesAndOutputFormat(resourceTypes, outputFormat);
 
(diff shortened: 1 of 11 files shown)


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -114,6 +110,14 @@
         return returnStatusForJobCreation(job);
     }
 
+    private void checkIfCurrentUserCanAddJob() {
+        if (!jobService.checkIfCurrentUserCanAddJob()) {
+            String errorMsg = "User already has enough active or submitted jobs for their limit";
```
dkrylovsb: This message will be presented to the users, I assume. Let's word it a bit differently so that it has more clarity: `You already have active export requests in progress. Please wait until they complete before submitting a new one.`

common/src/main/resources/db/changelog/v001/add_concurrent_job_column.sql
```diff
@@ -0,0 +1,7 @@
+--liquibase formatted sql
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset adaykin:add_concurrent_job_column failOnError:true
+ALTER TABLE user_account ADD COLUMN max_parallel_jobs INT DEFAULT 1 NOT NULL;
+
+UPDATE user_account SET max_parallel_jobs = 3;
```
dkrylovsb: Let's also update `AB2D-PDP-MULTICONTRACT` user account to have the limit at, say, 30.

adaykin: 0oa2t7wfhvB2qSqPg297 is the username connected to that sponsor, I think you mean that one?

dkrylovsb: Yes

PR #143, opened 2020-01-30
Created bash script to automate submittal of EOB bulk download

lailasharshar:
### Summary

This script obtains a token from Okta, submits a bulk download job, checks its status and once the job is done, displays the results. There are two variables at the top of the script that allow
configuration. One is the sandbox URL. This can be changed to different URLs like dev or localhost. The second is the Base64 encoded clientId:clientPassword for the basic auth to Okta. Sample values are found in the sandbox documentation.

Pieces of this are being used in the sandbox documentation but in a contextual way that describe each action in the cURL example.

### Checklist

- [ ] Tests and linting pass

### Security
This script does contain the encoded client id and password of a sample user.

AB2D-876 - I forgot to put this in the branch title.

### DIFF
--- a/Deploy/bash/geteob.sh
+++ b/Deploy/bash/geteob.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# Set to the name of the service you're testing
+SANDBOX=http://sandbox.ab2d.cms.gov/api/v1/fhir
+
+# Set up the auth to use for Okta
+AUTH=MG9hMnQwbHNyZFp3NXVXUngyOTc6SEhkdVdHNkxvZ0l2RElRdVdncDNabG85T1lNVmFsVHRINU9CY3VIdw==
+
+
+# Get the Bearer token from Okta
+
+BEARER_TOKEN=$(curl -X POST "https://test.idp.idm.cms.gov/oauth2/aus2r7y3gdaFMKBol297/v1/token?grant_type=client_credentials&scope=clientCreds" \
+        -H "Content-Type: application/x-www-form-urlencoded" \
+        -H "Accept: application/json" \
+        -H "Authorization: Basic ${AUTH}" | jq --raw-output ".access_token")
+
+# Create a job
+
+JOB=$(curl "${SANDBOX}/Patient/\$export?_outputFormat=application%2Ffhir%2Bndjson&_type=ExplanationOfBenefit" \
+        -sD - \
+        -H "accept: application/json" \
+        -H "Accept: application/fhir+json" \
+        -H "Prefer: respond-async" \
+        -H "Authorization: Bearer ${BEARER_TOKEN}" | grep "Content-Location" | sed 's/.*Job.//' | sed 's/..status//' | tr -d '[:space:]')
+echo $JOB created
+
+# Get the status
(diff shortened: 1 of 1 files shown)


Deploy/bash/geteob.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# Set to the name of the service you're testing
+SANDBOX=http://sandbox.ab2d.cms.gov/api/v1/fhir
+
+# Set up the auth to use for Okta
+AUTH=MG9hMnQwbHNyZFp3NXVXUngyOTc6SEhkdVdHNkxvZ0l2RElRdVdncDNabG85T1lNVmFsVHRINU9CY3VIdw==
```
dkrylovsb: This requires the user of this script to figure out the base64-encoded authorization header, which is slightly inconvenient albeit not critical. Should the script rather take in client id & client secret instead? The `Authorization` header can be derived in-script via ```export AUTH=`echo -n client_id-value:client_pass_value | base64` ```

Deploy/bash/geteob.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# Set to the name of the service you're testing
+SANDBOX=http://sandbox.ab2d.cms.gov/api/v1/fhir
```
dkrylovsb: Since this is a shell script to run from command line, would it make more sense to pass these values via command line parameters instead? `./geteob --url URL --clientId ... --clientPass ...`?

Deploy/bash/geteob.sh
```diff
@@ -0,0 +1,47 @@
...
+        sleep 5
+done
+
+# Print out the list of URLs
+echo $URLS
+
+# Iterate over the URLs
+for URL in ${URLS}
+do
+        curl "$URL" \
```
dkrylovsb: This `curl` will quickly dump the files to the console and then they're gone for good. Should we save the files to disk as well? Either by adding an option to `curl` or using `wget` instead, whichever works.

Deploy/bash/geteob.sh
lailasharshar: I actually put the Base64 encoding for all the values in the API Use confluence page. I had originally used the client id and password that, but I was getting a different second to last character than when it expected using the bas64 command in my script. I'm not entirely sure why that was happening, so to eliminate any similar issues for the person using the sandbox, I just put all the encoded values on the web page.

Deploy/bash/geteob.sh
lailasharshar: There are actually a lot of things we could do (there is very little error checking). I mostly used this as the verifiable way to make sure the commands I put in my Curl tutorial https://confluence.cms.gov/display/AB2D/Curl+Tutorial was correct and as a means to support our development. I think the documentation in the tutorial is sufficient that the script doesn't really need to be included on the web site and it can be useful for us. Maybe abstracting the URL and the base64 encoded client id and password are sufficient. Actually, probably also the contract number as an optional parameter.

Deploy/bash/geteob.sh
lailasharshar: We could name them the same as the files we're downloading.

PR #145, opened 2020-01-30
ab2d-931 - refactor async call to return void return type

spathiyil:
refactor async call to return void return type

### Summary

After the design change to enable worker to create multiple output files, the logic no longer needs the errorCount to be returned by the `PatientClaimsProcessor` class. `JobProcessor` is able to use the `dataFiles` & `errorFiles` obtained from the `JobDataWriter` class to create the relevant records in the `JobOutput`  table. This refactor removes the redundant `errorCount` return type from being returned from the async call and the corresponding clean up of the code in `JobProcessor`

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
@@ -238,11 +238,10 @@ public class JobProcessorImpl implements JobProcessor {
         int patientCount = patients.size();
         log.info("Contract [{}] has [{}] Patients", contractNumber, patientCount);
 
-        int errorCount = 0;
         boolean isCancelled = false;
 
         int recordsProcessedCount = 0;
-        var futureHandles = new ArrayList<Future<Integer>>();
+        var futureHandles = new ArrayList<Future<Void>>();
         for (PatientDTO patient : patients) {
             ++recordsProcessedCount;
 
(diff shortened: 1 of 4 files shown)


worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -331,18 +330,13 @@ private boolean isOptOutPatient(String patientId) {
...
         var iterator = futureHandles.iterator();
         while (iterator.hasNext()) {
             var future = iterator.next();
             if (future.isDone()) {
                 try {
-                    var responseCount = future.get();
-                    if (responseCount > 0) {
-                        errorCount += responseCount;
-                    }
+                    future.get();
```
adaykin: Curious why this approach was chosen of periodically sending a list to the processHandles method vs. using the API for all futures to complete like https://stackoverflow.com/a/51006865/12265284

spathiyil: Well, in that case, all the calls are made in a burst mode at the outset and it just needs to await for all of them to return and join. Since our batch could run for a very long time, we do some work periodically (eg track process etc).

adaykin: Ahh ok, was trying to think of ways to reduce complexity by possibly removing the 2nd call to processHandles

PR #147, opened 2020-02-03
add new parameter to set sslmode to connect to db

spathiyil:
Add a new parameter `sslmode` in the db connection URL to allow ssl connection to be turned on in specific environments

### Summary

Add a new parameter `sslmode` in the db connection URL to allow ssl connection to be turned on in specific environments
Set the default value to `allow` which will **not** require ssl connection for dev and other non-prod environments. 
To turn on ssl in prod(and any other environments) , the value could be set to `verify-full` by setting the environment variable `AB2D_DB_SSL_MODE`

### Checklist

- [x] Tests and linting pass

### DIFF
--- a/api/src/main/resources/application.properties
+++ b/api/src/main/resources/application.properties
@@ -1,4 +1,4 @@
-spring.datasource.url=jdbc:postgresql://${AB2D_DB_HOST}:${AB2D_DB_PORT}/${AB2D_DB_DATABASE}?sslmode=${AB2D_DB_SSL_MODE}
+spring.datasource.url=jdbc:postgresql://${AB2D_DB_HOST}:${AB2D_DB_PORT}/${AB2D_DB_DATABASE}?sslmode=${AB2D_DB_SSL_MODE:allow}
 spring.datasource.username=${AB2D_DB_USER}
 spring.datasource.password=${AB2D_DB_PASSWORD}
 spring.jpa.properties.hibernate.temp.use_jdbc_metadata_defaults=false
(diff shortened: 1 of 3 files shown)


api/src/main/resources/application.properties
```diff
@@ -1,4 +1,4 @@
-spring.datasource.url=jdbc:postgresql://${AB2D_DB_HOST}:${AB2D_DB_PORT}/${AB2D_DB_DATABASE}
+spring.datasource.url=jdbc:postgresql://${AB2D_DB_HOST}:${AB2D_DB_PORT}/${AB2D_DB_DATABASE}?sslmode=${AB2D_DB_SSL_MODE}
```
dkrylovsb: Can we specify the default value _inline_ as well, so that we don't even need to bother specifying the environment variable if we don't want SSL?  `${AB2D_DB_SSL_MODE:allow}` or similar.

docker-compose.yml
```diff
@@ -23,6 +23,7 @@ services:
       - AB2D_DB_HOST=db
       - AB2D_DB_PORT=5432
       - AB2D_DB_DATABASE=ab2d
+      - AB2D_DB_SSL_MODE=allow
```
dkrylovsb: Let's still keep this for clarity.

worker/src/main/resources/application.properties
```diff
@@ -1,5 +1,5 @@
 ## -------------------------------------------------------------------------------------------  DATA-SOURCE CONFIG
-spring.datasource.url=jdbc:postgresql://${AB2D_DB_HOST}:${AB2D_DB_PORT}/${AB2D_DB_DATABASE}
+spring.datasource.url=jdbc:postgresql://${AB2D_DB_HOST}:${AB2D_DB_PORT}/${AB2D_DB_DATABASE}?sslmode=${AB2D_DB_SSL_MODE}
```
dkrylovsb: Here too

docker-compose.yml
adaykin: Does verify-full make more sense since it will do more verification for us or will that mess things up when we don't have an actual vert like in dev? https://jdbc.postgresql.org/documentation/head/ssl-client.html Maybe it's for for when using a prod env.

spathiyil: My understanding after talking to @dkrylovsb  is that the `sslmode` would be turned on only in specific environments (such as for eg, prod).

worker/src/main/resources/application.properties
spathiyil: done

api/src/main/resources/application.properties
spathiyil: done

PR #154, opened 2020-02-06
Feature/ab2d 935 e2e tests

adaykin:
Adding more e2e tests for security, http request parameters, and gzip

### Summary

Add e2e tests that check that a user cannot make a request without a token, cannot use a self signed token, cannot access another user's job, verify gzip content, test for parameters sent for the export request.

### Checklist

- [x] Tests and linting pass

### Security

Testing for tokens, users accessing other user's jobs


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -321,7 +318,7 @@ private String getUrlPath(Job job, String filePath) {
     )
     @ResponseStatus(value = HttpStatus.OK)
     @GetMapping(value = "/Job/{jobUuid}/file/{filename}")
-    public ResponseEntity<Void> downloadFile(
+    public ResponseEntity<String> downloadFile(
```
dkrylovsb: Usually changing return types affects Swagger documentation, because Spring Fox generates Swagger from the code. Please take a look at Swagger to make sure nothing odd popped up as a result of this. @lailasharshar might help out as well.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -330,19 +327,17 @@ private String getUrlPath(Job job, String filePath) {
...
             IOUtils.copy(in, out);
 
             jobService.deleteFileForJob(downloadResource.getFile());
 
-            response.setHeader(HttpHeaders.CONTENT_TYPE, NDJSON_FIRE_CONTENT_TYPE);
+            HttpHeaders httpHeaders = new HttpHeaders();
+            httpHeaders.add("Content-Type", NDJSON_FIRE_CONTENT_TYPE);
 
-            return new ResponseEntity<>(null, null, HttpStatus.OK);
+            return new ResponseEntity<>(out.toString(), httpHeaders, HttpStatus.OK);
```
dkrylovsb: OK, I think I know why you had to  make changes to this method. The response wasn't setting the Content Type header? In this case simply move the `setHeader` call before `copy`; no other changes (including changing the return type) would be required.

The issue with the changes shown here is that they introduce significant additional memory consumption: writing to BAOS effectively copies the entire NDJSON file in memory, and those can get large.

common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -116,6 +116,13 @@ public Job updateJob(Job job) {
     public Resource getResourceForJob(String jobUuid, String fileName) throws MalformedURLException {
         Job job = getJobByJobUuid(jobUuid);
 
+        User user = userService.getCurrentUser();
+        if (!user.equals(job.getUser())) {
+            log.error("User attempted to download a file where they had a valid UUID, but was not logged in as the " +
+                    "user that created the job");
+            throw new InvalidJobAccessException("You don't have permissions to access the job " + jobUuid);
+        }
+
```
dkrylovsb: I am _very_ happy we caught this.

dkrylovsb: Let's move this check into `getJobByJobUuid`. That way we also enforce access checks in `getJobStatus` and `deleteRequest`.

dkrylovsb: Let's also rename `getJobByJobUuid` into `getAuthorizedJobByJobUuid`

common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -116,6 +116,13 @@ public Job updateJob(Job job) {
     public Resource getResourceForJob(String jobUuid, String fileName) throws MalformedURLException {
         Job job = getJobByJobUuid(jobUuid);
 
+        User user = userService.getCurrentUser();
+        if (!user.equals(job.getUser())) {
+            log.error("User attempted to download a file where they had a valid UUID, but was not logged in as the " +
+                    "user that created the job");
+            throw new InvalidJobAccessException("You don't have permissions to access the job " + jobUuid);
```
dkrylovsb: The problem with this error message text is that it serves as a confirmation to the unauthorized user that a job with such UUID indeed exists. We don't really need to tell them that. Let's just return `Unauthorized`.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -293,69 +295,112 @@ public void testDelete() throws IOException, InterruptedException {
...
+
+    @Test
+    public void testBadQueryParameterOutputFormat() throws IOException, InterruptedException {
+        var params = new HashMap<>(){{
+            put("_outputFormat", "BadParam");
+        }};
+        HttpResponse<String> exportResponse = apiClient.exportRequest(params);
+
+        Assert.assertEquals(400, exportResponse.statusCode());
+    }
```
dkrylovsb: Given the earlier comments, we need two more test cases:
- User cannot check status of a job that does not belong to them
- User cannot cancel other's job

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -327,17 +330,19 @@ private String getUrlPath(Job job, String filePath) {
 
         Resource downloadResource = jobService.getResourceForJob(jobUuid, filename);
 
+        RequestAttributes requestAttributes = RequestContextHolder.getRequestAttributes();
```
dkrylovsb: Are you using `requestAttributes` anywhere?

adaykin: Yes, it's used to get the response.

PR #155, opened 2020-02-06
ab2d 949 ensure there is at least one JobOutput upon processing a contract

spathiyil:
At the end of processing a contract, one more more JobOutput records should be created. 
If none were created, set the job to failed status

### Summary

At the end of processing a contract, one more more JobOutput records should be created. 
If none were created, set the job to failed status

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -402,6 +404,11 @@ private void trackProgress(ProgressTracker progressTracker) {
                 .collect(Collectors.toList());
         jobOutputs.addAll(errorJobOutputs);
 
+        if (jobOutputs.isEmpty()) {
+            var errMsg = "No JobOutput records were created.";
```
dkrylovsb: Will this message propagate all the way to the end user? If yes, let's make it somewhat less technical, i.e. "The export process has produced no results".

spathiyil: done - error message changed.

PR #156, opened 2020-02-06
Bug/ab2d 953 simple date format not thread safe

lailasharshar:
### Summary
The FilterOutByDate class had a SimpleDateFormat private attribute. SimpleDateFormat is not thread safe and the this class is used by the worker which is multi-threaded. There are three alternatives to to this:

1. Update the date formatting to a new version - The problem with that is the the FHIR library uses the Date object and I didn't want to introduce errors by converting back and forth given unknown time zone information.
2. Make the method that uses it synchronized - seems like overkill and potentially blocking
3. Take the SimpleDateFormat and put it in the methods that uses it, creating an instance each time you need it - I decided this was the simplest way and since it isn't a terrible expensive object, the best.

### Checklist

- [ ] Tests and linting pass

### DIFF
--- a/filter/src/main/java/gov/cms/ab2d/filter/FilterOutByDate.java
+++ b/filter/src/main/java/gov/cms/ab2d/filter/FilterOutByDate.java
@@ -17,8 +17,8 @@ import java.util.stream.Collectors;
  * and to minimize time zone issues, we kept everything as Date objects.
  */
 public final class FilterOutByDate {
-    private static SimpleDateFormat sdf = new SimpleDateFormat("MM/dd/yyyy");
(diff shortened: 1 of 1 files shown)


filter/src/main/java/gov/cms/ab2d/filter/FilterOutByDate.java
```diff
@@ -39,13 +39,15 @@
          * @throws ParseException if there was an error constructing the Date objects
          */
         public DateRange(Date start, Date end) throws ParseException {
+            SimpleDateFormat fullDateFormat = new SimpleDateFormat(FULL);
```
dkrylovsb: This variable does not appear to be used within the constructor; we can probably drop it.

PR #158, opened 2020-02-06
If the file can be connected to a job, retrieve the job so that the d…

adaykin:
…eletion can factor in whether or not the job has expired yet. If the file is not connected to a job, use the existing logic.

### Summary

If the file can be connected to a job, retrieve the job so that the deletion can factor in whether or not the job has expired yet. The completion date of the job will be used to determine if the job should be deleted or not. If the file is not connected to a job, use the existing logic.

### Checklist

- [x] Tests and linting pass

### Security

NA


audit/pom.xml
```diff
@@ -30,6 +30,14 @@
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>gov.cms.ab2d</groupId>
+            <artifactId>common</artifactId>
+            <version>0.0.1-SNAPSHOT</version>
+            <type>test-jar</type>
+            <scope>test</scope>
```
spathiyil: I think it might be a bit bit odd that the `type` is `test-jar` and the `scope` is `test`.
What happens if we remove these two attributes from here? Will it not work?

adaykin: It's only for test purposes, we wouldn't want this to be used for production. There's a JAR that contains reusable components for data setup in test.

spathiyil: Sounds good.

PR #161, opened 2020-02-09
Feature/ab2d 940 health checks

lailasharshar:
### Summary

Perform health checks for both API & Worker. 
1. Created generic health checks in commons for:
    - Sufficient Memory
    - Ability to write to a specified directory (current & EFS)
    - Ability to get a connection to the database
    - Ability to  log (although we should check to see if it makes it eventually into Splunk)
    - Ability to access to the Internet
2. Create endpoints for both worker and API called "/health" that is not password protected that use a combination of these checks as necessary to verify that the service is up and if not, be bounced.

### Checklist

- [ ] Tests and linting pass

### Security
This /health endpoint is not secure by design, but shouldn't be universally available to the outside. My random file name generator does trigger an issue with the secrets scan because it sees my full alphabet string is a password, but an exclusion will be created.


common/src/main/java/gov/cms/ab2d/common/health/DatabaseAvailable.java
```diff
@@ -0,0 +1,26 @@
...
+     */
+    public static boolean isDbAvailable(DataSource datasource) {
+        try (Connection conn = datasource.getConnection()) {
+            Statement stmt = conn.createStatement();
+            stmt.execute("SELECT 1");
+            return true;
+        } catch (Exception ex) {
+            return false;
+        }
+    }
```
dkrylovsb: Need to close both the statement and connection in a `finally` clause to avoid pool leaks.

common/src/main/java/gov/cms/ab2d/common/health/FileSystemCheck.java
```diff
@@ -0,0 +1,76 @@
...
+     * @param ext - the extension, such as "txt"
+     * @return - the generated file
+     */
+    public static String getRandomFileName(int length, String ext) {
+        String alphas = "ABCDEFGHIJKLMNOPQRSTUVWXYabcdefghijklmnopqrstuvwxyz";
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < length; i++) {
+            int index = (int) (alphas.length() * Math.random());
+            sb.append(alphas.charAt(index));
+        }
```
dkrylovsb: A shortcut: `org.apache.commons.lang3.RandomStringUtils.randomAlphabetic(length)`

dkrylovsb: ...`randomAlphanumeric` is probably even better to increase entropy.

common/src/main/java/gov/cms/ab2d/common/health/MemoryUtilization.java
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * This is done with a two dimensional array. It's an array of one MB arrays of bytes. I split it up
+     * so that we didn't have to have one contiguous set of memory
+     *
+     * @param numMBToCreate - the number of megabytes to create
+     * @return if the JVM runs out of memory while trying to allocate the array
+     */
+    public static boolean outOfMemory(int numMBToCreate) {
+        try {
+            byte[][] arr = new byte[numMBToCreate][1024];
```
dkrylovsb: Let's actually fill the array with random bytes. Who knows what kind of optimizations JVM undertakes when initializing (but not yet writing to) an array; may be it does _lazy init_ of some kind 🤷‍♂ . Switching to one-dimensional array will make this easier, too.

common/src/main/java/gov/cms/ab2d/common/health/UrlAvailable.java
```diff
@@ -0,0 +1,53 @@
...
+                log.debug("Socket opened to " + testUrl + " and is connected: " + socket.isConnected());
+            } catch (IOException e) {
+                log.info("Unable to open socket to " + testUrl);
+                return false;
+            }
+            return true;
+        } catch (MalformedURLException e1) {
+            log.info("Malformed URL: " + testUrl);
+            return false;
+        }
```
dkrylovsb: A shortcut: https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#toString(java.net.URL,%20java.nio.charset.Charset)

common/src/test/java/gov/cms/ab2d/common/health/DatabaseAvailableTest.java
```diff
@@ -0,0 +1,316 @@
...
+    @Container
+    private static final PostgreSQLContainer postgreSQLContainer= new AB2DPostgresqlContainer();
+
+    @Test
+    public void testDatasource() throws SQLException {
+        assertTrue(DatabaseAvailable.isDbAvailable(dataSource));
+        assertFalse(DatabaseAvailable.isDbAvailable(null));
+        Mockito.when(bogusDS.getConnection()).thenReturn(null);
+        assertFalse(DatabaseAvailable.isDbAvailable(bogusDS));
+        Mockito.when(bogusDS.getConnection()).thenReturn(new Connection() {
```
dkrylovsb: Hmm. Curious why we're having to do this. You could just create a `Connection` mock. And in case you need one (or more) of its methods to actually do _non-mock_ stuff (an action or something), Mockito has API for that as well.

common/src/test/java/gov/cms/ab2d/common/health/FileSystemCheckTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+class FileSystemCheckTest {
+
+    @Test
+    void canWriteFile() throws IOException {
+        assertFalse(FileSystemCheck.canWriteFile(null, false));
+        assertTrue(FileSystemCheck.canWriteFile("/tmp", false));
+        assertTrue(FileSystemCheck.canWriteFile(".", false));
+        assertFalse(FileSystemCheck.canWriteFile("/notarealdir", false));
+        assertFalse(FileSystemCheck.canWriteFile("/bin", false));
```
dkrylovsb: Is the idea this line will fail because the test isn't supposed to run as root?

dkrylovsb: ...I'd probably avoid trying to write to `/bin` (or any other system directory for that matter) in a test: depending on an individual developer's workstation configuration (permissions, etc), this write might actually succeed.

common/src/test/java/gov/cms/ab2d/common/health/FileSystemCheckTest.java
```diff
@@ -0,0 +1,44 @@
...
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
+        assertNotEquals(FileSystemCheck.getRandomFileName(10, "txt"), FileSystemCheck.getRandomFileName(10, "txt"));
```
dkrylovsb: Should we add a `for` loop?

common/src/test/java/gov/cms/ab2d/common/health/MemoryUtilizationTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class MemoryUtilizationTest {
+    @Test
+    public void testMemory() {
+        assertFalse(MemoryUtilization.outOfMemory(2));
+        assertTrue(MemoryUtilization.outOfMemory((int) (Math.pow(2, 31) - 1)));
```
dkrylovsb: Nice!

common/src/main/java/gov/cms/ab2d/common/health/MemoryUtilization.java
lailasharshar: I actually ran into that problem with a one-dimensional array.

common/src/test/java/gov/cms/ab2d/common/health/FileSystemCheckTest.java
lailasharshar: How else do you get it to fail except to write to forbidden directories? I guess you could create a read only directory and then try. If the app can write to /bin, maybe you want to know that.

common/src/main/java/gov/cms/ab2d/common/health/DatabaseAvailable.java
lailasharshar: Doesn't the try with resource close the connection?

common/src/main/java/gov/cms/ab2d/common/health/FileSystemCheck.java
lailasharshar: Yes, except that you don't want it to start with a numeric so I didn't bother. I just wanted to make sure that the next time you wrote the file, it wouldn't be the same name so garbage left in the directory from the last test wouldn't cause issues.

common/src/main/java/gov/cms/ab2d/common/health/MemoryUtilization.java
dkrylovsb: Ah. Max array size is the max Integer. Makes sense.

lailasharshar: The one-dimensional array actually seemed to be optimized and I couldn't get it to fail. That's why I chose the 2 dimensional.

common/src/main/java/gov/cms/ab2d/common/health/DatabaseAvailable.java
adaykin: I think you need both the statement and connection in the try (...) section

common/src/test/java/gov/cms/ab2d/common/health/FileSystemCheckTest.java
dkrylovsb: Creating a R/O directory as you said might work. 

Another challenge with `/bin` is that it makes the test platform-dependent in some way, i.e Windows does not have bin.

common/src/main/java/gov/cms/ab2d/common/health/DatabaseAvailable.java
dkrylovsb: @lailasharshar My bad.

Yeah, just add the statement to the `try` as Andy is suggesting

common/src/main/java/gov/cms/ab2d/common/health/UrlAvailable.java
lailasharshar: I just want to check to see if I can connect on the necessary port. I don't want the contents. That seems like a more expensive operation.

common/src/main/java/gov/cms/ab2d/common/health/FileSystemCheck.java
dkrylovsb: Wondering why can't it start with a numeric?

common/src/test/java/gov/cms/ab2d/common/health/DatabaseAvailableTest.java
lailasharshar: Lazy. When I entered new Connection, it auto-generated the methods that you have overwrite and it worked so I didn't bother changing it figuring I could fill the overwritten methods with other stuff later if I wanted better testing. I'll refactor.

common/src/main/java/gov/cms/ab2d/common/health/MemoryUtilization.java
dkrylovsb: Yeah. A one-dimensional byte array can't be larger than ~2GB, you likely have more memory available.

common/src/main/java/gov/cms/ab2d/common/health/UrlAvailable.java
dkrylovsb: 👍

common/src/main/java/gov/cms/ab2d/common/health/FileSystemCheck.java
lailasharshar: I noticed that people generating random file names didn't use it (and made a point of saying not to), although the reason seems unclear with further googling. I guess it's just an old school convention so you don't get in trouble with old operating systems like using spaces in filenames or multiple periods. You can probably get away with it, but why if you have a choice. Pick the most inoffensive choice possible that no operating system or application that loads files will confuse.

dkrylovsb: 👍

PR #162, opened 2020-02-10
ab2d 973 introduce failure threshold for the worker

spathiyil:
introduce failure threshold for the worker

### Summary

Given a job, the worker processes all the patients in the contract. Introduce a failure threshold so that the batch is not stopped on the first failure. Instead when a failure occurs, check against an externally configurable failure threshold and if the number of errors crosses the threshold, only then stop processing the job.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
@@ -342,9 +342,9 @@ public class JobProcessorImpl implements JobProcessor {
         while (iterator.hasNext()) {
             var future = iterator.next();
             if (future.isDone()) {
+                progressTracker.incrementProcessedCount();
                 try {
                     future.get();
-                    progressTracker.incrementProcessedCount();
                 } catch (InterruptedException | ExecutionException e) {
                     analyzeException(futureHandles, progressTracker, e);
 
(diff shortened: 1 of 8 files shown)


worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
```diff
@@ -57,5 +64,9 @@ public int getPercentageCompleted() {
         return percentCompleted;
     }
 
+    public boolean isErrorCountBelowThreshold() {
+        return (failureCount * 100) / processedCount < failureThreshold;
```
dkrylovsb: The failure threshold should be calculated against the total number of patients in the contract.

spathiyil: done.

worker/src/main/resources/application.properties
```diff
@@ -19,6 +19,9 @@ cancellation.check.frequency=10
 report.progress.db.frequency=100
 report.progress.log.frequency=10000
 
+## fail the job if >= 10% of the records fail
+failure.threshold=10
```
dkrylovsb: Let's start with a conservative default of 1%. Still a large number for large contracts.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -367,6 +361,20 @@ private void processHandles(List<Future<Void>> futureHandles, ProgressTracker pr
...
+    private void analyzeException(List<Future<Void>> futureHandles, ProgressTracker progressTracker, Exception e) {
+        progressTracker.incrementFailureCount();
+        progressTracker.incrementProcessedCount();
+        if (progressTracker.isErrorCountBelowThreshold()) {
+            final Throwable rootCause = ExceptionUtils.getRootCause(e);
+            log.error("exception while processing patient {}", rootCause.getMessage(), rootCause);
+            // log exception, but continue processing job as errorCount is below threshold
+        } else {
+            cancelFuturesInQueue(futureHandles);
+            log.error("{} out of {} records failed. Stopping job", progressTracker.getFailureCount(), progressTracker.getProcessedCount());
```
dkrylovsb: I guess now the 2nd parameter should be `totalCount`.

worker/src/test/resources/application.properties
```diff
@@ -22,6 +22,9 @@ cancellation.check.frequency=2
 report.progress.db.frequency=10
 report.progress.log.frequency=10
 
+## fail the job if >= 10% of the records fail
+failure.threshold=10
```
dkrylovsb: Fine to keep `10` in tests of course

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
spathiyil: changed.

PR #176, opened 2020-02-13
Adding dispatcher annotation

adaykin:
Adding dispatcher annotation

Fixing the custom job processing metric

### Checklist

- [x] Tests and linting pass

### Security

NA


dkrylovsb:
@lhanekam My understanding is that _custom_ transactions will appear as non-Web; I've verified that by looking at Andy's App within NR.

PR #182, opened 2020-02-14
Feature/ab2d 943 add zip to api endpoint

lailasharshar:
### Summary

Added the ability to download data files as .zip files and combine multiple ndjson files into one
- Added the ability to select zip file download using the API
- Added the ability to save that preference to the database
- Added the ability to load the information from the database and based on the selection, process the patient's EOB records and add them as a text file (ndjson) or a zip file. The files can then be downloaded.

### Checklist

- [ ] Tests and linting pass

### Security
I added a validation step to make sure the sponsor owns the contract if they pass one to the API before processing the job.

AB2D-944
AB2D-945


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -42,6 +42,7 @@
 
 import static gov.cms.ab2d.api.util.Constants.GENERIC_FHIR_ERR_MSG;
 import static gov.cms.ab2d.api.util.SwaggerConstants.*;
+import static gov.cms.ab2d.common.service.JobServiceImpl.ZIPFORMAT;
```
dkrylovsb: Should we move this constant to the `JobService` interface? That way you could avoid referencing an implementation class directly.

common/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -22,5 +22,7 @@ databaseChangeLog:
       file: db/changelog/v001/update_attestation_date.sql
   - include:
       file: db/changelog/v001/add_concurrent_job_column.sql
+  - include:
```
dkrylovsb: I think this should be moved to the end of this file, i.e. be the last change set.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -52,12 +56,13 @@
...
-    private static final String ERROR_FILE_SUFFIX = "_error.ndjson";
     private static final int SLEEP_DURATION = 250;
-    private static final long ROLL_OVER_THRESHOLD = 200; // 200 MB
-    private static final int ONE_MEGA_BYTE = 1024 * 1024;
 
+    @Value("${job.file.rollover.ndjson:200}")
+    private long ndjsonRollOver;
+
+    @Value("${job.file.rollover.zip:200}")
+    private long zipRollOver = 200;
```
dkrylovsb: This initialization is probably redundant, since the property already declares the same default value.

common/src/main/resources/db/changelog/v001/ab2d-934-allow-for-zip-downloads.sql
```diff
@@ -0,0 +1,7 @@
+--liquibase formatted sql
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset lsharshar:ab2d-934-allow-for-zip-downloads failOnError:true
+ALTER TABLE job ADD COLUMN output_format VARCHAR(255);
+
+--rollback  ALTER TABLE output_format DROP COLUMN IF EXISTS mbi;
```
spathiyil: The `rollback` statement needs to be updated to drop column `output_format` from `job` table.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -191,32 +237,44 @@ private void deleteExistingDirectory(Path outputDirPath) {
...
         final Contract jobSpecificContract = job.getContract();
         if (jobSpecificContract != null && jobSpecificContract.getAttestedOn() != null) {
-            log.info("Job [{}] submitted for a specific attested contract [{}] ", jobSpecificContract.getContractNumber());
+            boolean ownsContract = attestedContracts.stream()
+                    .anyMatch(c -> jobSpecificContract.getContractNumber().equalsIgnoreCase(c.getContractNumber()));
+            if (!ownsContract) {
+                log.info("Job [{}] submitted for a specific attested contract [{}] that the sponsor [{}] does not own",
+                        job.getJobUuid(), jobSpecificContract.getContractNumber(), sponsor.getOrgName());
+            }
+            log.info("Job [{}] submitted for a specific attested contract [{}] ", job.getJobUuid(), jobSpecificContract.getContractNumber());
```
dkrylovsb: We are already guarding against this in `gov.cms.ab2d.common.service.JobServiceImpl#createJob(java.lang.String, java.lang.String, java.lang.String)`. Do we need this additional check?

common/src/main/resources/db/changelog/v001/ab2d-934-allow-for-zip-downloads.sql
lailasharshar: Whoops, copy and paste.

common/src/main/resources/db/changelog/db.changelog-master.yaml
lailasharshar: Product of a merge. I'll move it.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
lailasharshar: It's not expensive, why not? If the other class gets refactored, it might disappear and it's a pretty big security hole if PDPs get access to contracts that aren't theirs.

worker/src/main/java/gov/cms/ab2d/worker/processor/StreamHelperImpl.java
```diff
@@ -0,0 +1,216 @@
...
+    @Getter
+    private final String contractNumber;
+
+    // Directory where to put the files
+    @Getter
+    private final Path path;
+
+    // Total number of bytes written to the file
+    @Getter @Setter
+    private long totalBytesWritten = 0;
```
spathiyil: I wonder if this needs to be `volatile` since it will be accessed from multiple threads.

worker/src/main/java/gov/cms/ab2d/worker/processor/StreamHelperImpl.java
```diff
@@ -0,0 +1,216 @@
...
+
+    // Total bytes allowed in the file
+    @Getter
+    private final long totalBytesAllowed;
+
+    // The current output stream
+    @Getter @Setter
+    private OutputStream currentStream;
```
spathiyil: Would this also need to be `volatile` because it will be access in multiple threads..
And one of the threads could have closed the previous file and assigned a newly created file to this variable?

worker/src/test/java/gov/cms/ab2d/worker/processor/JobDataWriterTest.java
```diff
@@ -75,19 +75,19 @@ void addThreeEntriesThatCrossesMaxFileSize_shouldCreateMultipleFiles() {
...
     }
 
     @Test
     void addMultipleErrorEntries_createsOneErrorFile() {
-        cut.addErrorEntry(line);
-        cut.addErrorEntry(line);
-        cut.addErrorEntry(line);
+        cut.addData(line);
+        cut.addData(line);
+        cut.addError(poem);
```
spathiyil: I think all of the above may need to write to the error file.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobDataWriterTest.java
```diff
@@ -15,11 +15,12 @@
 
 class JobDataWriterTest {
```
spathiyil: I wonder if this test should be renamed to `TextStreamHelperTest`, since the production class is no longer named `JobDataWriterImpl`

worker/src/test/java/gov/cms/ab2d/worker/processor/JobPreProcessorIntegrationTest.java
```diff
@@ -101,7 +100,7 @@ private User createUser(Sponsor sponsor) {
         user.setUsername("Harry_Potter");
         user.setFirstName("Harry");
         user.setLastName("Potter");
-        user.setEmail("harry_potter@hogwarts.com");
+        user.setEmail("harry_potter@hogwarts.edu");
```
spathiyil: LOL. funny!

lailasharshar: I hate to call you out on such a egregious error.

worker/src/main/java/gov/cms/ab2d/worker/processor/TextStreamHelperImpl.java
```diff
@@ -0,0 +1,93 @@
...
+     * Write data to the correct file and iterate to the next file if it exceeds the limit
+     *
+     * @param data - the data to write
+     */
+    @Override
+    public void addData(byte[] data) {
+        tryLock(getDataFileLock());
+        if (data == null || data.length == 0) {
+            return;
+        }
```
dkrylovsb: If this condition is ever true, the lock will never release. This `if` needs to be moved inside the `try/finally`.

dkrylovsb: ...actually, it needs to be placed right before `tryLock` instead.

worker/src/main/java/gov/cms/ab2d/worker/processor/TextStreamHelperImpl.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+            if (getTotalBytesWritten() + data.length > getTotalBytesAllowed() && getTotalBytesWritten() > 0) {
+                getCurrentStream().close();
+                setCurrentStream(createStream());
+                setTotalBytesWritten(0);
+            }
+            getCurrentStream().write(data);
+            setTotalBytesWritten(getTotalBytesWritten() + data.length);
+        } catch (Exception ex) {
+            log.error("Unable to create file output stream for contract " + getContractNumber() + "[" + (getCounter() - 1) + "]", ex);
```
dkrylovsb: This is likely a fatal error condition. Let's re-throw.

worker/src/main/java/gov/cms/ab2d/worker/processor/TextStreamHelperImpl.java
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Close the stream clean up any empty files in the files created list
+     */
+    @Override
+    public void close() {
+        if (getCurrentStream() == null) {
+            return;
+        }
+        try {
+            getCurrentStream().close();
```
dkrylovsb: A short-cut: `IOUtils.closeQuietly`. Avoids both the `if` and `try`.

worker/src/test/java/gov/cms/ab2d/worker/processor/JobDataWriterTest.java
lailasharshar: I've created a new TextStreamHelperImplTest. This was just to make sure I didn't break any previous expectations. More regression testing. We could name it something else??

worker/src/main/java/gov/cms/ab2d/worker/processor/TextStreamHelperImpl.java
lailasharshar: Good catch

lailasharshar: I just moved it above. No reason to get the lock if there is no data.

worker/src/main/java/gov/cms/ab2d/worker/processor/TextStreamHelperImpl.java
```diff
@@ -0,0 +1,93 @@
...
+            int numFiles = getFilesCreated().size();
+            if (getFilesCreated().get(numFiles - 1).toFile().length() == 0) {
```
dkrylovsb: We need a index check here. In case BFD returns _no_ patients for a contract, list of files will be empty.

dkrylovsb: ...need a test to cover this scenario as well

worker/src/test/java/gov/cms/ab2d/worker/processor/JobDataWriterTest.java
spathiyil: okay. Did not realize that. Thanks.

worker/src/main/java/gov/cms/ab2d/worker/processor/TextStreamHelperImpl.java
lailasharshar: That appears to be deprecated

worker/src/main/java/gov/cms/ab2d/worker/processor/TextStreamHelperImpl.java
lailasharshar: Since we add the file to the list when we create the stream, the testing of the length of the last file is doing that. If no data exists, it's a list of one that has no data in it, but this also allows us to check for instances where the boundary between files occurs exactly at the end of the data to add.

worker/src/main/java/gov/cms/ab2d/worker/processor/ZipStreamHelperImpl.java
```diff
@@ -0,0 +1,210 @@
...
+     * the zip file and either writes to the current entry, a new entry or a new zip file.
+     *
+     * @param data - the data to write
+     */
+    @Override
+    public void addData(byte[] data) {
+        tryLock(getDataFileLock());
+        if (data == null || data.length == 0) {
+            return;
+        }
```
dkrylovsb: Same comment

worker/src/main/java/gov/cms/ab2d/worker/processor/ZipStreamHelperImpl.java
```diff
@@ -0,0 +1,211 @@
...
+            // (Either the open one of if that's too large, a new one)
+            if (currentPartByteStream.size() + data.length > totalExpandedBytesInEntryAllowed) {
+                if (fileFull(currentPartByteStream)) {
+                    resetZipFile();
+                }
+                addPartToFile();
+            }
+            currentPartByteStream.write(data);
+        } catch (Exception ex) {
+            log.error("Unable to create file output stream for contract " + getContractNumber() + "[" + (currentZipIteration - 1) + "]", ex);
```
dkrylovsb: Need to wrap & re-throw

worker/src/main/java/gov/cms/ab2d/worker/processor/ZipStreamHelperImpl.java
```diff
@@ -0,0 +1,211 @@
...
+        try {
+            if (currentPartByteStream.size() != 0) {
+                if (fileFull(currentPartByteStream)) {
+                    resetZipFile();
+                }
+                addPartToFile();
+            }
+            getCurrentStream().close();
+            int numFiles = getFilesCreated().size();
+            if (getFilesCreated().get(numFiles - 1).toFile().length() == 0) {
```
dkrylovsb: Index check here, too

worker/src/main/java/gov/cms/ab2d/worker/processor/ZipStreamHelperImpl.java
```diff
@@ -0,0 +1,211 @@
...
+                getFilesCreated().remove(numFiles - 1);
+            }
+        } catch (Exception ex) {
+            log.error("Unable to close output stream for contract " + getContractNumber() + "[" + currentZipIteration + "]", ex);
```
dkrylovsb: Also need to wrap & re-throw this one

worker/src/main/java/gov/cms/ab2d/worker/processor/StreamHelperImpl.java
```diff
@@ -0,0 +1,216 @@
...
+            this.suffix = suffix;
+        }
+        public String getSuffix() {
+            return suffix;
+        }
+    }
+
+    // Current file counter
+    @Getter
+    private int counter = 1;
```
spathiyil: I wonder if this ought to be volatile as well

worker/src/main/java/gov/cms/ab2d/worker/processor/StreamHelperImpl.java
```diff
@@ -0,0 +1,216 @@
...
+                final String errMsg1 = errMsg + " after waiting " + tryLockTimeout + " seconds.";
+                throw new RuntimeException(errMsg1);
+            }
+        } catch (InterruptedException e) {
+            throw new RuntimeException(errMsg);
+        }
+    }
+
+    /**
+     * Take an arraw of bytes and append it to a file
```
spathiyil: typo alert:   arraw -> array

common/src/main/java/gov/cms/ab2d/common/service/JobService.java
```diff
@@ -7,10 +7,11 @@
 import java.net.MalformedURLException;
 
 public interface JobService {
+    static final String ZIPFORMAT = "application/zip";
```
spathiyil: The `static` keyword here is causing the build to fail in CI.

PR #190, opened 2020-02-17
Feature/ab2d 1029 create contract adapter impl

spathiyil:
Create contract adapter implementation that can call out to the BFD `contract-to-bene` api

### Summary

Create contract adapter implementation that can call out to the BFD `contract-to-bene` api.
The API only returns the beneficiaries active on the contract for a specific month.
So we would need to make multiple calls to the API, one for each month in the current year from January till the current month.

### Checklist

- [x] Tests and linting pass

### Security
N/A


worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
```diff
@@ -0,0 +1,186 @@
...
+     * calls BFD api to get a bundle of resources from which the patientIDs are extracted.
+     * If there are multiple pages available, fetches all pages.
+     *
+     * @param contractNumber
+     * @param month
+     * @return a list of PatientIds
+     */
+    private List<String> getPatientIdsForMonth(String contractNumber, Integer month) {
+        Bundle bundle = getBundle(contractNumber, month);
+        final List<String> patientIDs = extractPatientIDs(bundle);
```
dkrylovsb: Should we use a `Set` to cover a scenario (albeit unlikely) BFD returns duplicates for a _given_ month?

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
```diff
@@ -0,0 +1,186 @@
...
+     */
+    private String extractPatientId(Patient patient) {
+        return patient.getIdentifier().stream()
+                .filter(identifier -> isBeneficiaryId(identifier))
+                .map(Identifier::getValue)
+                .findFirst().orElse(null);
+    }
+
+    private boolean isBeneficiaryId(Identifier identifier) {
+        return identifier.getSystem().toLowerCase().endsWith("bene_id");
```
dkrylovsb: It might be best to do a full comparison against the identifier's System value.

worker/src/test/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.when;
+
+
+@ExtendWith(MockitoExtension.class)
+class ContractAdapterTest {
```
dkrylovsb: We need to "beef up" this test significantly. Mock various responses from BFD and assert the results are as expected. Examples:

- A Patient is returned for month 1 & 2, verify its date ranges set correctly
- A Patient is returned for month 1, 3, but not 2 -- verify the date ranges have a gap as expected
- A Patient is returned multiple times for a single month -- ensure we handle this by filtering out dupes. 
- Simulate paging: first mock Bundle has a paging link that leads to the 2nd bundle: verify that Patients from both bundles are collected.

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
```diff
@@ -0,0 +1,186 @@
...
+     * @param bundle
+     * @return a list of patientIds
+     */
+    private List<String> extractPatientIDs(Bundle bundle) {
+        return bundle.getEntry().stream()
+                .map(Bundle.BundleEntryComponent::getResource)
+                .filter(resource -> resource.getResourceType() == ResourceType.Patient)
+                .map(resource -> (Patient) resource)
+                .map(patient -> extractPatientId(patient))
+                .filter(Objects::nonNull)
```
dkrylovsb: We also want to filter our empty strings.

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
spathiyil: Do you know what that value would be? 
I do not what that complete value for System would be.

I used the logic from here: 
https://github.com/CMSgov/ab2d/blob/master/optout/src/main/java/gov/cms/ab2d/optout/OptOutConverterServiceImpl.java#L124

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
spathiyil: Good point.  Will change it to use `Set` to avoid duplicates.

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
dkrylovsb: @spathiyil `https://bluebutton.cms.gov/resources/variables/bene_id`

spathiyil: oh, I see. I should've used `equals("BENE_ID")` instead of`.endsWith("bene_id")`?

dkrylovsb: `getSystem().equals("https://bluebutton.cms.gov/resources/variables/bene_id")`

spathiyil: got it. thanks

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterImpl.java
spathiyil: done

worker/src/test/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractAdapterTest.java
spathiyil: added additional tests

spathiyil: done

PR #193, opened 2020-02-18
Feature/ab2d 1033 load properties

adaykin:
Move several properties to the database so that they can be reloaded

Create a migration to create a table for storing properties and insert 3 that will be able to be reloaded. On application startup the properties will be read from the database and then set so that places throughout the application that access properties can use the conventional way of accessing them in Spring.

### Checklist

- [x] Tests and linting pass

### Security

NA


common/src/main/resources/db/changelog/v001/add_properties_table.sql
```diff
@@ -0,0 +1,16 @@
...
+    key VARCHAR(80) NOT NULL,
+    value VARCHAR(255) NOT NULL
+);
+
+ALTER TABLE properties ADD CONSTRAINT "uc_properties_key" UNIQUE (key);
+ALTER TABLE properties ADD CONSTRAINT "pk_properties" PRIMARY KEY (id);
+
+INSERT INTO properties (id, key, value) VALUES((select nextval('hibernate_sequence')), 'pcp.core.pool.size', 10);
+INSERT INTO properties (id, key, value) VALUES((select nextval('hibernate_sequence')), 'pcp.max.pool.size', 150);
+INSERT INTO properties (id, key, value) VALUES((select nextval('hibernate_sequence')), 'pcp.scaleToMax.time', 900);
```
spathiyil: Rollback?

worker/src/test/java/gov/cms/ab2d/worker/config/AutoScalingServiceTest.java
```diff
@@ -97,7 +97,7 @@ void autoScalingKicksInAndResizes() throws InterruptedException {
         // Last metric taken should always be MAX_POOL_SIZE
         assertThat(new ArrayDeque<>(metrics).getLast(), equalTo(MAX_POOL_SIZE));
         // Some intermediate sizes should be in between, at least 3.
-        assertThat(metrics.size(), greaterThanOrEqualTo(4));
+        assertThat(metrics.size(), greaterThanOrEqualTo(3));
```
dkrylovsb: The comment above is a bit confusing, I agree. 3 intermediate metrics plus one final is 4 total; so we should keep `4`. Let's update the comment instead.

common/src/main/resources/db/changelog/v001/add_properties_table.sql
adaykin: I can drop the table

spathiyil: okay. 
Am not sure, but it is possible you may need to drop the unique constraint `uc_properties_key` as well, if it does not get automatically deleted when you drop the table.

adaykin: I think it's only a foreign key reference that would prevent deletion

spathiyil: okay

worker/Dockerfile
```diff
@@ -1,4 +1,10 @@
...
+
 FROM openjdk:13
-WORKDIR /usr/src/ab2d-worker
-ADD target /usr/src/ab2d-worker
-CMD java -javaagent:/usr/src/ab2d-worker/newrelic/newrelic.jar -jar worker-*-SNAPSHOT.jar
\ No newline at end of file
+VOLUME /tmp
+ARG JAR_FILE=target/worker-*-SNAPSHOT.jar
+COPY ${JAR_FILE} app.jar
+ENTRYPOINT ["java","-Djava.security.egd=file:/dev/./urandom","-jar","/app.jar"]
```
dkrylovsb: I think this is left-overs from debugging CI; we can revert

PR #198, opened 2020-02-20
Feature/ab2d 1069 recreate coverage table to include additional columns

spathiyil:
Drop & Recreate the `Coverage` table so that it has `id` as the primary key and a new column `month`  needed as the `Contract-2-bene` BFD api returns contract data for each month.

### Summary

Recreate Coverage table so that it has id as the primary key and a new column `month`.
Create the corresponding `Coverage` entity.
And update the JPA associations from `@ManyToMany` to 2 `@OneToMany` relationships that uses the new `Coverage` entity

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


common/src/main/resources/db/changelog/v001/ab2d-1069-recreate-coverage-table.sql
```diff
@@ -0,0 +1,35 @@
...
+
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset spathiyil:ab2d-1069-recreate-coverage-table-with-id-and-month failOnError:true
+CREATE TABLE coverage
+(
+    id                  BIGINT              NOT NULL,
+    contract_id         BIGINT              NOT NULL,
+    beneficiary_id      BIGINT              NOT NULL,
+    month               INTEGER             NOT NULL
```
spathiyil: Turns out `month` is a keyword in SQL. 
I do not know if there is any danger in using the column name as `month`.
I could rename it to something like `active_month` unless there is a better name?

dkrylovsb: `part_d_month`?

spathiyil: done

PR #202, opened 2020-02-25
Feature/ab2d 1034 reload properties api v2

adaykin:
Adding in an API to allow for updates/retrieval of properties stored in the database.

Add API endpoints to the admin controller so that an admin user can retrieve a list of the properties to see what the value is. An API endpoint will also exist so that the admin user can update properties, which will be validated.

### Checklist

- [x] Tests and linting pass

### Security

NA


api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
```diff
@@ -43,7 +40,7 @@
     }
 
     // Most errors that are the fault of the client
```
dkrylovsb: You bet

common/src/main/java/gov/cms/ab2d/common/dto/PropertiesDTO.java
```diff
@@ -0,0 +1,17 @@
+package gov.cms.ab2d.common.dto;
+
+import lombok.Data;
+
+import javax.validation.constraints.NotNull;
+
+@Data
+public class PropertiesDTO {
+
+    private Long id;
```
dkrylovsb: Since DTO is what flows through the API to the end user, we don't need to expose internal primary key (`id`).

PR #204, opened 2020-02-25
Feature ab2d 1070 Add ability to cache contract-to-bene data

spathiyil:
Add ability to cache contract-to-bene data

### Summary

Add the ability to worker to be able to cache the patientId and months active in a contract when the call the BDF contract-to-bene API returns a large amount of data. When processing a job, the `ContractAdapterImpl` will look at the cache first to see if we have the data, before attempting to call the remote api. A property named `contract2bene.caching.threshold` is used to configure this.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
@@ -12,7 +12,7 @@ import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Optional;
 import java.util.Set;
 
@@ -37,7 +37,7 @@ public class BeneficiaryServiceImpl implements BeneficiaryService {
     @Override
     public Set<String> findPatientIdsInDb(Long contractId, int month) {
         var patientIds = coverageRepo.findActivePatientIds(contractId, month);
-        return new HashSet<>(patientIds);
+        return new LinkedHashSet<>(patientIds);
     }
 
 
--- a/worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
@@ -1,0 +1,86 @@
+package gov.cms.ab2d.worker.service;
+
+import gov.cms.ab2d.common.model.Beneficiary;
+import gov.cms.ab2d.common.model.Contract;
+import gov.cms.ab2d.common.model.Coverage;
+import gov.cms.ab2d.common.repository.BeneficiaryRepository;
+import gov.cms.ab2d.common.repository.ContractRepository;
+import gov.cms.ab2d.common.repository.CoverageRepository;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class BeneficiaryServiceImpl implements BeneficiaryService {
+
+    private final BeneficiaryRepository beneRepo;
+    private final ContractRepository contractRepo;
+    private final CoverageRepository coverageRepo;
+
+
+    /**
+     * Given a contractId and a month,
+     * search for bene information in the local db first.
+     *
+     * @param contractId
+     * @param month
+     * @return
+     */
+    @Override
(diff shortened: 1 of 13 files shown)


worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+     * search for bene information in the local db
+     *
+     * @param contractId
+     * @param month
+     * @return
+     */
+    @Override
+    public Set<String> findPatientIdsInDb(Long contractId, int month) {
+        var patientIds = coverageRepo.findActivePatientIds(contractId, month);
+        return new HashSet<>(patientIds);
```
dkrylovsb: Should we use `LinkedHashSet`? (To keep the same order in which the IDs were inserted into the cache)

spathiyil: done

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+     * search for bene information in the local db
+     *
+     * @param contractId
+     * @param month
+     * @return
+     */
+    @Override
+    public Set<String> findPatientIdsInDb(Long contractId, int month) {
+        var patientIds = coverageRepo.findActivePatientIds(contractId, month);
+        return new LinkedHashSet<>(patientIds);
```
adaykin: Since this is being ordered, should there be an order by statement in the SQL query?

adaykin: And is it possible for the JPA method to return a linkedhashset?

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.LinkedHashSet;
+import java.util.Optional;
+import java.util.Set;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
```
adaykin: Do you need @Transactional at the class level here? One of the methods that hits the DB has it, but I'm wondering if the query that reads needs it too?

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
spathiyil: I do not really think we need to do an order by since the records are written in a transaction.
I also did a quick search on having Spring data return a Set instead of a List by default and it does not seem like there is an easy/simple way. 
The above code to use the constructor is clear & simple and I am inclined to leave it as is.

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
spathiyil: I feel only the insert method needs to be in a transaction.
I do not see any value to adding transactions to the read method.

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
adaykin: Ok. I'm a little confused by why it's keeping order with the patients in the first place then with the linkedhashset since the db wouldn't return order in a predictable way?

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
adaykin: There's some reasons discussed here for using transactions with read e.g. optimizations, but may not make a huge difference here https://stackoverflow.com/a/13540484

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
spathiyil: Do I need to do an order by @dkrylovsb ?  
If yes, which column should I use? 
It does not seem like `contract_id` or `beneficiary_id` would make sense to do an `order by`.

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
spathiyil: Thanks for the link. Will study it.

worker/src/main/java/gov/cms/ab2d/worker/service/BeneficiaryServiceImpl.java
dkrylovsb: DB orders by primary key by default; this should work for us

adaykin: Ok, didn't realize it would order by the PK by default.

PR #207, opened 2020-02-26
Feature/ab2d 1072 trace calls

adaykin:
Adding tracing and tokens so that metrics for job processing within async methods can be uploaded to new relic.

Add a trace method that receives a token and expires it when the processing has finished. Each contract processed with be within a segment that has the contract number in the title so that jobs with multiple contracts can be timed properly. 

### Checklist

- [x] Tests and linting pass

### Security

NA


worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -362,7 +363,11 @@ private long getRollOverThreshold() {
                 }
 
                 // Add the thread to process the patient and start the thread
-                futureHandles.add(patientClaimsProcessor.process(patient, helper, contract.getAttestedOn()));
+                // See https://docs.newrelic.com/docs/agents/java-agent/async-instrumentation/java-agent-api-asynchronous-applications
+                // for more detail on tokens with async calls
+                final Token token = NewRelic.getAgent().getTransaction().getToken();
+
```
spathiyil: Looks like a token is being created for each patient record being processed in a job.
Have you considered using the same token for all the patients in the same job?

We could have a huge amount of patients in a job.... anywhere from 100K to million+.
Is there a limit to the number of tokens available within a job?

adaykin: Each call needs a token, otherwise the trace won't make much sense in the New Relic UI and would miss calls. If it does become an issue with performance we can look at changing it.

spathiyil: sounds good.

PR #208, opened 2020-02-26
Feature/ab2d 1041 avoid race condition

lailasharshar:
### Summary
If multiple jobs are submitted near the same time, currently, the worker will work on one job and when that one is done, will start work on the next one. I created a BlockingQueue that pulls from a constantly updating list of thread requests. 

Let's say there were 2 jobs submitted (with one contract each), first Job A, then Job B. Job A has 10 patients, Job B has 5 patients. 

You would have the following order for threads:
(Job A submitted) A, A, A, A, (Job B submitted) A, A, A, A, A, A, B, B, B, B, B

The new queue execution will look like this:
(Job A submitted) A, A, A, A, (Job B submitted) B, A, B, A, B, A, B, A, B, A, A

To implement this, I wrote a class ReadRobinBlockingQueue to replace the BlockingQueue in ThreadPoolExecutor. The new queue had to accept both the callable object and a contract number associated with it. The queue would maintain a list of futures for each contract. When the ThreadPoolExecutor tried to pop an item out of the queue to execute next, the queue could get the "next" contract. Instead of just the next item in a queue, it would pull the oldest thread for that contract's queue and it would be executed by the pool.

The way the queue is created is by ThreadPoolTaskExecutor which creates a ThreadPoolExecutor which in turn creates a BlockingQueue. Because these objects and many of their methods operating on the queue are private, it required rewriting any methods that directly changed the values of the queue and then, any other items that were changed in any of the methods changing the queue. It ended up being most of it, so a lot of the code for ThreadPoolExecutor and ThreadPoolTaskExecutor was simply replicated into new subclasses.

On the other end where places where the worker asked for the threads to be executed.  The new queue required both a contract number and Callable object to be executed or submitted, so a lot of the code for ThreadPoolTaskExecutor had to be rewritten to support that. The rewriting wasn't that hard. All I had to do is copy the necessary methods and edit any changes required by submitting a contract but it created a lot of code to test that wasn't mine and it failed coverage requirements. I first went to the Spring github page and pulling any tests they had. This ended up being a good way to make sure I was using the correct methods in child and parent classes and I didn't break anything. Unfortunately, it was still under the minimum coverage amount so I added additional tests to account for that.

Finally, I integrated it into the existing code and verified it was actually performing as desired.

### Checklist

- [ ] Tests and linting pass

### Security
No security impact


StewGoin:
Neat! I do have a question (not sure I'm conversant enough in the language to suss this out): with this queueing approach how do we avoid mixing up which requests belong to which jobs? Want to make sure we have a robot way to make sure we don't mix up data and can trace the requests/data through the job where that's possible (if that makes sense).

worker/pom.xml
```diff
@@ -97,6 +103,12 @@
             <artifactId>newrelic-api</artifactId>
             <version>${newrelic.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>17.0.0</version>
+            <scope>compile</scope>
```
adaykin: Nitpick, but you don't have to include compile since it's the default

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,358 @@
...
+            Iterator it = linkedList.iterator();
+            List<E> toRemove = new ArrayList<>();
+            while (it.hasNext()) {
+                E next = (E) it.next();
+                if (op.test(next)) {
+                    toRemove.add(next);
+                }
+            }
+            removeAll(toRemove);
+            toRemove.clear();
```
adaykin: This call to clear() may be unnecessary since it's not used the rest of the loop and gets newed up each time through.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,358 @@
...
+            if (linkedList.size() == 0) {
+                categoryQueues.remove(entry.getKey());
+                categories.remove(entry.getKey());
+            }
+        }
+    }
+
+    // New code to add items
+    public boolean add(String category, E e) {
+        log.debug("Adding " + category + " - " + e);
```
adaykin: You can use the {} syntax in logs here for variables.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
adaykin: Oh ok nvm, I see that you're overriding the clear() below.

worker/src/test/java/gov/cms/ab2d/worker/processor/stub/ProcessThread.java
```diff
@@ -0,0 +1,31 @@
...
+    private WorkerConfig workerConfig;
+
+    public Future<?> sleep(String contract) {
+        RoundRobinThreadPoolTaskExecutor taskExecutor = (RoundRobinThreadPoolTaskExecutor) workerConfig.patientProcessorThreadPool();
+        return (taskExecutor.submitWithCategory(contract, () -> call(contract)));
+    }
+
+    public String call(String contract) {
+        try {
+            //log.info(contract + " - " + Thread.currentThread().getId() + " Started");
```
adaykin: Put log back in?

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
lailasharshar: Right, good catch.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinThreadPoolExecutor.java
```diff
@@ -0,0 +1,765 @@
...
+    private final Condition termination = mainLock.newCondition();
+    private int largestPoolSize;
+    private long completedTaskCount;
+    private volatile ThreadFactory threadFactory;
+    private volatile RejectedExecutionHandler handler;
+    private volatile long keepAliveTime;
+    private volatile boolean allowCoreThreadTimeOut;
+    private volatile int corePoolSize;
+    private volatile int maximumPoolSize;
+    private static final RejectedExecutionHandler DEFAULT_HANDLER = new AbortPolicy();
```
adaykin: This variable isn't used, did you need to call a method on it somewhere?

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Predicate;
+
+@Slf4j
+public class RoundRobinBlockingQueue<E> implements BlockingQueue<E> {
+    // The list of categories
+    private List<String> categories = new ArrayList<>();
+    // The individual category queues
+    private Map<String, LinkedList<E>> categoryQueues = new HashMap<>();
```
dkrylovsb: You don't really need to maintain two separate Collections -- categories and queues -- and keep them in sync throughout the code. All you need is a "list of queues", i.e. `List<List>` or `List<Queue>` or something similar. This will definitely help with making this class more generic and unaware of its "surrounding context".

worker/pom.xml
```diff
@@ -97,8 +103,11 @@
             <artifactId>newrelic-api</artifactId>
             <version>${newrelic.version}</version>
         </dependency>
-
-
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>17.0.0</version>
+        </dependency>
```
dkrylovsb: Do we need this? This library is for IDE & static analysis tools support.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+    // The current category index
+    private int currentIndex = 0;
+    // Main lock guarding all access
+    private ReentrantLock lock;
+    // Not empty condition on the lock
+    private final Condition notEmpty;
+    // Total number in the queues
+    private long count = 0;
```
dkrylovsb: This `count` would be redundant, since you can just check the size of the queue list.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+
+    RoundRobinBlockingQueue() {
+        log.info("Created Round Robin Blocking Queue");
+        lock = new ReentrantLock();
+        notEmpty = lock.newCondition();
+    }
+
+    // New code to purge items
+    public void removeIfTrue(Predicate<E> op) {
```
dkrylovsb: This method is not a part of the `BlockingQueue` interface or any of its super-interfaces, which means we most likely don't need it. I know it is referenced from `RoundRobinThreadPoolExecutor`, but that class will likely have to go (see other comments).

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinThreadPoolExecutor.java
```diff
@@ -0,0 +1,765 @@
...
+            interruptIdleWorkers();
+    }
+
+    public int getMaximumPoolSize() {
+        return maximumPoolSize;
+    }
+
+    public void setKeepAliveTime(long time, TimeUnit unit) {
+        if (time < 0)
+            throw new IllegalArgumentException();
```
adaykin: I think it would be helpful to include a message in the exception here about time being less than 0 so that it's easy to see what went wrong.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+            while (it.hasNext()) {
+                E next = (E) it.next();
+                if (op.test(next)) {
+                    toRemove.add(next);
+                }
+            }
+            removeAll(toRemove);
+            if (linkedList.size() == 0) {
+                categoryQueues.remove(entry.getKey());
+                categories.remove(entry.getKey());
```
dkrylovsb: Having to keep these two in sync is one of the issues maintaining categories & queues separately creates.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+        try {
+            if (!categories.contains(category)) {
+                categories.add(category);
+            }
+            Queue<E> categoryQueue = categoryQueues.computeIfAbsent(category, k -> new LinkedList<>());
+            count += c.size();
+            // c.size might be 0. Don't signal not empty until we know it's non-zero
+            if (count > 0) {
+                notEmpty.signal();
+            }
```
dkrylovsb: So these two methods would be replaced with `addQueue(List list)` that adds another individual queue to the round-robin list.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+            return categoryQueue.addAll(c);
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public boolean add(E e) {
+        throw new UnsupportedOperationException("Cannot add an item directly on the queue");
+    }
```
dkrylovsb: 👍

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+        }
+    }
+
+    @Override
+    public boolean retainAll(@NotNull Collection<?> c) {
+        throw new UnsupportedOperationException("Complicated and probably not needed so unless it is, not implementing");
+    }
+
+    public int size() {
+        return (int) count;
```
dkrylovsb: I guess we'd need to sum up sizes of individual queues.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /**
+     * Get the next round robin Object
+     *
+     * @return
+     */
+    private E getNext() {
+        final ReentrantLock lock = this.lock;
+        lock.lock();
```
dkrylovsb: Wonder if we could just declare all methods that need locking `synchronized` for simplicity; the effect appears to be the same.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
```diff
@@ -0,0 +1,357 @@
...
+        try {
+            if (categoryQueues == null || categories.isEmpty()) {
+                return null;
+            }
+            if (currentIndex >= categories.size()) {
+                currentIndex = 0;
+            }
+            String currentContract = categories.get(currentIndex);
+            return categoryQueues.get(currentContract).peek();
```
dkrylovsb: We should only do `peek` on the current queue is it is not empty. Otherwise, `peek` returns null, and `peekNext` returns null, too, while there may be other queues in rotation that are not empty.

dkrylovsb: ...unless of course I have missed a spot that ensures we never have empty queues in rotation.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinThreadPoolExecutor.java
```diff
@@ -0,0 +1,765 @@
+package gov.cms.ab2d.worker.config;
```
dkrylovsb: This is a quite large and complex body of code that we are copying from Java's standard library, but we don't have to. We only need the custom BlockingQueue implementation above.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinThreadPoolTaskExecutor.java
```diff
@@ -0,0 +1,117 @@
+package gov.cms.ab2d.worker.config;
```
dkrylovsb: Same as above.

worker/src/main/java/gov/cms/ab2d/worker/config/WorkerConfig.java
```diff
@@ -64,8 +63,8 @@
     private int jobQueueCapacity;
 
     @Bean
-    public Executor patientProcessorThreadPool() {
-        final ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
+    public ThreadPoolTaskExecutor patientProcessorThreadPool() {
+        final ThreadPoolTaskExecutor taskExecutor = new RoundRobinThreadPoolTaskExecutor();
```
dkrylovsb: Here we'd instantiate the same `ThreadPoolTaskExecutor` class, _but_ with an instance of the custom `BlockingQueue`.

worker/src/main/java/gov/cms/ab2d/worker/config/RoundRobinBlockingQueue.java
lailasharshar: Yes, I test to see if the queue is empty when I remove an item. If it is, I remove that list and the "category", aka contract from the contact list.

            if (linkedList.size() == 0) {
                categoryQueues.remove(entry.getKey());
                categories.remove(entry.getKey());
            }

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -362,7 +362,7 @@ private long getRollOverThreshold() {
                 }
 
                 // Add the thread to process the patient and start the thread
-                futureHandles.add(patientClaimsProcessor.process(patient, helper, contract.getAttestedOn()));
+                futureHandles.add(patientClaimsProcessor.process(contractNumber, patient, helper, contract.getAttestedOn()));
```
dkrylovsb: May be use a `ThreadLocal` to pass contract number to RoundRobinBlockingQueue without having to modify method signatures or to bring in custom Executor implementations.

lailasharshar:
@StewGoin - The individual threads are passed this information. It knows what it needs to do. We're just changing the order in which it does it. It's like changing the order in which http requests get to a server. Each request is atomic and has its own parameters, it's just a matter of making sure an individual user does not dominate the server's resources. Does that answer your question?

lailasharshar:
Superceded by https://github.com/CMSgov/ab2d/pull/218

PR #210, opened 2020-02-27
Feature/ab2d 1035 update properties

adaykin:
Add a scheduler that will update the properties every 10 minutes by checking for updates in the database.

Add a scheduler that checks the database every 10 minutes. Refactor how the loading happens so that calls can be made to update the properties. 

### Checklist

- [x] Tests and linting pass

### Security

NA


worker/src/main/java/gov/cms/ab2d/worker/properties/PropertiesChangeDetection.java
```diff
@@ -0,0 +1,19 @@
...
+
+
+@Component
+class PropertiesChangeDetection {
+
+    @Autowired
+    private PropertiesInit propertiesInit;
+
+    // Every 10 minutes
+    @Scheduled(fixedDelay = 10 * 60 * 1000)
```
dkrylovsb: We definitely want property changes to take effect sooner. Let's do 20 seconds instead.

worker/src/main/java/gov/cms/ab2d/worker/properties/PropertiesInit.java
```diff
@@ -12,9 +12,18 @@
...
+
     public PropertiesInit(PropertiesService propertiesService, ConfigurableEnvironment configurableEnvironment) {
-        log.info("Adding properties");
+        this.propertiesService = propertiesService;
+        this.configurableEnvironment = configurableEnvironment;
+    }
+
+    // Load all of the properties from the database and insert/overwrite
+    public void updatePropertiesFromDatabase() {
+        log.info("Updating properties");
```
dkrylovsb: Let's change to `debug` or `trace` to avoid excessive logging in the environments.

worker/src/main/java/gov/cms/ab2d/worker/properties/PropertiesInit.java
```diff
@@ -12,9 +12,18 @@
...
         final Map<String, Object> properties = propertiesService.getAllProperties().stream()
                 .collect(Collectors.toMap(Properties::getKey, Properties::getValue));
         configurableEnvironment.getPropertySources().addLast(new MapPropertySource("application", properties));
```
dkrylovsb: The problem is that even though the properties are reloaded and added to the `configurableEnvironment`, the beans have already been wired. Property changes won't be reflected in beans and thus will have no effect. 

Take a look at the last 3 paragraphs in https://jira.cms.gov/browse/AB2D-1035. Each property change might need a small amount of handling logic.

worker/src/test/java/gov/cms/ab2d/worker/properties/PropertiesChangeDetectionTest.java
```diff
@@ -0,0 +1,58 @@
...
+    private PropertiesRepository propertiesRepository;
+
+    @Autowired
+    private ConfigurableEnvironment configurableEnvironment;
+
+    @Container
+    private static final PostgreSQLContainer postgreSQLContainer = new AB2DPostgresqlContainer();
+
+    @Test
+    public void testChangeInProperties() {
```
dkrylovsb: This test case also needs to verify the property changes got reflected in the corresponding beans (again, see AB2D-1035)

worker/src/main/java/gov/cms/ab2d/worker/config/AutoScalingServiceImpl.java
```diff
@@ -111,6 +129,16 @@ private void scaleBackToMin() {
...
     }
 
+    @Override
+    public void onApplicationEvent(PropertiesChangedEvent propertiesChangedEvent) {
+        corePoolSize = Integer.parseInt(propertiesChangedEvent.getPropertiesMap().get(PCP_CORE_POOL_SIZE).toString());
+        maxPoolSize = Integer.parseInt(propertiesChangedEvent.getPropertiesMap().get(PCP_MAX_POOL_SIZE).toString());
+        scaleToMaxTime = Double.parseDouble(propertiesChangedEvent.getPropertiesMap().get(PCP_SCALE_TO_MAX_TIME).toString());
+
+        this.executor.setCorePoolSize(corePoolSize);
+        this.executor.setMaxPoolSize(maxPoolSize);
```
dkrylovsb: No need to modify max pool size on executor, since the Auto Scaling service dynamically manages that value.

PR #211, opened 2020-02-27
ab2d 1071 create admin api to clear coverage cache

spathiyil:
Create admin api to clear coverage cache

### Summary

Create admin api to clear coverage cache

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


api/src/main/java/gov/cms/ab2d/api/controller/AdminAPI.java
```diff
@@ -92,4 +106,13 @@
     public ResponseEntity<List<PropertiesDTO>> updateProperties(@RequestBody List<PropertiesDTO> propertiesDTOs) {
         return new ResponseEntity<>(propertiesService.updateProperties(propertiesDTOs), null, HttpStatus.OK);
     }
+
+
+    @PostMapping("/coverage/clearCache")
```
adaykin: I think this should be DeleteMapping, I can't see anything being created...you can then return an HTTP status of deleted.

spathiyil: I thought about that at first. 
But one is not allowed to use a `@RequestBody` with Http DELETE verb.

adaykin: Right, you would use the HTTP parameters for the variables, there's examples in BulkDataAccessAPI controller that use {variable} in the URL.

common/src/main/java/gov/cms/ab2d/common/repository/CoverageRepository.java
```diff
@@ -22,4 +23,28 @@
 
     Optional<Coverage> findByContractAndBeneficiaryAndPartDMonth(Contract contract, Beneficiary bene, int month);
 
+
+
+    @Modifying
+    @Query(" DELETE FROM Coverage c  ")
+    int deleteAllInBulk();
```
dkrylovsb: `deleteAll`?

spathiyil: This is the signature for delete all 
```
   void deleteAll();
```

It all does not return the number of rows deleted. I was logging the number of rows deleted for the other calls, so opted to be consistent with that.  But I could change it to the deleteAll() method for this case.

common/src/main/java/gov/cms/ab2d/common/repository/CoverageRepository.java
```diff
@@ -22,4 +23,28 @@
...
+
+    @Modifying
+    @Query(" DELETE FROM Coverage c  " +
+            " WHERE c.partDMonth = :month  ")
+    int deleteInBulk(int month);
+
+    @Modifying
+    @Query(" DELETE FROM Coverage c  " +
+            " WHERE c.contract.id = :contractId  ")
+    int deleteInBulk(Long contractId);
```
dkrylovsb: If I have an integer, say, `5`, then `deleteInBulk(5)` would _appear_ to fit either of these invocation, which might be confusing a bit. `deleteByMonth` and `deleteByContractId` would fix this.

common/src/main/java/gov/cms/ab2d/common/repository/CoverageRepository.java
```diff
@@ -22,4 +23,28 @@
...
+
+
+    @Modifying
+    @Query(" DELETE FROM Coverage c  " +
+            " WHERE c.contract.id = :contractId  " +
+            "   AND c.partDMonth = :month  ")
+    int deleteInBulk(Long contractId, int month);
```
dkrylovsb: ...following up on the above, `deleteByMonthAndContract` would settle this interface.

spathiyil: done

common/src/main/java/gov/cms/ab2d/common/repository/CoverageRepository.java
spathiyil: done

common/src/main/java/gov/cms/ab2d/common/repository/CoverageRepository.java
spathiyil: done - changed to use `deleteAll()`

PR #216, opened 2020-02-28
Feature/ab2d 1058 switch

adaykin:
Add an API call to turn on/off maintenance mode.

Add a properties value that determines whether or not we are in maintenance mode. This property is stored in the database and is checked by the API when a request comes in to create a job. If we are in maintenance mode, no new jobs can be created, but users can still download/delete/get the status of jobs.

### Checklist

- [x] Tests and linting pass

### Security

NA


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -112,6 +118,12 @@
         return returnStatusForJobCreation(job);
     }
 
+    private void checkIfInMaintenanceMode() {
+        if (propertiesService.isInMaintenanceMode()) {
+            throw new InMaintenanceModeException("The system is currently in maintenance mode. Please try the request again later.");
```
spathiyil: Would the `ErrorHandler` automatically handle this new `InMaintenanceModeException` ?

adaykin: Yes since it extends RuntimeException, which is a subclass of Exception, and then matches the line `@ExceptionHandler(Exception.class)`

PR #218, opened 2020-03-02
AB2D-1041: fixed the race condition by introducing RoundRobinBlockingQueue

dkrylovsb:
Fixed the race condition by introducing RoundRobinBlockingQueue.

### Summary

A joint effort by @lailasharshar and @dkrylovsb. A custom implementation of `BlockingQueue` made to fit the needs of AB2D. Because we have a single `ThreadPoolExecutor` that handles all the tasks across jobs, we've had a bit of a race condition for a while, which was being mitigated by throttling to some extent. This `BlockingQueue` implementation helps to solve that by introducing internal queues scoped to each particular job (or a PDP contract to be precise), and serving tasks to the surrounding `ThreadPoolExecutor` in round-robin fashion to ensure fair resource allocation.

### Checklist

- [x] Tests and linting pass

### Security
NA


worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -363,11 +364,17 @@ private long getRollOverThreshold() {
...
                 // See https://docs.newrelic.com/docs/agents/java-agent/async-instrumentation/java-agent-api-asynchronous-applications
                 // for more detail on tokens with async calls
                 final Token token = NewRelic.getAgent().getTransaction().getToken();
 
+                // Using a ThreadLocal to communicate contract number to RoundRobinBlockingQueue
+                // could be viewed as a hack by many; but on the other hand it saves us from writing
+                // tons of extra code.
+                RoundRobinBlockingQueue.CATEGORY_HOLDER.set(progressTracker.getJobUuid());
                 futureHandles.add(patientClaimsProcessor.process(patient, helper, contract.getAttestedOn(), token));
+                RoundRobinBlockingQueue.CATEGORY_HOLDER.remove();
```
adaykin: Should remove be in a finally block?

PR #221, opened 2020-03-02
ab2d 1110 Add a toggle to turn off zip support

spathiyil:
Add a toggle to turn off zip support

### Summary

Add a toggle to turn off zip support.
It looks for a toggle named `ZipSupportOn` in the `properties` table with the value `true`

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
@@ -3,11 +3,20 @@ package gov.cms.ab2d.api.controller;
 import com.google.gson.Gson;
 import gov.cms.ab2d.api.config.SwaggerConfig;
 import gov.cms.ab2d.api.util.SwaggerConstants;
+import gov.cms.ab2d.common.model.Job;
 import gov.cms.ab2d.common.service.InvalidUserInputException;
 import gov.cms.ab2d.common.service.JobService;
-import gov.cms.ab2d.common.model.Job;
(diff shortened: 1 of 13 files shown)


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -137,10 +146,16 @@ private void checkResourceTypesAndOutputFormat(String resourceTypes, String outp
...
         }
+
         if (outputFormat != null && !ALLOWABLE_OUTPUT_FORMAT_SET.contains(outputFormat)) {
             log.error("Received _outputFormat {}, which is not valid", outputFormat);
             throw new InvalidUserInputException("An _outputFormat of " + outputFormat + " is not valid");
         }
+
+        final boolean zipSupportOn = propertiesService.isToggleOn(ZIP_SUPPORT_ON);
+        if (!zipSupportOn && ZIPFORMAT.equalsIgnoreCase(outputFormat)) {
+            throw new InvalidUserInputException("An _outputFormat of " + outputFormat + " is not valid");
```
dkrylovsb: This duplicates line 152 earlier. Should we at least extract the error message into a variable, or merge the two `if`s together with an `||`?

spathiyil: I'll extract the error message into a common variable momentarily.

spathiyil: done

PR #224, opened 2020-03-04
Feature/ab2d 1111 code climate

adaykin:
Avoid deleting the common/target/site directory as part of the build process

Avoid deleting the common/target/site directory as part of the build process. This is to preserve the jacoco test coverage results, since a mvn clean package is being run in the docker-compose build, which is run by the e2e tests.

### Checklist

- [x] Tests and linting pass

### Security

NA


common/pom.xml
```diff
@@ -90,6 +90,29 @@
                 </executions>
             </plugin>
 
+            <plugin>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>3.0.2</version>
+                <executions>
+                    <execution>
+                        <id>copy-resources</id>
+                        <phase>package</phase>
```
dkrylovsb: @adaykin Take a look at the CI output: this does not seem to have worked, likely due to the wrong `phase` value?
image

PR #228, opened 2020-03-06
Feature AB2D 1109 multiple opt out files

spathiyil:
Add the ability to process multiple opt out files from S3

### Summary

Add the ability to process multiple opt out files from S3

### Checklist

- [x] Tests and linting pass

### Security
N/S

N/A

### DIFF
--- a/optout/src/main/java/gov/cms/ab2d/optout/OptOutProcessorImpl.java
+++ b/optout/src/main/java/gov/cms/ab2d/optout/OptOutProcessorImpl.java
@@ -11,8 +11,6 @@ import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.util.List;
 
 @Slf4j
@@ -29,11 +27,11 @@ public class OptOutProcessorImpl implements OptOutProcessor {
     public void process() {
 
         final List<String> filenames = s3Gateway.listOptOutFiles();
-        filenames.stream().forEach(fileName -> fetchOptOutFile(fileName));
+        filenames.stream().forEach(this::fetchAndProcessOptOutFile);
 
     }
 
-    private void fetchOptOutFile(final String filename) {
+    private void fetchAndProcessOptOutFile(final String filename) {
         try (var inputStreamReader = s3Gateway.getOptOutFile(filename);
              var bufferedReader = new BufferedReader(inputStreamReader)) {
 
@@ -42,9 +40,8 @@ public class OptOutProcessorImpl implements OptOutProcessor {
             importOptOutRecords(bufferedReader);
 
             log.info("[{}] - import completed successfully", filename);
-        } catch (IOException e) {
-            log.error("[{}] - import FAILED", filename);
-            throw new UncheckedIOException(e);
+        } catch (Exception e) {
+            log.error("[{}] - import FAILED ", filename, e);
         }
     }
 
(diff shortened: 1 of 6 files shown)


optout/src/main/java/gov/cms/ab2d/optout/OptOutProcessorImpl.java
```diff
@@ -27,16 +27,23 @@
...
 
-        try (var inputStreamReader = s3Gateway.getOptOutFile();
-             var bufferedReader = new BufferedReader(inputStreamReader)
-        ) {
+        final List<String> filenames = s3Gateway.listOptOutFiles();
+        filenames.stream().forEach(fileName -> fetchOptOutFile(fileName));
+
+    }
+
+    private void fetchOptOutFile(final String filename) {
```
dkrylovsb: Since this method does more than fetching files, we should probably rename to `fetchAndProcessOptOutFile`

optout/src/main/java/gov/cms/ab2d/optout/OptOutProcessorImpl.java
```diff
@@ -27,16 +27,23 @@
...
+            log.info("importing [{}]", filename);
+
             importOptOutRecords(bufferedReader);
 
-            log.info("import opt-out data - DONE.");
+            log.info("[{}] - import completed successfully", filename);
         } catch (IOException e) {
-            log.info("import opt-out data - FAILED.");
+            log.error("[{}] - import FAILED", filename);
             throw new UncheckedIOException(e);
```
dkrylovsb: Now that we have multiple files, a failure of a single file should not cancel the entire import process: other files should still be attempted to process. I guess we can just log the stack trace and remove the `throw`

optout/src/main/java/gov/cms/ab2d/optout/gateway/S3GatewayImpl.java
```diff
@@ -62,10 +93,10 @@ public InputStreamReader getOptOutFile() {
...
 
 
-    private void validateFileName() {
-        if (!FILENAME_PATTERN.matcher(s3Filename).matches()) {
+    private void validateFileName(String fileName) {
+        if (!FILENAME_PATTERN.matcher(fileName).matches()) {
             final String errMsg = "Filename is invalid ";
-            log.error("{} : {} ", errMsg, s3Filename);
+            log.error("{} : {} ", errMsg, fileName);
             throw new RuntimeException(errMsg);
```
dkrylovsb: Similar to the above, let's remove the `throw`, so that we could move on to the next file.

spathiyil: hmm... I think this might still be needed here. 
I could change the other catch to `Exception` from `IOException` to handle both these cases....

dkrylovsb: That works too

optout/src/main/java/gov/cms/ab2d/optout/OptOutProcessorImpl.java
spathiyil: done

optout/src/main/java/gov/cms/ab2d/optout/OptOutProcessorImpl.java
spathiyil: done

optout/src/main/java/gov/cms/ab2d/optout/gateway/S3GatewayImpl.java
spathiyil: done

PR #229, opened 2020-03-06
Feature/ab2d 1146 status file response

adaykin:
Adding file length and checksums to the Bulk Status API call when a 200 is returned.

Create a migration so that checksum and file length columns are added to the database. Set the checksum and file length from the worker when the job is completed so that when a user makes a Bulk Status call that has a 200 status, the checksum and file length will be displayed to the user.

### Checklist

- [x] Tests and linting pass

### Security

NA


api/pom.xml
```diff
@@ -97,13 +97,19 @@
...
             <version>5.10.0</version>
             <scope>provided</scope>
             <type>zip</type>
-	</dependency>
+	    </dependency>
+
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk15on</artifactId>
+            <version>${bounceycastle.version}</version>
```
dkrylovsb: typo alert

api/src/main/java/gov/cms/ab2d/api/controller/JobCompletedResponse.java
```diff
@@ -58,30 +65,40 @@ public void setError(List<Output> error) {
...
+            this.extension = valueOutputs;
         }
+    }
 
-        public String getType() {
-            return type;
-        }
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @Getter
+    public static final class ValueOutput {
```
dkrylovsb: `ValueOutput` does not clearly communicate the purpose of this class. How about renaming to `FileMetadata` ?

common/src/main/java/gov/cms/ab2d/common/model/JobOutput.java
```diff
@@ -39,4 +39,10 @@
 
     @NotNull
     private Boolean downloaded = false;
+
+    @NotNull
```
dkrylovsb: Why not just store the checksum in hex right away?

common/src/main/resources/db/changelog/v001/add_checksum_length_job_output.sql
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted sql
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset adaykin:add_checksum_length_job_output failOnError:true
+ALTER TABLE job_output ADD COLUMN checksum BYTEA NOT NULL;
```
dkrylovsb: ...that would help to avoid having a "byte" table column.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -186,6 +188,20 @@ private String verifyJsonFromStatusResponse(HttpResponse<String> statusResponse,
...
+        Assert.assertEquals("https://ab2d.cms.gov/checksum", checkSumUrl);
+        String checkSum = checkSumObject.getString("valueString");
+        Assert.assertEquals("sha256:", checkSum.substring(0, 7));
+        Assert.assertEquals(checkSum.length(), 71);
+
+        JSONObject lengthObject = extension.getJSONObject(1);
+        String lengthUrl = lengthObject.getString("url");
+        Assert.assertEquals("https://ab2d.cms.gov/file_length", lengthUrl);
+        long length = lengthObject.getLong("valueDecimal");
+        Assert.assertThat(length, greaterThan(0L));
```
dkrylovsb: Can we take this a step further? Let's have the e2e test actually download the files and _verify_ that the checkum & length match.

pom.xml
```diff
@@ -36,6 +36,7 @@
         <jacoco.version>0.8.5</jacoco.version>
         <checkstyle.config>${project.root}/src/main/resources/checkstyle.xml</checkstyle.config>
         <logback-encoder.version>6.3</logback-encoder.version>
+        <bounceycastle.version>1.64</bounceycastle.version>
```
dkrylovsb: typo alert

worker/pom.xml
```diff
@@ -98,6 +98,11 @@
             <version>${newrelic.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk15on</artifactId>
+            <version>${bounceycastle.version}</version>
+        </dependency>
```
dkrylovsb: Also, we _might_ have Apache Coded already on the classpath, in which case you could just use `org.apache.commons.codec.binary.Hex` for Hex-ing

worker/src/main/java/gov/cms/ab2d/worker/service/FileServiceImpl.java
```diff
@@ -1,8 +1,11 @@
 package gov.cms.ab2d.worker.service;
 
 import lombok.extern.slf4j.Slf4j;
+import org.bouncycastle.jcajce.provider.digest.SHA256;
```
dkrylovsb: If we remove bouncy castle, `org.apache.commons.codec.digest.DigestUtils` can be used instead.

worker/src/main/java/gov/cms/ab2d/worker/service/FileServiceImpl.java
```diff
@@ -34,4 +37,15 @@ public Path createDirectory(Path outputDir) {
         }
         return outputDirectory;
     }
+
+    @Override
+    public byte[] generateChecksum(File file) {
+        try (FileInputStream fileInputStream = new FileInputStream(file)) {
+            byte[] bytes = new SHA256.Digest().digest(fileInputStream.readAllBytes());
```
dkrylovsb: We should definitely try and _stream_ the file into the digest function instead. `readAllBytes` will read the entire file into memory and our files are big.

common/src/main/java/gov/cms/ab2d/common/model/JobOutput.java
adaykin: I was looking at what DPC did in their app, they kept it in the byte[] until displaying it to the user, I wasn't sure if there was a reason they were waiting to convert to hex.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
adaykin: Yeah I can return the data from this method and then verify it in the later step that downloads the file.

PR #233, opened 2020-03-09
Features/ab2d 1144 add _since to query

lailasharshar:
### Summary

Added the ability to pass _since to the export requests. It will check if the _since parameter is valid and then pass it to the BFD call. The request saves the since in the job table and once the job is run, it retrieves that value and passes it to BFD.

### Checklist

- [ ] Tests and linting pass

### Security


bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClientImpl.java
```diff
@@ -88,15 +93,38 @@ public BFDClientImpl(IGenericClient bfdFhirRestClient) {
...
-                .and(excludeSAMHSA)
-                .count(pageSize)
-                .returnBundle(Bundle.class)
-                .encodedJson()
-                .execute();
+                .and(excludeSAMHSA);
+        if (sinceTime != null) {
+            Date d = Date.from(sinceTime.toInstant());
+            query.lastUpdated(new DateRangeParam(d, null));
+        }
```
dkrylovsb: This `if` block breaks the method chaining that was flowing so nicely :) You can move it to the very beginning of the method, set a `lastUpdated` variable or something, and then just add `.lastUpdated(lastUpdated` to the chain (it accepts nulls).

worker/src/test/java/gov/cms/ab2d/worker/adapter/bluebutton/FilterByUpdateDate.java
```diff
@@ -0,0 +1,73 @@
...
+            assertEquals(listOne.get(i), listTwo.get(i));
+        }
+    }
+
+    private boolean updatedTimeAfter(Bundle.BundleEntryComponent component, OffsetDateTime earliest) {
+        if (component == null || component.getResource() == null || component.getResource().getMeta() == null
+                || component.getResource().getMeta().getLastUpdated() == null) {
+            return false;
+        }
+        System.out.println(component.getResource().getMeta().getLastUpdated());
```
dkrylovsb: We have to watch these `println` statements to avoid hitting Travis CI limits as we did last time.

dkrylovsb:
Item 7 from the Implementation Guide might have been missed:

> When "Get Job Status" endpoint returns 200, it populates transactionTime with the job's completion timestamp. This is incorrect. That value should be set to the job's creation time.

worker/src/test/java/gov/cms/ab2d/worker/adapter/bluebutton/FilterByUpdateDate.java
lailasharshar: My old school debugging is showing.

PR #234, opened 2020-03-10
AB2D-1168 Refine opt-out process to handle duplicate records

spathiyil:
Refine opt-out process to handle duplicate records

### Summary

Refine opt-out process to handle duplicate records. 
Store the filename in the opt_out table.
When processing `opt-out` record from a file, select for existing record based on CCW_ID and HICN. If found, if the `effectiveDate` is different, update the record along with the new `filename`
else skip it.

### Checklist

- [x] Tests and linting pass

### Security
N/A

N/A


optout/src/main/java/gov/cms/ab2d/optout/OptOutProcessorImpl.java
```diff
@@ -60,13 +61,13 @@ private void importOptOutRecords(BufferedReader bufferedReader) {
...
-                if (optOuts.size() > 1) {
-                    log.info("Multiple({}) Patients for HICN {}", optOuts.size(), optOuts.get(0).getHicn());
+                List<OptOut> newOptOuts = optOutConverterService.convert(line);
+                if (newOptOuts.size() > 1) {
+                    log.info("Multiple({}) Patients for HICN {}", newOptOuts.size(), newOptOuts.get(0).getHicn());
                 }
 
-                optOuts.forEach(o -> optOutRepository.save(o));
-                insertedRowCount = insertedRowCount + optOuts.size();
+                newOptOuts.forEach(newOptOut -> upsertOptOutRecord(newOptOut, filename));
```
dkrylovsb: typo alert

spathiyil: updated method name

PR #235, opened 2020-03-10
Feature/ab2d 1147 api switch

adaykin:
Adding an API switch to automatically go into maintenance mode when BFD goes down.

Add a cron job that will run on a configurable interval, which will look at whether or not the BFD API is up or down. If the API is up or down a configurable amount of times, flip the maintenance mode switch in the database accordingly.

### Checklist

- [x] Tests and linting pass

### Security

NA


worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
```diff
@@ -0,0 +1,64 @@
...
+
+    // Nothing else should be calling this component except for the scheduler, so keep
+    // state here
+    private int consecutiveSuccesses = 0;
+
+    private int consecutiveFailures = 0;
+
+    void checkBFDHealth() {
+
+        CapabilityStatement capabilityStatement = bfdClient.capabilityStatement();
```
dkrylovsb: This line can throw an exception I think, which should be caught and treated as a failure.

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
```diff
@@ -0,0 +1,64 @@
...
+        if (capabilityStatement == null || capabilityStatement.getStatus() != Enumerations.PublicationStatus.ACTIVE) {
+            consecutiveFailures++;
+            consecutiveSuccesses = 0;
```
dkrylovsb: Let's add a log statement here saying `N consequent failures to connect to BFD`.

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
```diff
@@ -0,0 +1,64 @@
...
+        } else {
+            consecutiveSuccesses++;
```
dkrylovsb: Let's add a log statement here saying `N consequent successes to connect to BFD`.

dkrylovsb: The problem with this logic is that it will keep counting the successes even when *not* in maintenance made and then constantly updating the property to `false` even though it is already `false`.

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
```diff
@@ -0,0 +1,64 @@
...
+    // state here
+    private int consecutiveSuccesses = 0;
+
+    private int consecutiveFailures = 0;
+
+    void checkBFDHealth() {
+
+        CapabilityStatement capabilityStatement = bfdClient.capabilityStatement();
+        if (capabilityStatement == null || capabilityStatement.getStatus() != Enumerations.PublicationStatus.ACTIVE) {
+            consecutiveFailures++;
```
dkrylovsb: Same as the comment below. If BFD is down, this code will keep on counting failures and constantly updating the property table without any effect (from `true` to `true`).

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Value("${bfd.health.check.consecutive.failures}")
+    private int consecutiveFailuresToTakeDown;
+
+    // Nothing else should be calling this component except for the scheduler, so keep
+    // state here
+    private int consecutiveSuccesses = 0;
+
+    private int consecutiveFailures = 0;
+
```
dkrylovsb: To address the issues mentioned below, I think we need another instance variable such as `private Status bfdStatus` where `Status` would be a private `enum` with `UP` and `DOWN` values. This variable would hold the currently known status of BFD.

The code below would count failures _only_ when the status is UP and vice versa.

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
```diff
@@ -0,0 +1,64 @@
...
+            PropertiesDTO propertiesDTO = new PropertiesDTO();
+            propertiesDTO.setKey(MAINTENANCE_MODE);
+            propertiesDTO.setValue("false");
+            propertiesService.updateProperties(List.of(propertiesDTO));
+        } else if (consecutiveFailures == consecutiveFailuresToTakeDown) {
+            consecutiveFailures = 0;
+            PropertiesDTO propertiesDTO = new PropertiesDTO();
+            propertiesDTO.setKey(MAINTENANCE_MODE);
+            propertiesDTO.setValue("true");
+            propertiesService.updateProperties(List.of(propertiesDTO));
```
dkrylovsb: Need log statements indicating we are flipping the switch.

worker/src/main/resources/application.properties
```diff
@@ -54,3 +54,7 @@ health.urlsToCheck=http://www.google.com,http://www.facebook.com
 
 spring.autoconfigure.exclude[0]=org.springframework.boot.autoconfigure.security.servlet.SecurityAutoConfiguration
 
+## -----------------------------------------------------------------------------------------------------  BFD Health Check
+bfd.health.check.schedule=*/15 * * * * ?
+bfd.health.check.consecutive.successes=3
+bfd.health.check.consecutive.failures=3
```
dkrylovsb: `3` _might_ be too aggressive, should we do `5`?

worker/src/test/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheckTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import java.io.IOException;
+import java.util.List;
+
+import static gov.cms.ab2d.common.util.Constants.MAINTENANCE_MODE;
+
+@SpringBootTest(classes = SpringBootApp.class, properties = "bfd.serverBaseUrl=http://localhost:8083/v1/fhir/")
+@Testcontainers
+public class BFDHealthCheckTest {
```
dkrylovsb: Very nice

PR #236, opened 2020-03-11
Check the X-Forwarded-Proto header to determine if the status & file names should return https

lailasharshar:
### Summary
The status endpoint returns URLs (for both status checking and downloads) as http, not https. There are two places in BulkDataAccessAPI where we generate URLs that go back to the end user. In those places, we need to check if the Request contains X-Forwarded-Proto HTTP header set to https. If it does, replace http with https in the generated URLs.

Also, when we create Jobs we use ServletUriComponentsBuilder.fromCurrentRequest().toUriString()` which later comes back as part of the Status 200 body. We have to replace http with https as described above as well.

Updated BulkDataAccessAPIIntegrationTests to have a test case that passes that header and verifies https.

### Checklist

- [ ] Tests and linting pass

### Security
Adds https instead of http in places.


api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -189,15 +205,23 @@ private void logSuccessfulJobCreation(Job job) {
...
                 HttpStatus.ACCEPTED);
     }
 
+    String getUrl(String ending) {
+        String url = ServletUriComponentsBuilder.fromCurrentRequestUri().replacePath(ending).toUriString();
+        url = url.replace(":80/", "/");
+        if (!shouldReplaceWithHttps()) {
+            return url;
+        }
+        return url.replace("http://", "https://");
```
dkrylovsb: These lines duplicate those in `getCurrentUrl`; let's refactor into a common method.

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -131,6 +131,30 @@ public void testBasicPatientExport() throws Exception {
         Assert.assertEquals(job.getUser(), userRepository.findByUsername(TEST_USER));
     }
 
+    @Test
+    public void testBasicPatientExportWithHttps() throws Exception {
+        ResultActions resultActions = this.mockMvc.perform(
+                get(API_PREFIX + FHIR_PREFIX + PATIENT_EXPORT_PATH).contentType(MediaType.APPLICATION_JSON)
+                        .header("Authorization", "Bearer " + token)
+                        .header("X-Forwarded-Proto", "https"))
+                .andDo(print());
```
dkrylovsb: Let's just make sure this does not print too much to console

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -124,14 +125,29 @@
...
+        Job job = jobService.createJob(resourceTypes, getCurrentUrl(), null, outputFormat, since);
 
         logSuccessfulJobCreation(job);
 
         return returnStatusForJobCreation(job);
     }
 
+    private String getCurrentUrl() {
+        String request = ServletUriComponentsBuilder.fromCurrentRequest().toUriString();
+        request = request.replace(":80/", "/");
```
dkrylovsb: I did not see `80` anywhere within the URLs, in which case we _might_ not need this replacement

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -124,14 +125,29 @@
...
+        if (!shouldReplaceWithHttps()) {
+            return request;
+        }
+        request = request.replace("http://", "https://");
```
dkrylovsb: If you'd like to do this replacement in a more OOP way, `org.springframework.web.util.UriComponentsBuilder#scheme` method will do the trick

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
lailasharshar: Oh we do. For some reason, at some point, it ends up creating the url https://localhost:80.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
lailasharshar: Ah, thanks. That is what I was looking for. I just spent so much time googling for it unsuccessfully, that I figured I would put it out there. Maybe that will solve it coming back with :80/

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
lailasharshar: Fair enough, but it's been there forever and I actually found it really helpful.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
lailasharshar: And adding scheme doesn't seem to fix that. It's when it's generating the file names but the request comes in as http.

api/src/main/java/gov/cms/ab2d/api/controller/BulkDataAccessAPI.java
```diff
@@ -124,14 +125,25 @@
         checkResourceTypesAndOutputFormat(resourceTypes, outputFormat);
         checkSinceTime(since);
 
-        Job job = jobService.createJob(resourceTypes, ServletUriComponentsBuilder.fromCurrentRequest().toUriString(),
-                null, outputFormat, since);
+        Job job = jobService.createJob(resourceTypes, getCurrentUrl(), null, outputFormat, since);
```
spathiyil: nice clean up here. I am able to read it better.

PR #237, opened 2020-03-11
Adding a maintenance mode endpoint that is publicly available so that…

adaykin:
… a frontend component will be able to query and determine whether or not the system is in maintenance mode.

Adding a maintenance mode endpoint that is publicly available so that a frontend component will be able to query and determine whether or not the system is in maintenance mode.

### Checklist

- [x] Tests and linting pass

### Security

NA


dkrylovsb:
Similar to `/health`, let's just do `/status` without a prefix.

PR #240, opened 2020-03-12
Feature/ab2d 1150 frontend message v2

adaykin:
Adding a status page to the site with documentation so that developers can check the status

Adding a new md file that will be a separate status page, updating link structure so that links can be navigated too, adding cross origin to the API call, upgrading jquery and moving to the full version, adding status AJAX call to periodically check the status of the page.

### Checklist

- [x] Tests and linting pass

### Security

NA


website/status.md
```diff
@@ -0,0 +1,45 @@
...
+
+---
+
+<script>
+    let baseUrl = 'https://sandbox.ab2d.cms.gov/';
+
+    function pollServer() {
+        $.get(baseUrl + 'status', function(data) {
+            if(data.maintenanceMode === 'false') {
+                $('#status-content').html("<img style=\"vertical-align: middle;\" src='assets/img/status-up.png' /> <span>The system is operating normally</span>");
```
adaykin: Just used a check and x since I couldn't find a thumbs up/down that looked good.

PR #242, opened 2020-03-13
Adding _since parameter to the documentation

adaykin:
Adding _since parameter to documentation.

### Summary

Adding _since parameter to documentation.

### Checklist

- [x] Tests and linting pass

### Security

NA

### DIFF
--- a/website/sandbox/advanced_user_guide.md
+++ b/website/sandbox/advanced_user_guide.md
@@ -212,11 +212,13 @@ GET /api/v1/fhir/Patient/$export
 #### Parameters
 
 The _since parameter can be used to limit data to only data that has been updated since the specified parameter.
-The format is the [ISO 8601 DateTime standard](https://www.w3.org/TR/NOTE-datetime) e.g. YYYY-MM-DDThh:mm:ssTZD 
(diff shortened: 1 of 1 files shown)


website/sandbox/advanced_user_guide.md
```diff
@@ -209,6 +209,15 @@ or all Part D patients registered with the sponsor:
 GET /api/v1/fhir/Patient/$export
 ```
 
+#### Parameters
+
+The _since parameter can be used to limit data to only data that has been updated since the specified parameter.
+The format is the [ISO 8601 DateTime standard](https://www.w3.org/TR/NOTE-datetime) e.g. YYYY-MM-DDThh:mm:ssTZD
```
dkrylovsb: It is important that we talk about the earliest possible date that is allowed for `_since`. Here is an example from BCDA for a reference:

> Only include resource versions that were created at or after the given instant in time.  Format of string must align with the FHIR dateTime datatype.  Dates prior to 2020-02-13 are not supported and will result in a failure response.  Example: `2020-02-13T08:00:00.000-05:00`

adaykin: Changed example date and added text

PR #244, opened 2020-03-13
Clean up BulkDataAccessApi & SecurityConfig

lailasharshar:
### Summary

Changed the swagger ui and api endpoint files so everything wasn't in the same file
Changed instructional text to support changes to swagger ui
Removed non-existent swagger endpoint
Changed /health to a variable instead of using the string everywhere
Reduced the complexity of the doFilterInternal method in JwtTokenAuthenticationFilter

### Checklist

- [ ] Tests and linting pass

### Security

No security changes, just refactoring

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/FileDownloadApi.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/FileDownloadApi.java
@@ -1,0 +1,91 @@
+package gov.cms.ab2d.api.controller;
+
+import gov.cms.ab2d.api.config.SwaggerConfig;
+import gov.cms.ab2d.common.service.JobService;
+import io.swagger.annotations.*;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.IOUtils;
+import org.slf4j.MDC;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.io.Resource;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+import javax.servlet.http.HttpServletResponse;
+import javax.validation.constraints.NotBlank;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+import static gov.cms.ab2d.api.util.Constants.GENERIC_FHIR_ERR_MSG;
+import static gov.cms.ab2d.common.service.JobService.ZIPFORMAT;
+import static gov.cms.ab2d.common.util.Constants.*;
+import static gov.cms.ab2d.common.util.Constants.NDJSON_FIRE_CONTENT_TYPE;
+
+@Slf4j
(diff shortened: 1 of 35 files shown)


api/src/main/java/gov/cms/ab2d/api/controller/StatusAPI.java
```diff
@@ -0,0 +1,191 @@
...
+import static gov.cms.ab2d.api.controller.BulkDataAccessAPI.JOB_NOT_FOUND_ERROR_MSG;
+import static gov.cms.ab2d.api.util.Constants.GENERIC_FHIR_ERR_MSG;
+import static gov.cms.ab2d.api.util.SwaggerConstants.BULK_CANCEL;
+import static gov.cms.ab2d.common.service.JobService.ZIPFORMAT;
+import static gov.cms.ab2d.common.util.Constants.*;
+
+@Slf4j
+@Api(value = "Bulk Data Access API", description = "API to determine the status of the job, the files to download " +
+        "once the job is complete and an endpoint to cancel a job",
+        tags = {"Status"})
```
adaykin: Since this has delete in it too, can we add a tag for Delete?

api/src/main/java/gov/cms/ab2d/api/controller/FileDownloadApi.java
```diff
@@ -0,0 +1,91 @@
...
+import static gov.cms.ab2d.common.service.JobService.ZIPFORMAT;
+import static gov.cms.ab2d.common.util.Constants.*;
+import static gov.cms.ab2d.common.util.Constants.NDJSON_FIRE_CONTENT_TYPE;
+
+@Slf4j
+@Api(value = "Bulk Data File Download API", description = "After creating a job, the API to download the generated bulk download files",
+        tags = {"Download"})
+@RestController
+@RequestMapping(path = API_PREFIX + FHIR_PREFIX, produces = {"application/json", NDJSON_FIRE_CONTENT_TYPE})
+public class FileDownloadApi {
```
dkrylovsb: Let's capitalize `Api` to be consistent with naming across controllers.

api/src/main/java/gov/cms/ab2d/api/controller/StatusAPI.java
lailasharshar: Both GET and DELETE for /status are under this. Technically delete is status.

lailasharshar: Oh, and in case it was clear from the code, the "tags" item is what defines the title of the section.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -254,6 +254,39 @@ private void verifyJsonFromfileDownload(String fileContent, JSONArray extension)
         Assert.assertEquals(length, fileContent.getBytes().length);
     }
 
+    private boolean validFields(JSONObject jsonObject) {
+        List<String> allowedFields = Arrays.asList("identifier", "item", "meta", "patient", "billablePeriod", "diagnosis",
```
adaykin: I think these makes more sense as a Set, since we're doing lookups it's O(1) vs. doing O(n) as a List, and order doesn't seem to matter. A set can still be iterated over.

lailasharshar: ok

PR #262, opened 2020-03-19
Updating efs mount to /opt/ab2d

adaykin:
Updating efs mount to /opt/ab2d

### Summary

Updating efs mount to /opt/ab2d

### Checklist

- [x] Tests and linting pass

### Security

NA


adaykin:
Developers will have to update the keystore location in their env variables, as well as run the command `docker volume rm ab2d_tmp-volume`

PR #267, opened 2020-03-26
Framework for event logging

lailasharshar:
### Summary

This is the first version of the report event logging framework. I wanted to put this out early before I implemented it to get comments and thoughts. 

### Checklist

- [ ] Tests and linting pass

### Security
Unknown at this point - will reassess as we implement


eventlogger/src/main/java/gov/cms/ab2d/eventlogger/LoggableEvent.java
```diff
@@ -0,0 +1,10 @@
+package gov.cms.ab2d.eventlogger;
+
+import java.time.OffsetDateTime;
+
+/**
+ * Interface describing a loggable event
+ */
+public interface LoggableEvent {
+    boolean log(OffsetDateTime eventTime);
```
dkrylovsb: What does return value (true/false) indicate?

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/ApiRequestEventLogger.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.time.OffsetDateTime;
+
+/**
+ * Class to create and log an API request coming from a user
+ */
+@Data
+public class ApiRequestEventLogger implements LoggableEvent {
+    // id
+    private Long id;
```
dkrylovsb: It looks like each and every implementation of `LoggableEvent` has the `id` field, which means it could be a good candidate for moving to an abstract base class?

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/ApiRequestEventLogger.java
```diff
@@ -0,0 +1,29 @@
...
+
+import lombok.Data;
+
+import java.time.OffsetDateTime;
+
+/**
+ * Class to create and log an API request coming from a user
+ */
+@Data
+public class ApiRequestEventLogger implements LoggableEvent {
```
dkrylovsb: So the interface is called `event` while the implementation classes are called `loggers`, which might sound a bit awkward, since loggers aren't events. Should we drop the `Logger` post-fix from the class names?

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/LoggableEvent.java
dkrylovsb: Should `Event` really be able to log itself, or should we separate the concerns and delegate actual logging to `Logger` implementations?

dkrylovsb: I'm sure we'll have at least two `Logger` implementation: one for Splunk, and the other for writing into a DB table.

lailasharshar: That it was able to successfully log the event. Not sure if we'll ever complain about it. What are we going to do, log it?

dkrylovsb: I'm thinking we could demand that each event implementation be able to serialize itself using an agreed-upon format (JSON), in which case this interface would just contain a single method `JSON toJSON()`. Thoughts?

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/ApiRequestEventLogger.java
lailasharshar: Sure

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/LoggableEvent.java
lailasharshar: That's a good point. Maybe even New Relic.

dkrylovsb: `getId` could probably also be declared in the interface, since it appears you're using `id` throughout.

e2e-test/src/test/java/gov/cms/ab2d/e2etest/TestRunner.java
```diff
@@ -104,9 +104,9 @@ public void init() throws IOException, InterruptedException, JSONException {
                     .withScaledService("worker", 2)
                     .withExposedService("db", 5432)
                     .withExposedService("api", 8080, new HostPortWaitStrategy()
-                        .withStartupTimeout(Duration.of(150, SECONDS)));
-                    //.withLogConsumer("worker", new Slf4jLogConsumer(log)) // Use to debug, for now there's too much log data
-                    //.withLogConsumer("api", new Slf4jLogConsumer(log));
+                        .withStartupTimeout(Duration.of(150, SECONDS)))
+                    .withLogConsumer("worker", new Slf4jLogConsumer(log)) // Use to debug, for now there's too much log data
+                    .withLogConsumer("api", new Slf4jLogConsumer(log));
```
dkrylovsb: Let's comment this once the CI issue is fixed

PR #270, opened 2020-03-27
Add page for workflow

adaykin:
Add page for the workflow that will be a skeleton

### Summary

Add page for the workflow that will be a skeleton

### Checklist

- [x] Tests and linting pass

### Security

NA

### DIFF
--- a/website/index.md
+++ b/website/index.md
@@ -13,6 +13,8 @@ sections:
   - name: For Developers
   - name: Status
     link: status
+  - name: Workflow
+    link: workflow
 ctas:
 
 ---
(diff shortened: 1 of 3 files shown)


website/index.md
```diff
@@ -13,6 +13,8 @@ sections:
   - name: For Developers
   - name: Status
     link: status
+  - name: Workflow
```
dkrylovsb: Is this the name we've settled on?

adaykin: I'll check with John

adaykin: Updated, they requested it be ab2d easy

PR #276, opened 2020-03-30
Feature/ab2d 1256 login form

adaykin:
Adding a login form that sends a request to a proxy to okta.

### Summary

Add a login form that has inputs for client ID and client secret. The data is sent to the AB2D server and then sent over to okta to get a token. The token is sent back to the client, which will be able to store it and use for future requests.

### Checklist

- [x] Tests and linting pass

### Security

Uses the test login to generate tokens that will be used by the client.


api/src/main/java/gov/cms/ab2d/api/controller/OktaProxyAPI.java
```diff
@@ -0,0 +1,38 @@
...
+public class OktaProxyAPI {
+
+    @Autowired
+    private OktaTokenGenerator oktaTokenGenerator;
+
+    // This API endpoint does not have authentication, an exception was made in SecurityConfig.java and JWTAuthenticationFilter.java
+    @CrossOrigin // test server for the sandbox runs from port 4000
+    @ResponseStatus(value = HttpStatus.OK)
+    @PostMapping(OKTA_PROXY_ENDPOINT)
+    public ResponseEntity<OktaProxyResponse> getOktaToken(@RequestParam String clientID, @RequestParam String clientSecret) throws URISyntaxException,
```
dkrylovsb: These come within POST body, not query string, correct?

api/src/main/java/gov/cms/ab2d/api/security/OktaTokenGeneratorImpl.java
```diff
@@ -0,0 +1,48 @@
...
+            .version(HttpClient.Version.HTTP_2)
+            .build();
+
+    public String generateToken(String clientID, String clientSecret) throws URISyntaxException, IOException, InterruptedException, JSONException {
+        log.info("Received request to generate token");
+
+        String authString = Base64.getEncoder().encodeToString((clientID + ":" + clientSecret).getBytes());
+
+        HttpRequest request = HttpRequest.newBuilder()
+                .uri(new URI("https://test.idp.idm.cms.gov/oauth2/aus2r7y3gdaFMKBol297/v1/token?grant_type=client_credentials&scope=clientCreds"))
```
dkrylovsb: Okta URL should come from the application configuration, `api.okta-jwt-issuer` specifically.

api/src/main/java/gov/cms/ab2d/api/controller/OktaProxyAPI.java
adaykin: Correct it's coming from the post body 
Screen Shot 2020-03-31 at 11 21 02 AM

PR #278, opened 2020-04-01
Feature/ab2d 1242 reporting persist events

lailasharshar:
### Summary

Added an SQL implementation to the event reporting framework. Each event type has its own table.

### Checklist

- [ ] Tests and linting pass

### Security
These are more tables in the ab2d db. They may not end up there if we decide they need different access permissions.


common/src/main/resources/db/changelog/v001/log_event_schema.sql
```diff
@@ -0,0 +1,131 @@
...
+
+CREATE TABLE event_api_request
+(
+        id SERIAL,
+        time_of_event TIMESTAMP WITH TIME ZONE,
+        job_id VARCHAR(255),
+        user_id VARCHAR(255),
+        url VARCHAR(2048),
+        ip_address VARCHAR(32),
+        token_hash VARCHAR(32),
```
dkrylovsb, line 14: What's token hash? How would we use it?

eventlogger/src/test/resources/application.properties
```diff
@@ -0,0 +1,12 @@
+spring.datasource.url=${DB_URL}
```
adaykin, line 1: Does this work without this file? The modules that aren't main ones have names like application.audit.properties and don't need to specify things like `spring.datasource.username` that are specified elsewhere.

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/sql/ApiRequestEventMapper.java
```diff
@@ -3,12 +3,64 @@
...
 
-public class ApiRequestEventMapper implements SqlEventMapper {
     @Override
     public void log(LoggableEvent event) {
         if (event.getClass() != ApiRequestEvent.class) {
             throw new EventLoggingException("Used " + event.getClass().toString() + " instead of " + ApiRequestEvent.class.toString());
         }
+        String query = "insert into event_api_request " +
+                " (time_of_event, user_id, job_id, url, ip_address, token_hash, request_id) " +
+                " values (?, ?, ?, ?, ?, ?, ?)";
```
dkrylovsb, line 30: Using named sql parameters would definitely be more readable

common/src/main/resources/db/changelog/v001/log_event_schema.sql
lailasharshar: I wanted a way to "save" the token so we can see if they are using the same token but not necessarily have to actually save the entire token - we don't know how large it will end up being (larger than useful) and even so, we pick out the needed data in other places like username. I just want to know if they are the same one to answer questions like, are they are using a new one every time, are they using it after it expires? That sort of thing.

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/sql/ApiRequestEventMapper.java
```diff
@@ -3,12 +3,64 @@
...
+                    .prepareStatement(query, new String[] {"id"});
+            ps.setObject(1, UtilMethods.convertToUtc(be.getTimeOfEvent()));
+            ps.setString(2, be.getUser());
+            ps.setString(3, be.getJobId());
+            ps.setString(4, be.getUrl());
+            ps.setString(5, be.getIpAddress());
+            ps.setString(6, be.getTokenHash());
+            ps.setString(7, be.getRequestId());
+            return ps;
+        }, keyHolder);
```
adaykin, line 45: Is this not using hibernate since it's out of the application's context?

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/sql/ApiRequestEventMapper.java
```diff
@@ -3,12 +3,64 @@
...
+public class ApiRequestEventMapper extends SqlEventMapper {
+    private JdbcTemplate template;
+
+    public ApiRequestEventMapper(JdbcTemplate template) {
+        this.template = template;
+    }
 
-public class ApiRequestEventMapper implements SqlEventMapper {
     @Override
     public void log(LoggableEvent event) {
```
dkrylovsb, line 24: For each `log` method we probably want `@Transactional(propagation = Propagation.REQUIRES_NEW)`

eventlogger/src/test/resources/application.properties
lailasharshar: I just got a bunch or errors without a lot of those properties because I was dependent on common. I might actually move this code into common if common needs to do any logging, otherwise, I get a circular dependency.

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/sql/ApiRequestEventMapper.java
lailasharshar: Interesting. I will.

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/sql/ApiRequestEventMapper.java
lailasharshar: Because I'm separating the event from the logging (so we can log the same event to different places), the event isn't a management entity.

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/sql/ApiRequestEventMapper.java
lailasharshar: ok

dkrylovsb, line 24: We probably also want to enclose the entire body of `log` into try/catch and log (not re-throw) any exception in order to prevent any possible errors/bugs within the logging framework from compromising parent workflows.

PR #279, opened 2020-04-03
Feature/ab2d 1273

adaykin:
Add the status section of ab2d-easy

### Summary

Add a button to start a bulk export that can optionally use a contract. Add a cancel button that can cancel an in progress export. On the backend add cross origin configuration so that headers can be sent and received via cross origin requests. Add a progress bar at the bottom that polls every 6 seconds to check on the status.

### Checklist

- [x] Tests and linting pass

### Security

Add CrossOrigin annotation to several API endpoints.


api/src/main/java/gov/cms/ab2d/api/config/CorsConfig.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+@Configuration
+public class CorsConfig {
+
+    @Bean
+    public FilterRegistrationBean corsFilter() {
+        UrlBasedCorsConfigurationSource source = new UrlBasedCorsConfigurationSource();
+        CorsConfiguration config = new CorsConfiguration();
+        config.setAllowCredentials(true);
+        config.addAllowedOrigin("*");
```
dkrylovsb, line 23: Let's specify the allowed origin instead.

api/src/main/java/gov/cms/ab2d/api/config/CorsConfig.java
```diff
@@ -20,7 +20,7 @@ public FilterRegistrationBean corsFilter() {
         UrlBasedCorsConfigurationSource source = new UrlBasedCorsConfigurationSource();
         CorsConfiguration config = new CorsConfiguration();
         config.setAllowCredentials(true);
-        config.addAllowedOrigin("*");
+        config.addAllowedOrigin("127.0.0.1:4000");
```
dkrylovsb, line 23: Shouldn't this be `ab2d.cms.gov`?

dkrylovsb: We could make this yet another configuration property via an env variable

adaykin: I think when it's deployed there it would be coming from the same origin, when it's running locally the site runs on port 4000, and the app is on port 8080. Unless there's different origins between the app and website when it's deployed?

dkrylovsb: When initiated via the Web site, API requests would show origin of `https://ab2d.cms.gov`, which is still different from the API subdomain, so it will likely fail. We should probably make this an environment variable-driven property just like everything else, and use `https://ab2d.cms.gov` as the default value, while CI & local envs could override as needed.

PR #286, opened 2020-04-09
Feature/ab2d 1303 log events api search load errors

lailasharshar:
### Summary

Logged events associated with:

- API Events
- Error Events
- Beneficiary Searches
- Data loading

The API events were created in the Token filter and the responses either where the API responded with a 200 or where errors were intercepted (in the error handler). 

### Checklist

- [ ] Tests and linting pass

### Security


api/src/main/java/gov/cms/ab2d/api/controller/HealthAPI.java
```diff
@@ -1,24 +1,36 @@
...
 
+    @Autowired
+    private EventLogger eventLogger;
+
     @GetMapping(HEALTH_ENDPOINT)
-    public ResponseEntity<Void> getHealth() {
+    public ResponseEntity<Void> getHealth(HttpServletRequest request) {
         if (healthCheck.healthy()) {
+            eventLogger.log(new ApiResponseEvent(MDC.get(USERNAME), null, HttpStatus.OK, "API Health Ok",
+                    null, (String) request.getAttribute(REQUEST_ID)));
```
dkrylovsb, line 29: Health checks happen very often, so this is going to produce a lot of noise. I'd recommend removing.

api/src/main/java/gov/cms/ab2d/api/controller/HealthAPI.java
```diff
@@ -1,24 +1,36 @@
...
             return new ResponseEntity<>(HttpStatus.OK);
         } else {
+            eventLogger.log(new ApiResponseEvent(MDC.get(USERNAME), null, HttpStatus.INTERNAL_SERVER_ERROR, "API Health NOT Ok",
+                    null, (String) request.getAttribute(REQUEST_ID)));
```
dkrylovsb, line 33: But this, on the other hand, we should keep. "Health Not Ok" should definitely be logged.

e2e-test/src/test/resources/local-config.yml
```diff
@@ -1,2 +1,2 @@
 okta-url: 'https://test.idp.idm.cms.gov/oauth2/aus2r7y3gdaFMKBol297/v1/token'
-base-url: 'http://localhost:8080'
\ No newline at end of file
+base-url: 'http://localhost:8090'
```
dkrylovsb, line 2: Intentional?

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/utils/UtilMethods.java
```diff
@@ -43,4 +48,43 @@ public static String hashIt(String val) {
...
+
+    public static String getURL(HttpServletRequest req) {
+        String method = req.getMethod();             // GET, POST
+        String scheme = req.getScheme();             // http
+        String serverName = req.getServerName();     // hostname.com
+        int serverPort = req.getServerPort();        // 80
+        String contextPath = req.getContextPath();   // /mywebapp
+        String servletPath = req.getServletPath();   // /servlet/MyServlet
+        String pathInfo = req.getPathInfo();         // /a/b;c=123
+        String queryString = req.getQueryString();   // d=789
```
dkrylovsb, line 69: The API is behind a load balancer, so we'll need to take the `X-Forwarded` headers into account: https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html

e2e-test/src/test/resources/local-config.yml
lailasharshar: Locally. Didn't realize I had checked it in. I'll revert it.

PR #289, opened 2020-04-13
Adding a refresh so that the token will never expire

adaykin:
Adding a refresh so that the token will never expire

### Summary

Adding a refresh so that the token will never expire. This will run every 58 minutes, since the okta token expires after an hour. If a user authenticates with another set of credentials, the timer resets itself.

### Checklist

- [x] Tests and linting pass

### Security

Reauthenticates with okta


adaykin:
Closing since ab2d-easy isn't scheduled for production now

PR #290, opened 2020-04-13
Featuer/ab2d 1241 log events for jobs

lailasharshar:
### Summary

Added log events for job state changes

I did have to move common out of the eventlogger because common needs eventlogger.

### Checklist

- [ ] Tests and linting pass

### Security

### DIFF
--- a/eventlogger/pom.xml
+++ b/eventlogger/pom.xml
@@ -62,7 +62,6 @@
         <dependency>
             <groupId>org.springframework.data</groupId>
             <artifactId>spring-data-jpa</artifactId>
-            <version>2.2.6.RELEASE</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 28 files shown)


eventlogger/pom.xml
```diff
@@ -45,6 +37,44 @@
...
+            <artifactId>commons-codec</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.data</groupId>
+            <artifactId>spring-data-jpa</artifactId>
+            <version>2.2.6.RELEASE</version>
```
dkrylovsb, line 65: Any chance we could avoid specifying version number here so that Spring version remains defined in a single place in parent POM?

lailasharshar: Ok

PR #294, opened 2020-04-15
Feature/ab2d 1239 log events for files

lailasharshar:
### Summary

Added file events for logging

### Checklist

- [ ] Tests and linting pass

### DIFF
--- a/common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
+++ b/common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
@@ -10,6 +10,7 @@ import gov.cms.ab2d.common.model.Sponsor;
 import gov.cms.ab2d.common.model.Contract;
 import gov.cms.ab2d.common.repository.ContractRepository;
 import gov.cms.ab2d.eventlogger.EventLogger;
+import gov.cms.ab2d.eventlogger.events.FileEvent;
 import gov.cms.ab2d.eventlogger.events.JobStatusChangeEvent;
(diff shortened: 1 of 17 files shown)


common/src/main/java/gov/cms/ab2d/common/service/JobServiceImpl.java
```diff
@@ -185,16 +186,18 @@ public Resource getResourceForJob(String jobUuid, String fileName) throws Malfor
...
         JobOutput jobOutput = jobOutputService.findByFilePathAndJob(fileName, job);
         jobOutput.setDownloaded(true);
         jobOutputService.updateJobOutput(jobOutput);
+        eventLogger.log(new FileEvent(
+                job == null || job.getUser() == null ? null : job.getUser().getUsername(),
+                jobUuid, file, FileEvent.FileStatus.DELETE));
+        boolean deleted = file.delete();
+        if (!deleted) {
+            log.error("Was not able to delete the file {}", file.getName());
+        }
```
dkrylovsb, line 200: Did this need to be moved from the beginning of the method to the end?

dkrylovsb: Since this block isn't throwing any exceptions, the move does not seem to have any effect.

lailasharshar: You need to create the event before the file is deleted (otherwise, you can't get the hash or file size) but still need to populate it with job information.

PR #295, opened 2020-04-16
Feature/ab2d 1357 parallel build

adaykin:
Adding parallelism to the builds where there are no tests being ran. 

### Summary

Adding parallelism to the builds where there are no tests being ran. Avoiding areas where mvn calls tests, since there's issues where maven is not able to find a jar, error message in logs says "java.nio.file.NoSuchFileException <name of jar>. The jar is present though when just running `mvn -T 1C clean package -DskipTests`

### Checklist

- [x] Tests and linting pass

### Security

NA

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ before_script:
 - chmod +x ./cc-test-reporter
 - "./cc-test-reporter before-build"
 script:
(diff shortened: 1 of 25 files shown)


adaykin:
Closing for now since parallel build isn't working with tests.

PR #302, opened 2020-04-20
Feature/ab2d 1395 create job for user

adaykin:
…lf. Sets the security context with the specified user's credentials.

Adding an API call for an admin user to create a job on a user's behalf.

### Summary

Adding an API call for an admin user to create a job on a user's behalf. Sets the security context with the specified user's credentials. Uses the same parameters as the patient export call and then delegates to the controller.

### Checklist

- [x] Tests and linting pass

### Security

Modifies the security context and sets it to the user specified by the admin.


api/src/main/java/gov/cms/ab2d/api/controller/AdminAPI.java
```diff
@@ -11,13 +11,16 @@
 import gov.cms.ab2d.eventlogger.events.ReloadEvent;
 import gov.cms.ab2d.hpms.processing.ExcelReportProcessor;
 import gov.cms.ab2d.hpms.processing.ExcelType;
+import io.swagger.annotations.ApiParam;
```
dkrylovsb, line 14: Do we need this since Admin API is not a part of Swagger?

adaykin: No, I just cleaned up the unused imports

PR #307, opened 2020-04-22
Adding self signed certs and TLS setup for app. Allow env variables t…

adaykin:
…o be passed in for prod, sbx environments.

Configuring TLS for the application.

### Summary

Adding self signed certs and TLS setup for app. Allow env variables to be passed in for prod, sbx environments.

### Checklist

- [x] Tests and linting pass

### Security

NA


api/src/main/resources/application.properties
```diff
@@ -31,3 +31,9 @@ logging.level.liquibase=WARN
 ## -----------------------------------------------------------------------------------------------------  Health Checks
 health.requiredSpareMemoryInMB=32
 health.urlsToCheck=http://www.google.com,http://www.facebook.com
+
+## -----------------------------------------------------------------------------------------------------  TLS
+server.ssl.key-store-type=PKCS12
```
dkrylovsb, line 36: We should probably add `server.port=8443` here as well to be explicit.

adaykin: Will that change the port to 8443 in prod or will something override it to be 443?

dkrylovsb: Correct. We're fine with 8443 in Prod since we are behind a balancer anyway.

PR #316, opened 2020-05-04
Feature/ab2d 1459 make kinesis asynchronous

lailasharshar:
### Summary

Added the ability to log to Kinesis asynchronously. This only happens when you are ONLY logging to Kinesis. If you are logging to both places, it blocks so that it can retrieve the AWS Request ID and then save it in the database. Also, currently the only place we need this is where we want something in the data lake but not in the database, for example, when we are doing individual beneficiary searches. If you wanted to do both, you'd have to do it manually, then save the AWS Request ID to the database manually after all logging is done. It didn't make sense to make KinesisLogger aware of the SqlLogger.

### Checklist

- [ ] Tests and linting pass

### Security

AWS credentials are inherited from the container


eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/kinesis/KinesisEventLogger.java
```diff
@@ -1,63 +1,54 @@
...
         }
     }
 
-    static String getJsonString(LoggableEvent event) throws JsonProcessingException {
-        ObjectMapper mapper = new ObjectMapper()
-                .registerModule(new ParameterNamesModule())
-                .registerModule(new Jdk8Module())
-                .registerModule(new JavaTimeModule());
-        return mapper.writeValueAsString(event);
+    boolean isFinished() {
```
adaykin, line 50: Is this used outside of testing? Perhaps put a @VisibleForTesting on it

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/kinesis/KinesisEventLogger.java
```diff
@@ -1,63 +1,54 @@
...
-
-            PutRecordResult putRecordResult = client.putRecord(putRecordRequest);
-            ResponseMetadata data = putRecordResult.getSdkResponseMetadata();
-            event.setAwsId(data.getRequestId());
-        } catch (Exception ex) {
-            log.error("Error in pushing event to Kinesis " + json + " - " + ex.getMessage());
+        ThreadPoolTaskExecutor ex = config.kinesisLogProcessingPool();
+        KinesisEventProcessor processor = new KinesisEventProcessor(event, client, streamId);
+        Future<Void> future = ex.submit(processor);
+        if (block) {
```
adaykin, line 39: Curious what the use case would be when we would want to block?

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/kinesis/KinesisEventProcessor.java
```diff
@@ -0,0 +1,60 @@
...
+    private LoggableEvent event;
+
+    public KinesisEventProcessor(LoggableEvent event, AmazonKinesisFirehose client, String streamPrefix) {
+        this.client = client;
+        this.streamId = streamPrefix;
+        this.event = event;
+    }
+
+    static String getJsonString(LoggableEvent event) throws JsonProcessingException {
+        ObjectMapper mapper = new ObjectMapper()
```
adaykin, line 32: ObjectMapper is threadsafe, you can declare it as a private static field https://stackoverflow.com/questions/3907929/should-i-declare-jacksons-objectmapper-as-a-static-field

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/kinesis/KinesisEventLogger.java
lailasharshar: In theory, it could be. I had thought about adding it to the shutdown process.

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/kinesis/KinesisEventLogger.java
lailasharshar: There are two types of logging. SQL and Kinesis. There is only one situation that we log to Kinesis and not SQL - when we're logging every bene search. We don't want that in the database. It's also a flood of events which is why we want the multi threading. In all other cases, the events are more moderate (job changes status for example) and you want one in the DB and one put to Kinesis. You want it blocking if you want the DB table to include the AWS ID. You only get the AWS ID once the thread is complete (and the put was successful). Blocking allows you to wait for that thread to end so you can get it.

eventlogger/src/main/java/gov/cms/ab2d/eventlogger/eventloggers/kinesis/KinesisEventProcessor.java
lailasharshar: Good to know, thanks.

PR #346, opened 2020-05-28
Add a mapping to return contracts with a user. Create a ContractDTO t…

adaykin:
…hat has fields from the Contract model.

Add a GET request for returning User data.

### Summary

Add a mapping to return contracts with a user. Create a ContractDTO that has fields from the Contract model. The user will have a Set of contracts associated with it.

### Checklist

- [x] Tests and linting pass

### Security

NA


adaykin:
Screen Shot 2020-05-28 at 1 38 03 PM

PR #356, opened 2020-06-16
Bug/ab2d 1667 earliest data date is jan 1 2020

lailasharshar:
### Summary

Previously, we only checked to see if the data was past the attestation date but if they attested in December of last year, they still can only have data past 1/1/2020. Fixed that. Also noticed a bug in eventlogger and fixed it.

### Checklist

- [ ] Tests and linting pass

### Security

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ before_install:
 - export AB2D_HICN_HASH_PEPPER=`cat bfd_sbx_hash_pepper`
 - export AB2D_HICN_HASH_ITER=1000
 - export AB2D_CLAIMS_SKIP_BILLABLE_PERIOD_CHECK=true
+- export AB2D_EARLIEST_DATA_DATE=01/01/1900
 install: true
 before_script:
 - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64
--- a/eventlogger/src/test/java/gov/cms/ab2d/eventlogger/eventloggers/sql/AllMapperEventTest.java
+++ b/eventlogger/src/test/java/gov/cms/ab2d/eventlogger/eventloggers/sql/AllMapperEventTest.java
@@ -49,10 +49,26 @@ public class AllMapperEventTest {
     }
 
     @Test
+    void createError() {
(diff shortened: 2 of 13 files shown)


eventlogger/src/test/java/gov/cms/ab2d/eventlogger/eventloggers/sql/AllMapperEventTest.java
```diff
@@ -48,11 +48,27 @@ void exceptionApiRequestTests() {
                 new ApiRequestEventMapper(null).log(new ErrorEvent()));
     }
 
+    @Test
+    void createError() {
+        // There is no mapping. Make sure it doesn't create an issue
+        BeneficiarySearchEvent bene = new BeneficiarySearchEvent();
+        sqlEventLogger.log(bene);
+        sqlEventLogger.updateAwsId("TEST", bene);
```
adaykin, line 56: Is the idea with this test and the one below to see if it doesn't throw an exception?

.travis.yml
```diff
@@ -40,6 +40,7 @@ before_install:
 - export AB2D_HICN_HASH_PEPPER=`cat bfd_sbx_hash_pepper`
 - export AB2D_HICN_HASH_ITER=1000
 - export AB2D_CLAIMS_SKIP_BILLABLE_PERIOD_CHECK=true
+- export AB2D_EARLIEST_DATA_DATE=01/01/1900
```
adaykin, line 43: It's 1900 here and 2020 elsewhere is that b/c our tests are relying on earlier dates?

eventlogger/src/test/java/gov/cms/ab2d/eventlogger/eventloggers/sql/AllMapperEventTest.java
lailasharshar: Yes, even though it doesn't have an sql mapping. I don't want logging to mess anything up.

.travis.yml
lailasharshar: Correct. The real one is 2020, but the e2e tests rely on test data from 199-2002.

PR #357, opened 2020-06-16
Feature/ab2d 1521 create aggregate job event

lailasharshar:
### Summary

Created an aggregate object that will be sent to Kinesis that combines data from the other events and enables us to do easier searches in Athena. Originally this was created in AWS Lambda but it took minutes to run when you can just do it for free in the DB when a job is done. A job is done when all its files are deleted, either when they have all been downloaded or when they have expired.

### Checklist

- [ ] Tests and linting pass

### Security

### DIFF
--- a/audit/src/main/java/gov/cms/ab2d/audit/cleanup/FileDeletionServiceImpl.java
+++ b/audit/src/main/java/gov/cms/ab2d/audit/cleanup/FileDeletionServiceImpl.java
@@ -5,8 +5,8 @@ import gov.cms.ab2d.common.service.JobService;
 import gov.cms.ab2d.common.service.ResourceNotFoundException;
 import gov.cms.ab2d.eventlogger.LogManager;
 import gov.cms.ab2d.eventlogger.events.FileEvent;
+import gov.cms.ab2d.eventlogger.reports.sql.DoSummary;
 import lombok.extern.slf4j.Slf4j;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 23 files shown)


audit/src/main/java/gov/cms/ab2d/audit/cleanup/FileDeletionServiceImpl.java
```diff
@@ -52,11 +59,25 @@ public void deleteFiles() {
...
+                // Get the JobId from the name
+                var jobUuid = new File(file.toUri()).getParentFile().getName();
+                // See if the file should be deleted
+                boolean deleted = deleteFile(file);
+                if (deleted) {
+                    // If it was deleted
+                    jobsDeleted.add(jobUuid);
+                }
+            }
+            for (String job : jobsDeleted) {
```
adaykin, line 75: Not a huge deal but why do you need an extra for loop here instead of logging in the if statement above?

lailasharshar: I only adds jobs whose files are deleted because those are the only jobs that are done. If you call deleteFile it doesn't necessarily delete the file. It checks to see if it's old enough and if it isn't, it doesn't delete it. I didn't want to log jobs whose files were deleted in that process.

adaykin: Wouldn't it then make sense to log it in the if statement since you know at that point it was deleted? (and you'd avoid having to maintain the list).

lailasharshar: That would result in the multiple aggregate events being created for the same job since there could be multiple files deleted for the same job. jobsDeleted is a Set so it deduplicates.

PR #360, opened 2020-06-22
Enhancement/ab2d 1597 turn off opt out

lailasharshar:
### Summary

Created flag to bypass opt out data import and not check to see if a beneficiary has opted out during contract search.

### Checklist

- [ ] Tests and linting pass

### Security


adaykin:
Surprised the testOptOut in TestRunner is still passing, did you see if opt out is still running there?

PR #361, opened 2020-06-23
Using the attestation date for the since parameter if since is not pa…

adaykin:
…ssed in.

Using the attestation date for the since parameter if since is not passed in.

Using the attestation date for the since parameter if since is not passed in.

### Checklist

- [x] Tests and linting pass

### Security

NA


lailasharshar:
Correction, those with attestation dates before 2/13 should default to no _since parameter.

PR #362, opened 2020-06-24
Feature/ab2d 1679 create executable jar for e2e

lailasharshar:
### Summary

Create fat jar so that e2e tests can be run on the command line. The syntax is:

java -cp e2e-test-0.0.1-SNAPSHOT-fat-tests.jar gov.cms.ab2d.e2etest.TestLauncher DEV

where the parameter can be (DEV, LOCAL, SBX, IMP, PROD)

### Checklist

- [ ] Tests and linting pass

### Security


adaykin:
For some reason the LGTM on java is failing here, I tried rerunning and it didn't work

e2e-test/src/main/java/Test.java
```diff
@@ -0,0 +1,5 @@
+public class Test {
+    public static void main(String[] args) {
+
```
adaykin, line 3: What's the purpose of this file?

lailasharshar: Whenever I would build the jar, it would say that the jar was empty because there was nothing in the main directory so it was confused. It would be a warning locally but when Travis built it, it would be an error and not move forward. I can probably figure out why this is happening and if I can fix it, but I didn't want to hold anything up over it.

PR #368, opened 2020-06-30
Feature/ab2d 1709 contract to beneficiary mapping calls to bfd multithreaded

lailasharshar:
### Summary

Made the search for beneficiaries in a contract multithreaded per contract/month

### Checklist

- [ ] Tests and linting pass

### Security


worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractBeneSearchImpl.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+        return null;
+    }
+
+    private void addDateRangeToExistingOrNewPatient(ContractBeneficiaries beneficiaries,
+                                                    DateRange monthDateRange, String bfdPatientId) {
+        if (beneficiaries == null) {
+            return;
+        }
+        Optional<PatientDTO> optPatient = beneficiaries.getPatients().stream()
```
adaykin, line 141: This code may run into issues when there's a large amount of patients if it's searching through a list of possibly hundreds of thousands for each mapping, did you measure this on PDP-10000? It may be worth keeping a map of patient ids -> patients, or add the date range at an earlier point in time so that not all of these are being searched through.

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientContractCallable.java
```diff
@@ -0,0 +1,103 @@
...
+
+    public PatientContractCallable(String contractNumber, int month, BFDClient bfdClient) {
+        this.month = month;
+        this.contractNumber = contractNumber;
+        this.bfdClient = bfdClient;
+    }
+
+    @Override
+    public ContractMapping call() throws Exception {
+        System.out.println(Thread.currentThread().getName() + "-" + Thread.currentThread().getId());
```
adaykin, line 32: This should be a log statement or just remove if it was for debugging.

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientContractCallable.java
```diff
@@ -0,0 +1,103 @@
...
+            Bundle bundle = getBundle(contractNumber, month);
+            final Set<String> patientIDs = extractPatientIDs(bundle);
+
+            while (bundle.getLink(Bundle.LINK_NEXT) != null) {
+                bundle = bfdClient.requestNextBundleFromServer(bundle);
+                patientIDs.addAll(extractPatientIDs(bundle));
+            }
+            mapping.setPatients(patientIDs);
+            log.debug("finished reading [{}] Set<String>resources", patientIDs.size());
+            System.out.println("finished reading [" + patientIDs.size() + "] patients");
```
adaykin, line 45: Same with this statement.

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientContractCallable.java
lailasharshar: I wanted an easy to check in the log when jobs start and end so that we can see it's doing them in parallel. I can remove it later. During some testing, I had issues with things going one by one, not sure why, but this way, we can make sure that's not happening..

worker/src/main/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractBeneSearchImpl.java
lailasharshar: I can clean this up, but I don't think it will be a significant factor in your numbers so shouldn't hold up performance tests.

lailasharshar: I did some quick tests and it's probably still under a second to do all these for up to a million patients.

adaykin: Can you add a quick log message for timing between lines 60 and 68? When I did just the filtering iterating on a list of 200K in a test it took over 2 min.

Just something like 
long start = System.currentTimeMillis();
long end = System.currentTimeMillis();
log.info("Time to add date range to patients {}", end - start);

lailasharshar: Just saw this. Okay, done.

PR #399, opened 2020-07-27
Make sure patients not belonging to contracts are not included in files

lailasharshar:
### Summary

If the patients not found in the contract beneficiary map, they should be excluded from the file.

Cleaned up the sample test data so they all referred to the same patient ID because some tests were using one patient ID but returning an EOB file with a different patient so based on this update, it would have been excluded.

### Checklist

- [ ] Tests and linting pass

### Security
This increases security by double checking results from BFD making sure the EOB records are only from patients who were included in the list of the contract's beneficiaries

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
@@ -218,6 +218,7 @@ public class PatientClaimsProcessorImpl implements PatientClaimsProcessor {
         patientId = patientId.replaceFirst("Patient/", "");
         if (patients.get(patientId) == null) {
             log.error(patientId + " returned in EOB, but not a member of a contract");
+            return false;
         }
         return true;
     }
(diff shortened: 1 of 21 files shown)


worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
```diff
@@ -191,7 +192,29 @@ private void handleException(StreamHelper helper, String data, Exception e) thro
...
+     * returns true if the patient is a valid member of a contract, false otherwise. If either value is empty,
+     * it returns false
+     *
+     * @param benefit - The benefit to check
+     * @param patients - the patient map containing the patient id & patient object
+     * @return true if this patient is a member of the correct contract
+     */
+    boolean validPatientInContract(ExplanationOfBenefit benefit, Map<String, ContractBeneficiaries.PatientDTO> patients) {
+        if (benefit == null || patients == null) {
+            return false;
```
adaykin, line 211: There should be some sort of logging if there is invalid patients. We would want to know about this and try and notify BFD.

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
```diff
@@ -191,7 +192,29 @@ private void handleException(StreamHelper helper, String data, Exception e) thro
...
+        }
+        String patientId = benefit.getPatient().getReference();
+        if (patientId == null) {
+            return false;
+        }
+        patientId = patientId.replaceFirst("Patient/", "");
+        return patients.get(patientId) != null;
```
adaykin, line 218: Same goes for these other statements.

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
lailasharshar: Returning false here is most likely a usage issue, not a data issue so just made it a log.debug here, but I added a log.error to the not found one.

PR #409, opened 2020-08-07
Feature/ab2d 1838 disengage worker

nflearl:
**JIRA Tickets:**

[AB2D-1838](https://jira.cms.gov/browse/AB2D-1838) - Create the ability for a worker to be taken offline

***Related Tickets***
 
### What Does This PR Do?

Upon setting the **worker.engaged** property to _idle_, jobs will no longer be processed.  This allows current jobs to finish so the system can be upgraded.

Making sure that the **worker.engaged** property is appropriately handled.

N/A

N/A

### Database Changes

N/A

### Limitations

N/A

### Security Implications

Nothing

### Submitter Checklist


common/src/main/java/gov/cms/ab2d/common/service/PropertiesServiceImpl.java
```diff
@@ -68,27 +69,53 @@ public Properties getPropertiesByKey(String key) {
...
-                propertiesDTO.getKey().equals(ZIP_SUPPORT_ON)) {
-                validateBoolean(key, propertiesDTO);
-                addUpdatedPropertiesToList(propertiesDTOsReturn, propertiesDTO);
-            }
+            validateProperty(propertiesDTOsReturn, propertiesDTO, key);
         }
         return propertiesDTOsReturn;
     }
 
+    private void validateProperty(List<PropertiesDTO> propertiesDTOsReturn, PropertiesDTO propertiesDTO, String key) {
```
lailasharshar, line 77: Nice refactor

nflearl: Thanks.  More than one elseif statement is usually a sign that delegation would be better (or something else more direct(.

nflearl:
> I'm curious why you decided on a String instead of a boolean for worker engagement? Are you imagining other states? Also, I don't see a test to verify that you can change the value and that the job will not be engaged.

I tend to shy away from booleans and prefer enums (even for 2 states).  Gives one better type safety when assigning values and also easily accommodates expansion should new states/values come up in requirements.  Also, enum values have more meaning and read more literally than figuring out what true or false means.

I will add another test and resubmit.

PR #418, opened 2020-08-17
Feature/ab2d 1944 limit logged requests

wnyffenegger:
**JIRA Tickets:**

[AB2D-1944](https://jira.cms.gov/browse/AB2D-1944) - Description

***Related Tickets***
 
### What Does This PR Do?

Allows regex filters to be applied to public request logging on the api. By default only one filter is included for /health; however, more filters can be applied via environment variable.

* Issue with separating regular expressions representing URIs using commas in the properties file / environment variable.
* Concerns about how responses are handled if no request id is set because the request was not logged

N/A

N/A

### Database Changes

N/A schema changes; but, you will see less events logged to the database.

### Limitations

Currently is not applied to authenticated requests.

### Security Implications

### Submitter Checklist

### DIFF
--- a/api/src/main/resources/application.properties
+++ b/api/src/main/resources/application.properties
@@ -15,7 +15,7 @@ api.okta-jwt-audience=${AB2D_OKTA_JWT_AUDIENCE:#{'AB2D'}}
 api.okta-connection-timeout=${AB2D_OKTA_CONNECTION_TIMEOUT:#{5}}
 api.okta-read-timeout=${AB2D_OKTA_READ_TIMEOUT:#{5}}
 
-api.requestlogging.filter=/health.*
+api.requestlogging.filter=^/health$
 
 server.compression.enabled=true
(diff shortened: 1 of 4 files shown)


api/src/main/resources/application.properties
```diff
@@ -15,6 +15,8 @@ api.okta-jwt-audience=${AB2D_OKTA_JWT_AUDIENCE:#{'AB2D'}}
 api.okta-connection-timeout=${AB2D_OKTA_CONNECTION_TIMEOUT:#{5}}
 api.okta-read-timeout=${AB2D_OKTA_READ_TIMEOUT:#{5}}
 
+api.requestlogging.filter=^/health$
```
adaykin, line 18: Let's add a note here that adding a comma will cause the filters to be split, just in case someone tries to add a comma in a regex itself.

wnyffenegger: commented and pushed

PR #421, opened 2020-08-20
Enhancement/ab2d 1885 start bene search while still doing contract search

lailasharshar:
**JIRA Tickets:**

[AB2D-42](https://jira.cms.gov/browse/AB2D-1839) - Begin searching for beneficiaries once a contract beneficiary search is complete

### What Does This PR Do?

Currently, when we perform a batch download of EOBs we perform the following steps for each contract in the job:
- Search for all beneficiaries associated with a contract for each month since 1/1
- For each beneficiary in the list, search for their EOBs

Each beneficiary search creates a map of coverage for each beneficiary so you might know that a beneficiary was serviced by that PDP for 3 out of 8 months. When you do an EOB search, you search for all EOBs for that beneficiary for all time ranges. If an EOB comes back outside of that range, you discard it.

The problem we discovered is that searching for the beneficiaries themselves, although only 8 calls, is taking a third up to 2/3 of the job time. We multithreaded the searches and that helped, but I realized you don't have to wait for all contract beneficiaries to come back before you start searching for individual beneficiaries EOBs. The algorithm is more like this:
- Create threads for each contract beneficiary membership search
- When one of those threads completes, search for all unsearched beneficiaries that was returned from that search.
- Hold EOBs that you don't know should be included in the results until that month has been searched. Let's say it's March. That means you've searched from contract membership for Jan, Feb, March. Let's say that the January search is done first. You search for all EOBs for beneficiaries that were members in January. If you get EOBs with a March billing date, you don't yet know if they were members in March until the March membership mappings are done. However, if you've already done March and the person wasn't a member, you know you can discard the EOB.

Here's an example:

- It's March, you create membership search thread for Jan, Feb, March.
- March comes back and returns: b3, b4, b6, b7 (you annotate those beneficiaries with coverage in March)
- You start the search for EOBs for b3, b4, b6, b7
- You receive the EOB search results for b3. If the EOB billing date is in March, it's valid. If it's in Jan, we don't know yet, hold on to it.
- January comes back and returns: b1, b2, b6, b7 (you annotate those beneficiaries with coverage in Jan)
- You start the search for EOBs for b1, b2 (these are the new beneficiaries not seen in March's coverage)
- You receive the EOB search result for b2. If the EOB billing date is in January, it's valid. If it's in March, it isn't because we've already received those membership results. If it's in February, we still don't know.
- February comes back and returns: b3, b4, b5, b6, b7
- You start the search for EOBs for b5 (the only item not in other EOB searches)

We wait for all threads to return discarding EOBs as more information about membership is known. Once an EOB is validated above, it is written to the file.

The benefits of this approach is the following:
- We can begin the search of EOBs after only one membership search and after each search, each resulting search has fewer beneficiaries to search.
- It required a lot of unraveling of code to do this. Instead of most of the work being done in the threads, I changed it just to return the search results of either the beneficiary membership or EOBs and had the job manage things like filtering and writing the data. This allows us to separate concerns and is a better design so in the future if we want to separate actual searches from the worker code.

Disadvantages
- The job processor has to keep track of a lot of details. When to search, when to filter, when to write, what do we know, what are we still waiting for. The complexity of that element increased.
- It won't be needed if we eventually cache membership data which we might have to do because we can only search this year's membership.

Needed improvements:
- New Relic wasn't added to this code. Will need to consult with Andy.
- We don't deduplicate membership across contracts. If a beneficiary moves between contract 1 and contract 2 in that time period for the same organization for a multi-contract search, we search for that member twice.
- Only one contract is running at a time.

The logic of JobProcessorImpl.processContracts.

### Security Implications

### Submitter Checklist


worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ContractEobManager.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    static List<Integer> getCoveredMonths(Period period) {
+        Date startDate = period.getStart();
+        int startMonth = getMonth(startDate);
+        int startYear = getYear(startDate);
+        Date endDate = period.getEnd();
+        int endMonth = getMonth(endDate);
+        int endYear = getYear(endDate);
+        List<Integer> allMonths = List.of(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12);
```
adaykin, line 199: Small change, but seems like allMonths can move to the class level with it being read only to avoid initializing every time.

worker/src/test/resources/application.properties
```diff
@@ -64,4 +61,5 @@ optout.used=${AB2D_OPT_OUT_ENABLED:#{false}}
 
 ## ------------------------------------------------------------------------------ Earliest Data Date
 bfd.earliest.data.data=01/01/2020
-
+bfd.earliest.data.date.special.contracts=01/01/1950
+bfd.special.contracts=S0001,Z0001
```
adaykin, line 65: Since this is referenced in code do you need to add it to the regular properties file too?

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -76,7 +112,7 @@ public Job process(final String jobUuid) {
 
         // Load the job
         final Job job = jobRepository.findByJobUuid(jobUuid);
-        log.info("Found job");
+        log.info("Found job " + job.getJobUuid());
```
adaykin, line 115: I think the uuid is already coming through as part of MDC

adaykin: In JobHandler.handleMessage

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -111,65 +149,290 @@ public Job process(final String jobUuid) {
...
+            String contractNum = contract.getContractNumber();
+            progressTracker.addContract(contractNum);
+            ContractEobManager contractEobManager = new ContractEobManager(fhirContext, skipBillablePeriodCheck,
+                    getStartDate(contractNum), contract.getAttestedOn());
+
+            // Retrieve the contract beneficiaries
+            //for (Contract contract : contracts) {
+            try {
+                // Init objects
+                StreamHelper helper = new TextStreamHelperImpl(outputDirPath, contractNum, ndjsonRollOver * Constants.ONE_MEGA_BYTE, tryLockTimeout, eventLogger, job);
```
adaykin, line 176: Can you do a try w/ resources here?

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -111,65 +149,290 @@ public Job process(final String jobUuid) {
...
+
+                ContractBeneficiaries contractBeneficiaries = buildCB(contractNum);
+
+                progressTracker.setCurrentMonth(month);
+
+                List<Future<ContractMapping>> contractBeneFutureHandles = createAllContractMappingFutures(month, contractNum, contract.getAttestedOn());
+                List<Future<EobSearchResponse>> eobFutureHandles = new ArrayList<>();
+
+                 // Iterate through the futures
+                while (!contractBeneFutureHandles.isEmpty() || !eobFutureHandles.isEmpty()) {
```
adaykin, line 189: Could contractBeneFutureHandles be empty before the eobFutreHandles is, leaving eobFutureHandles that are unprocessed?

lailasharshar: No, but maybe:
! list1.isEmpty() || ! list2.isEmpty()
is not as clear as:
list1.size() > 0 || list2.size() > 0 although I think isEmpty() is in theory more efficient

nflearl: isEmpty() is absolutely more efficient for certain kinds of collections (such as a queue whose size is constantly fluctuation) and is strongly preferred for existence checks.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -308,23 +555,99 @@ private void doDelete(Path path) {
...
+
+        String dateToUse = startDate;
+        if (isContractSpecial(contract)) {
+            dateToUse = startDateSpecialContracts;
+        }
+        Date date;
+        try {
+            date = sdf.parse(dateToUse);
+        } catch (ParseException e) {
+            LocalDateTime d = LocalDateTime.of(2020, 1, 1, 0, 0, 0, 0);
```
adaykin, line 644: Should this be current year instead of hardcoding it to 2020?

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
```diff
@@ -154,90 +76,59 @@ private void handleException(StreamHelper helper, String data, Exception e) thro
...
+                // .filter(resource -> skipBillablePeriodCheck || FilterOutByDate.valid((ExplanationOfBenefit) resource, attDate, earliestDate, dateRanges))
                 // filter it
                 .map(resource -> ExplanationOfBenefitTrimmer.getBenefit((ExplanationOfBenefit) resource))
                 // Remove any empty values
                 .filter(Objects::nonNull)
                 // Remove Plan D
                 .filter(resource -> !isPartD(resource))
                 // Make sure the returned patient ID is actually part of the contract
-                .filter(resource -> validPatientInContract(resource, patientsMap))
+                //.filter(resource -> validPatientInContract(resource, patientsMap))
```
adaykin, line 130: Do these lines still need to be commented out?

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
nflearl: It vastly depends upon the implementation of the Collection.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
```diff
@@ -111,65 +149,276 @@ public Job process(final String jobUuid) {
...
      * @param month - the month to search for beneficiaries for
      * @param outputDirPath - the location of the job output
      * @param progressTracker - the progress tracker which indicates how far the job is along
      * @throws ExecutionException when there is an issue with searching
      * @throws InterruptedException - when the search is interrupted
      */
-    void processContract(Contract contract, Job job, int month, Path outputDirPath, ProgressTracker progressTracker) throws ExecutionException, InterruptedException {
-        log.info("Job [{}] - contract [{}] ", job.getJobUuid(), contract.getContractNumber());
-        // Retrieve the contract beneficiaries
+    private void processContracts(List<Contract> contracts, Job job, int month, Path outputDirPath,
```
nflearl, line 160: This method is way too big to understand plainly and would be significantly helped by several refactoring to methods for the inner operations.  Then, this method would read more plainly and have the same level of abstraction.

worker/pom.xml
```diff
@@ -132,7 +132,7 @@
                                         <limit>
                                             <counter>LINE</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>87%</minimum>
+                                            <minimum>85%</minimum>
```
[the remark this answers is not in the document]
adaykin: We shoot for being above 85% some modules are easier to cover than others due to inherent differences in code.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
lailasharshar: Technically, it's true, for ArrayList, it is the same because the implementation maintains the size as an independent variable and updates it through removes and adds, but in terms of best practices, it is recommended to use isEmpty. I'm guessing it's because you may end up changing the underlying collection for some reason. Everytime I use size(), some plugin of IntelliJ complains.

worker/src/main/java/gov/cms/ab2d/worker/processor/JobProcessorImpl.java
lailasharshar: No, it's the earliest date we are allowing data to be returned. Next year, it will also be 1/1/2020.

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
lailasharshar: Yes. I left them in here for the time being because I want it obvious it has moved out of this section. We don't have the full list of patients and when they were members of the contract at this point.

filter/src/main/java/gov/cms/ab2d/filter/FilterOutByDate.java
```diff
@@ -231,12 +232,14 @@ public static DateRange getDateRange(int month, int year) throws ParseException
         return benes.stream().filter(b -> valid(b, attestationDate, earliestDate, dateRanges)).collect(Collectors.toList());
     }
 
-    public static boolean valid(ExplanationOfBenefit bene, Date attestationDate, Date earliestDate, List<DateRange> dateRanges) {
+    public static boolean valid(ExplanationOfBenefit bene, OffsetDateTime attestationDate, Date earliestDate, List<DateRange> dateRanges) {
```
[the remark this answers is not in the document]
lailasharshar: It's such a pain to do conversions.

adaykin (dismissed):

worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ContractEobManager.java
```diff
@@ -0,0 +1,231 @@
...
+
+    public void writeValidEobs(StreamHelper helper) throws IOException {
+        IParser jsonParser = fhirContext.newJsonParser();
+        String payload = "";
+        int count = 0;
+        try {
+            for (Map.Entry<String, EobSearchResponse> response : validEobs.entrySet()) {
+                Iterator<Resource> resources = response.getValue().getResources().iterator();
+                while (resources.hasNext()) {
+                    Resource r = resources.next();
```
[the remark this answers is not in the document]
lailasharshar: Iterator.forEachRemaining won't work because you are removing items from the list during the loop as you are writing to the file. I extract the file writing to another method.

PR #427, opened 2020-08-31
Fixed progress tracker percentage

lailasharshar:
**JIRA Tickets:**

[AB2D-42](https://jira.cms.gov/browse/AB2D-2017) - Description

### What Does This PR Do?

Jobs with multiple contracts were showing percent done above 100%. Thought it may be because it was giving 100% to each contract, but found that wasn't the case. The only thing I could think of is when a contract's beneficiary list is not defined yet, it's not included in the percentage. Had the algorithm estimate based on existing contracts what those values will be. Also, if it wants to return a value > 100, let it be 99. The good news is with the new design, we won't have to worry about this problem because we'll know the entire job contract beneficiaries before the search starts.

### Security Implications

### Submitter Checklist

- [X] This PR includes any required documentation changes, (`README`, changelog / release notes, website).
- [X] All tech debt and/or shortcomings introduced by this PR are detailed in `TODO` and/or `FIXME` comments.

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
@@ -146,7 +146,7 @@ public class ProgressTracker {
         double percentBenesDone = 0;
         int totalPossibleCount = getTotalPossibleCount();
         if (totalPossibleCount != 0) {
-            double percentBenesDonePart = ((double) processedCount / totalPossibleCount);
+            double percentBenesDonePart = (double) processedCount / totalPossibleCount;
             if (percentBenesDonePart > 1.0) {
                 log.error("Percent of beneficiaries done is more than 100%");
                 percentBenesDonePart = 1.0;
--- a/worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
@@ -5,6 +5,7 @@ import lombok.Builder;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.Singular;
+import lombok.extern.slf4j.Slf4j;
 
 import java.util.ArrayList;
 import java.util.IntSummaryStatistics;
@@ -12,6 +13,7 @@ import java.util.List;
 
 @Getter
 @Builder
+@Slf4j
 public class ProgressTracker {
 
     private final String jobUuid;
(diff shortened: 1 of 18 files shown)


worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
```diff
@@ -97,6 +98,21 @@ public int getTotalCount() {
         return totalCount;
     }
 
+    public int getTotalPossibleCount() {
+        int remainingToBeDefined = numContracts - patientsByContracts.size();
+        return getTotalCount() + (remainingToBeDefined * getTotalAverageNumber());
+    }
+
+    public int getTotalAverageNumber() {
+        // If we don't have any contract mappings, assume that it's max value so the denominator will underestimate amount done
```
nflearl, line 107: That is one serious underestimation.

worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
```diff
@@ -7,6 +7,7 @@
 import lombok.Singular;
 
 import java.util.ArrayList;
+import java.util.IntSummaryStatistics;
```
nflearl, line 10: Did not know about this handy class.

worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
```diff
@@ -97,6 +98,21 @@ public int getTotalCount() {
...
+        return getTotalCount() + (remainingToBeDefined * getTotalAverageNumber());
+    }
+
+    public int getTotalAverageNumber() {
+        // If we don't have any contract mappings, assume that it's max value so the denominator will underestimate amount done
+        if (getTotalCount() == 0) {
+            return Integer.MAX_VALUE;
+        }
+        // return the average number of patients in known contracts
+        IntSummaryStatistics stats = patientsByContracts.stream().mapToInt(c -> c.getPatients().size()).summaryStatistics();
```
adaykin, line 112: If this is done partway through before all patient counts per contract have been figured out what sort of number will it give?

worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
lailasharshar: Me either. Thank you StackOverflow.

worker/src/main/java/gov/cms/ab2d/worker/processor/domainmodel/ProgressTracker.java
lailasharshar: If there are two out of three contracts defined, the potential total number of patients are a + b + average(a + b). If a is 10, b is 20 and we don't know about c, the potential is estimated to be 45.

adaykin: Ok. I guess my concern was since when I saw the issue in prod the progress shot up pretty quickly to 400% done (when it was probably really only 30%) is it now going to go to 99% pretty quickly too with the same sponsor?

nflearl (dismissed):

PR #428, opened 2020-09-01
Feature/ab2d 1849 poll hpms

nflearl:
**JIRA Tickets:**

Load data from HPMS into AB2D

***Related Tickets***

None 
### What Does This PR Do?

Off a scheduled timer it calls HPMS endpoints to download current organization and attestation information.

The use of WebFlux which allows for asynchronous HTTP calls coupled with the use of only one thread for many requests.

None

None

### Database Changes
None

### Limitations
Without a real site to test against, the scheduling code has been commented out.

### Security Implications

None

### Submitter Checklist

- [ x] This PR includes any required documentation changes, (`README`, changelog / release notes, website).
- [ x] All tech debt and/or shortcomings introduced by this PR are detailed in `TODO` and/or `FIXME` comments.
- [ x] Code checked for PHI/PII exposure

### DIFF
--- a/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
+++ b/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
@@ -12,9 +12,12 @@ import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
 import javax.persistence.OneToMany;
-import javax.persistence.Transient;
 import javax.validation.constraints.NotNull;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -23,7 +26,6 @@ import java.util.Set;
 @Setter
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
 public class Contract {
-    public enum ContractState { UNATTESTED, VALID, ATTESTATION_WITHDRAWN, DELETED }
 
     @Id
     @GeneratedValue
@@ -47,14 +49,34 @@ public class Contract {
     @OneToMany(mappedBy = "contract")
     private Set<Coverage> coverages = new HashSet<>();
 
-    @Transient
-    private ContractState contractState;
-
-    public boolean isDeleted() {
-        return ContractState.DELETED == contractState;
+    public boolean hasAttestation() {
+        return attestedOn != null;
     }
 
-    public void markDeleted() {
-        contractState = ContractState.DELETED;
+    public void clearAttestation() {
+        attestedOn = null;
+    }
+
+    public void markAttested(OffsetDateTime attestedOnDate) {
+        attestedOn = attestedOnDate;
+    }
+
+    /*
+     * Returns true if new state differs from existing which requires a save.
+     */
+    public boolean updateAttestation(boolean attested, String attestationDate, DateTimeFormatter formatter) {
+        boolean hasAttestation = hasAttestation();
+        if (attested == hasAttestation) {
+            return false;   // No changes needed
+        }
+
+        if (hasAttestation) {
+            clearAttestation();
+            return true;
+        }
+
+        LocalDateTime ldt = LocalDate.parse(attestationDate, formatter).atStartOfDay();
+        markAttested(OffsetDateTime.of(ldt, ZoneOffset.UTC)); // todo: make sure offset is correct.
+        return true;
     }
 }
--- a/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
+++ b/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
@@ -22,6 +22,7 @@ import java.util.Set;
 @Setter
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
 public class Contract {
+    public enum ContractState { UNATTESTED, VALID, ATTESTATION_WITHDRAWN, DELETED };
 
     @Id
     @GeneratedValue
@@ -44,4 +45,15 @@ public class Contract {
 
     @OneToMany(mappedBy = "contract")
     private Set<Coverage> coverages = new HashSet<>();
+
+    @Transient
+    private ContractState contractState;
+
+    public boolean isDeleted() {
+        return ContractState.DELETED == contractState;
+    }
+
+    public void markDeleted() {
+        contractState = ContractState.DELETED;
+    }
 }
--- a/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
+++ b/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
@@ -12,6 +12,7 @@ import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
 import javax.persistence.OneToMany;
+import javax.persistence.Transient;
 import javax.validation.constraints.NotNull;
 import java.time.OffsetDateTime;
 import java.util.HashSet;
@@ -22,7 +23,7 @@ import java.util.Set;
 @Setter
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
 public class Contract {
-    public enum ContractState { UNATTESTED, VALID, ATTESTATION_WITHDRAWN, DELETED };
+    public enum ContractState { UNATTESTED, VALID, ATTESTATION_WITHDRAWN, DELETED }
 
     @Id
     @GeneratedValue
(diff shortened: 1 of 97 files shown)


common/src/main/java/gov/cms/ab2d/common/model/Contract.java
```diff
@@ -44,4 +48,35 @@
...
+            return false;   // No changes needed
+        }
+
+        if (hasAttestation) {
+            clearAttestation();
+            return true;
+        }
+
+        LocalDateTime ldt = LocalDate.parse(attestationDate, formatter).atStartOfDay();
+        markAttested(OffsetDateTime.of(ldt, ZoneOffset.UTC)); // todo: make sure offset is correct.
```
adaykin, line 79: I think we've said that these attestations were stored in EST since CMS is in Baltimore. AttestationReportProcessor.java uses EST

common/src/main/java/gov/cms/ab2d/common/model/Contract.java
```diff
@@ -44,4 +48,35 @@
...
+
+    public boolean hasAttestation() {
+        return attestedOn != null;
+    }
+
+    public void clearAttestation() {
+        attestedOn = null;
+    }
+
+    public void markAttested(OffsetDateTime attestedOnDate) {
```
adaykin, line 60: You can deleted this, there already is a setter with lombok

hpms/pom.xml
```diff
@@ -57,6 +74,12 @@
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
                 <version>${jacoco.version}</version>
+                <configuration>
+                    <excludes>
+                        <exclude>**/gov/cms/ab2d/hpms/hpmsapi/**</exclude>
```
adaykin, line 79: What's the reason for excluding these?

hpms/pom.xml
```diff
@@ -15,6 +15,14 @@
     </properties>
 
     <dependencies>
+        <dependency>
```
adaykin, line 18: These should be coming in from other projects I could be wrong but don't think you need them

hpms/src/main/java/gov/cms/ab2d/hpms/hmsapi/HPMSOrganizationInfo.java
```diff
@@ -0,0 +1,18 @@
+package gov.cms.ab2d.hpms.hmsapi;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
```
adaykin, line 7: Do you need all the constructors used here? Same with the other classes

hpms/src/main/resources/application.hpms.properties
```diff
@@ -0,0 +1 @@
+hpms.base.url=https://hpmsimpl.cms.gov
```
adaykin, line 1: This might be fine for now, eventually we can introduce an env variable.

hpms/src/main/java/gov/cms/ab2d/hpms/service/AttestationUpdaterService.java
```diff
@@ -0,0 +1,151 @@
...
+    List<Contract> addNewContracts(List<HPMSOrganizationInfo> newContracts) {
+        if (newContracts.isEmpty()) {
+            return new ArrayList<>();
+        }
+        return newContracts.stream().map(this::sponserAdd).collect(Collectors.toList());
+    }
+
+    private Contract sponserAdd(HPMSOrganizationInfo hpmsInfo) {
+        Sponsor sponsor = new Sponsor();
+        sponsor.setHpmsId(hpmsInfo.getParentOrgId());
```
adaykin, line 113: Add a call to sponsor.setLegalName

hpms/src/main/java/gov/cms/ab2d/hpms/service/AttestationUpdaterService.java
```diff
@@ -0,0 +1,151 @@
...
+        sponsor.setOrgName(hpmsInfo.getParentOrgName());
+        Sponsor savedSponser = sponsorRepository.save(sponsor);
+
+        Contract retContract = new Contract();
+        retContract.setContractName(hpmsInfo.getContractName());
+        retContract.setContractNumber(hpmsInfo.getContractId());
+        retContract.setSponsor(savedSponser);
```
adaykin, line 120: Based on what Kevin was saying I think this would be associating the contract with the parent org, contracts should be associated with the organization that they are apart of.

adaykin: You may have to search and see if the contract exists, and then get the sponsor based off of that

hpms/pom.xml
nflearl: Those are purely DTOs, nothing to test inside practically.

hpms/src/main/java/gov/cms/ab2d/hpms/service/AttestationUpdaterService.java
adaykin: Ok it sounds like the orgMarketingName is what may match our org name for the organization. You may want to confirm by comparing some prod data to what you'd get back from their API.

adaykin (changes requested):
See comments

hpms/pom.xml
nflearl: Fixed

common/src/main/java/gov/cms/ab2d/common/model/Contract.java
nflearl: Done - it was an artifact of a fancier implementation (with another persisted field).

common/src/main/java/gov/cms/ab2d/common/model/Contract.java
nflearl: Fixed

hpms/src/main/java/gov/cms/ab2d/hpms/hmsapi/HPMSOrganizationInfo.java
nflearl: With adding the AllArgsAnnotation, I was concerned that marshalling would break, hence I added the NoArgs too.  

I just did some testing and confirmed that NoArgs is not necessary and have removed it.

hpms/src/main/java/gov/cms/ab2d/hpms/service/AttestationUpdaterService.java
nflearl: What if the Contract does not exist?

adaykin: Then we'd want to create a new sponsor.

hpms/src/main/java/gov/cms/ab2d/hpms/service/AttestationUpdaterService.java
nflearl: Done.

hpms/src/main/java/gov/cms/ab2d/hpms/service/AttestationUpdaterService.java
nflearl: I key off of Contract, not Sponsor initially: contractRepo.findAll() => existingMap.

Where things are a bit rough, is when we detect a new contract.

hpms/pom.xml
adaykin: Ok I think that makes sense for the DTOs, what about HPMSFetcherImpl though?

nflearl: That class is purely HTTP wiring and is mocked out in the test world.  We don't have MOCK HPMS server spun up anywhere in order to simulate what it would return.

Once HPMS is live, we can revisit and potentially add some integration smoke tests with them.

common/src/main/resources/application.common.properties
```diff
@@ -1,3 +1,5 @@
 efs.mount=${AB2D_EFS_MOUNT:#{systemProperties['java.io.tmpdir']+'/jobdownloads'}}/
 
-audit.files.ttl.hours=24
\ No newline at end of file
+audit.files.ttl.hours=24
+
+hpms.base.url=https://hpmsimpl.cms.gov
```
lailasharshar, line 5: Make this an environment variable with this value as a default. Environment variables start with AB2D_

lailasharshar (changes requested):
Can we add last updated timestamps to the different elements so we can have the health check verify that the import is happening correctly?

common/src/main/resources/application.common.properties
nflearl: Done

nflearl:
In terms of getting back production data from their API, their API is not released to staging yet.  I will create an integration ticket. 
-  turn on the scheduler
- Verify mappings/that we create data correctly

In terms of adding updateDates, I'll create another ticket for that case and add createDate and updateDate to all our domain objects.  This just makes spelunking easier.

lailasharshar (approved):

adaykin (approved):

PR #430, opened 2020-09-02
Feature/ab2d 2013 travis dsl to jenkins dsl

wnyffenegger:
**JIRA Tickets:**

[AB2D-2013](https://jira.cms.gov/browse/AB2D-2013) - Travis DSL to Jenkins DSL

***Related Tickets***

[AB2D-2028](https://jira.cms.gov/browse/AB2D-2028) - Document Jenkins Setup Steps
 
### What Does This PR Do?

Make necessary changes to move from Travis CI to our own AWS hosted Jenkins server for CI and testing.

This PR does not make the actual switch from Jenkins to Travis, but does make all necessary code changes.

#### Jenkins Agent Considerations

* Agents are VMs that process CI jobs. Multiple jobs may be running concurrently on any given Jenkins agent
* Agents create a separate workspace (File structure) for each job
* Agents run jobs as the jenkins user and not as root so some sudo operations are off the table
* Agents have one Docker socket running that all CI jobs on that agent must share. This means all containers and container networks must be appropriately separated to prevent conflicts between jobs

#### Docker Compose Changes for E2E Test

* The docker-compose extends functionality to override the original `docker-compose.yml` with `docker-compose.jenkins.yml` including:
      * Forcing each Docker container to use the Jenkins user and group (prevents file permissions issues because Docker by default runs as root)
      * Forcing the api and worker instances to wait on the database to fully load (prevents race conditions)
      * Overriding the location of the fake efs mount directory used for testing via an environment variable (parameterizes the tests to the workspace the job has been constructed in)
* By default the database port is not exposed on the machine and is only available on the internal Docker network
* The api port is now randomly generated by the TestRunner at startup for the CI build to allow multiple jobs to run on the same agent

#### Dockerfile Changes

Changed Dockerfiles for API and Worker to accept environment variables specifying the minimum, maximum, and initial RAM used by the JVM. These environment variables represent percentages not absolute numbers, which is something we should change in the future. These environment variables are given defaults to prevent issues but can be overridden at runtime.

These changes should not impact production unless those specific environment variables are added.

For more information on how the extends functionality behaves see: https://docs.docker.com/compose/extends/

#### Test Runner Changes

* The test runner either uses a default api port `8443` or generates a random port in a CI environment
* The test runner now recognizes a new Environment (CI)
* The test runner now sets environment variables for the Docker compose like the api port
* The test runner init method has been broken up into several methods

#### Jenkins File Details

With Jenkins several interesting changes are possible that allow breaking up the build more:

* We can store credentials in Jenkins and load them as environment variables at run time
* We can break the build up into smaller steps more reliably
* We can log more data about what is happening during the build via scripts

#### Jenkins Stages

* Set environment variables for build
* Set tools necessary for build (Docker, Java, and Maven)
* Specify the agent this build can run on
* Make necessary resources and download necessary tools
* Build the Java code
* Run unit and integration tests for all projects except the e2e-test
* Run the e2e-test
* Evaluate code coverage
* Cleanup after the build

#### Jenkins Hacks

* We still have to generate temporary AWS Dev credentials during the build since permanent credentials are being phased out
* Many environment variables and credentials are provided just for the end to end test. The workspace for the job isn't determined until runtime meaning many of those parameters must be set at run time.

* This PR is for changes to testing infrastructure but a few changes are made to source code. Those changes need to be scrutinized.
* Any changes to the docker-compose structure that may not be stable long term
* Any remaining areas where running multiple builds in parallel on the same machine may cause conflicts
* Any handling of credentials by the Jenkins script that may be problematic. One thing that may be problematic is that in Jenkins we generate temporary credentials for AWS (valid 1 hr.) and those credentials are logged to the console by Jenkins.

To test the CI build 

1) Go to the "old" Jenkins agent on the AWS Management account (look up
2) Go to Jenkins -> AB2D Testing -> d-feature-ab2d-2013....
3) Hit "Build Now"
4) Come back in 17 minutes when the build is done and read the "Console Output" from the build

### Security Implications

### Submitter Checklist


api/src/main/java/gov/cms/ab2d/api/util/HealthCheck.java
```diff
@@ -30,8 +30,7 @@ public boolean healthy() {
                 DatabaseAvailable.isDbAvailable(dataSource) &&
                 // EFS Mount can be written to
                 FileSystemCheck.canWriteFile(efsMount, true) &&
-                // Can write to home directory
-                FileSystemCheck.canWriteFile(".", false) &&
```
adaykin, line 34: Why was this deleted?

wnyffenegger: So it breaks the CI build in a really inconvenient and difficult to handle way. It also doesn't make sense because you really shouldn't be writing to the working directory of the api container `usr/src/ab2d-api` where the JAR that the container runs is.

wnyffenegger: Is the deploy Dockerfile different from `ab2d/api/Dockerfile`? This is what I get when running `ls` on the `/usr/src/ab2d-api` directory in the container. If this is what is in production then we are publishing everything from the build as part of the docker image.

image

Jenkinsfile
```diff
@@ -0,0 +1,177 @@
...
+        OKTA_CLIENT_PASSWORD = credentials('OKTA_CLIENT_PASSWORD')
+        SECONDARY_USER_OKTA_CLIENT_ID = credentials('SECONDARY_USER_OKTA_CLIENT_ID')
+        SECONDARY_USER_OKTA_CLIENT_PASSWORD = credentials('SECONDARY_USER_OKTA_CLIENT_PASSWORD')
+
+        AB2D_HICN_HASH_PEPPER = credentials('HICN_HASH_PEPPER')
+        AB2D_HICN_HASH_ITER = 1000
+
+        AB2D_CLAIMS_SKIP_BILLABLE_PERIOD_CHECK = true
+        AB2D_OPT_OUT_JOB_SCHEDULE = "*/15 * * * * ?"
+        AB2D_OPT_OUT_ENABLED = true
```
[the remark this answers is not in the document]
wnyffenegger: It appears @adaykin that removing the optout properties causes the e2e tests to fail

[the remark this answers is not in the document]
adaykin: Which ones? Can you describe how they're failing?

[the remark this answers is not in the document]
wnyffenegger: The e2e-test fails here is the log for two jobs with the change: http://35.173.33.85:8080/job/ab2d-testing/job/feature%252Fab2d-2013-travis-dsl-to-jenkins-dsl/239/ and http://35.173.33.85:8080/job/ab2d-testing/job/feature%252Fab2d-2013-travis-dsl-to-jenkins-dsl/238/

bfd/src/test/resources/application.bfd.properties
```diff
@@ -1,6 +1,6 @@
 bfd.keystore.password=changeit
 bfd.keystore.location=/bb.keystore
-bfd.serverBaseUrl=http://localhost:8083/v1/fhir/
+#bfd.serverBaseUrl=http://localhost:8083/v1/fhir/
```
adaykin, line 3: Does this need to be commented out?

Jenkinsfile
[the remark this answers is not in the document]
wnyffenegger: No longer an issue since updates removing optout code.

nflearl (approved):

lhanekam (dismissed):
LGTM

docker-compose.yml
```diff
@@ -51,28 +65,36 @@ services:
...
       - AB2D_DB_PORT=5432
       - AB2D_DB_DATABASE=ab2d
       - AB2D_DB_SSL_MODE=allow
       - AB2D_DB_USER=ab2d
       - AB2D_DB_PASSWORD=ab2d
+      # Location to store files within container
       - AB2D_EFS_MOUNT=/opt/ab2d
-      - AB2D_EXECUTION_ENV=local
+      # LOCAL if running on person machine, CI if running on Jenkins agent VM
+      - AB2D_EXECUTION_ENV=${AB2D_EXECUTION_ENV}
```
lailasharshar, line 82: This is used for EventLogger. local is the only value that turns off kinesis logging so the value of ci would cause a bunch of error messages.

lailasharshar: AB2D_EXECUTION_ENV, that is

lailasharshar (changes requested):

nflearl (approved):

docker-compose.yml
wnyffenegger: Addressed. Hard coded to local in docker-compose now. Testing in Jenkins and Travis

lhanekam (approved):

lailasharshar (approved):

PR #432, opened 2020-09-09
Add Authentication Headers on the HPMS call

nflearl:
**JIRA Tickets:**

[AB2D-1969](https://jira.cms.gov/browse/AB2D-1969) - Add Authentication Headers on the HPMS call

***Related Tickets***

None
 
### What Does This PR Do?

This PR authenticates with HPMS and retrieves an authorization token for use in future HPMS API calls.

The token is cached for subsequent calls.

None

None

### Database Changes

None

### Limitations

Dependent upon HPMS to issue more permanent credentials.  The current credential is limited to the IMPL environment for the month of September 2020.

### Security Implications

### Submitter Checklist

- [x ] This PR includes any required documentation changes, (`README`, changelog / release notes, website).
- [ x] All tech debt and/or shortcomings introduced by this PR are detailed in `TODO` and/or `FIXME` comments.

### DIFF
--- a/common/src/main/resources/application.common.properties
+++ b/common/src/main/resources/application.common.properties
@@ -3,3 +3,4 @@ efs.mount=${AB2D_EFS_MOUNT:#{systemProperties['java.io.tmpdir']+'/jobdownloads'}
 audit.files.ttl.hours=24
 
 hpms.base.url=${AB2D_HPMS_URL:#{'https://hpmsimpl.cms.gov'}}
+hpms.auth.url=${AB2D_HPMS_AUTH_URL:#{'https://hpmsimpl.cms.gov/api/idm/oauth/token?ACS=wlVuThEThipRlBu37Pra'}}
--- a/hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
+++ b/hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
@@ -87,7 +87,7 @@ public class HPMSAuthServiceImpl implements HPMSAuthService {
         return tokenExpires;
     }
 
-    public void setTokenExpires(long tokenExpires) {
+    void setTokenExpires(long tokenExpires) {
         this.tokenExpires = tokenExpires;
     }
 }
--- a/hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
+++ b/hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
@@ -87,7 +87,7 @@ public class HPMSAuthServiceImpl implements HPMSAuthService {
         return tokenExpires;
     }
 
-    void setTokenExpires(long tokenExpires) {
-        this.tokenExpires = tokenExpires;
+    void clearTokenExpires() {
+        this.tokenExpires = 0;
     }
 }
--- a/hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
+++ b/hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
@@ -1,0 +1,93 @@
+package gov.cms.ab2d.hpms.service;
+
+import gov.cms.ab2d.hpms.hmsapi.HPMSAuthResponse;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+import org.springframework.stereotype.Service;
+import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.publisher.Flux;
+
+import java.time.Duration;
+
+import static org.springframework.http.HttpHeaders.*;
+
+@Service
+public class HPMSAuthServiceImpl implements HPMSAuthService {
+
+    @Value("${hpms.auth.url}")
+    private String authURL;
+
+    @Value("${HPMS_AUTH_KEY_ID}")
+    private String hpmsAPIKeyId;
+
+    @Value("${HPMS_AUTH_KEY_SECRET}")
+    private String hpmsSecret;
+
+    private volatile String authToken;
+
+    private volatile long tokenExpires;
+
+    @Override
+    public void buildAuthHeaders(HttpHeaders headers) {
+        headers.set("X-API-CONSUMER-ID", hpmsAPIKeyId);
+        headers.set(AUTHORIZATION, retrieveAuthToken());
+    }
+
+    private String retrieveAuthToken() {
+        final long currentTimestamp = System.currentTimeMillis();
+        if (authToken == null || currentTimestamp >= tokenExpires) {
+            refreshToken(currentTimestamp);
+        }
+
+        return authToken;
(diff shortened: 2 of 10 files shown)


hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Value("${HPMS_AUTH_KEY_ID}")
+    private String hpmsAPIKeyId;
+
+    @Value("${HPMS_AUTH_KEY_SECRET}")
+    private String hpmsSecret;
+
+    private volatile String authToken;
+
+    private volatile long tokenExpires;
```
adaykin, line 29: Why do you need volatile here?

hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public void buildAuthHeaders(HttpHeaders headers) {
+        headers.set("X-API-CONSUMER-ID", hpmsAPIKeyId);
```
adaykin, line 33: Can you explain how this class will be used? It doesn't look like anything is setup to call the retrieveAuthToken and refreshToken methods.

hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+
+    String getAuthToken() {
+        return authToken;
+    }
+
+    long getTokenExpires() {
+        return tokenExpires;
+    }
+
+    public void setTokenExpires(long tokenExpires) {
```
adaykin, line 90: Looks like this is only for testing purposes? If so add a @VisibleForTesting to it.

hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
nflearl: This service is effectively a singleton with potentially multiple callers coming in on multiple threads.   Using volatile at least guarantees a memory latch which insures that each thread has an opportunity to see the latest state which may be a newly minted token.  While not guaranteeing multiple token requests, it makes it far more likely that there won't be.

Otherwise, different threads may not see the common state which can cause churning.

hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+        tokenExpires = currentTimestamp + authResponse.getExpires() * 900;
+        authToken = authResponse.getAccessToken();
+    }
+
+    private String retrieveAuthRequestPayload() {
+        return String.format(AUTH_PAYLOAD_TEMPLATE, hpmsAPIKeyId, hpmsSecret);
+    }
+
+    // Text Blocks are still a preview feature in JDK 13 and 14, otherwise, it would really clean this up.
+    private static final String AUTH_PAYLOAD_TEMPLATE = "{\n" +
```
adaykin, line 74: It's a little more typesafe to do these as actual java objects and have them translated by an ObjectMapper or GSON but this works.

nflearl: If it was more complicated, then I would do something more sophisticated.  I just try not to pull in more libraries/objects unless their weight justifies their use.  For this case, I don't really care about all the other stuff, just need to insert secrets to get the token.

hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
adaykin: I suppose that would be a possibility. Might be a bit of an edge case since I think an admin request would be the only caller

hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
nflearl: Didn't mean to make it public originally, slipped by.  It is now package-private.  Do you still want the annotation for package-private methods?

hpms/src/test/resources/application.hpms.properties
```diff
@@ -9,4 +9,5 @@ spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
 spring.jpa.hibernate.ddl-auto=none
 spring.liquibase.enabled=true
 
-hpms.base.url=${AB2D_HPMS_URL:#{'http://localhost:8080'}}
\ No newline at end of file
+hpms.base.url=${AB2D_HPMS_URL:#{'http://localhost:8080'}}
+hpms.auth.url=${AB2D_HPMS_AUTH_URL:#{'https://hpmsimpl.cms.gov/api/idm/oauth/token?ACS=wlVuThEThipRlBu37Pra'}}
```
adaykin, line 13: You may need to add these to the docker-compose.yml so they make it in there when running from a deployment, you can see examples in there of how it's done in the different environment sections. Same with the HPMS_AUTH_KEY_ID key

common/src/main/resources/application.common.properties
```diff
@@ -2,4 +2,5 @@ efs.mount=${AB2D_EFS_MOUNT:#{systemProperties['java.io.tmpdir']+'/jobdownloads'}
 
 audit.files.ttl.hours=24
 
-hpms.base.url=${AB2D_HPMS_URL:#{'https://hpmsimpl.cms.gov'}}
\ No newline at end of file
+hpms.base.url=${AB2D_HPMS_URL:#{'https://hpmsimpl.cms.gov'}}
+hpms.auth.url=${AB2D_HPMS_AUTH_URL:#{'https://hpmsimpl.cms.gov/api/idm/oauth/token?ACS=wlVuThEThipRlBu37Pra'}}
```
adaykin, line 6: This is in both property files, does the HPMS not inherit from common? I'm not sure common is being used though?

lailasharshar (approved):

hpms/src/main/java/gov/cms/ab2d/hpms/service/HPMSAuthServiceImpl.java
nflearl: At a service level, I tend not to assume how its called.  Volatile is very cheap (execution-wise) and calls attention to the potential threading issues.

hpms/src/test/resources/application.hpms.properties
nflearl: Done by Will.

common/src/main/resources/application.common.properties
nflearl: Cleaned this up.

adaykin (dismissed):

lailasharshar (approved):

adaykin (approved):

PR #433, opened 2020-09-09
Feature/ab2d 2110 db schema updates for contract bene mapping

wnyffenegger:
**JIRA Tickets:**

[AB2D-2110](https://jira.cms.gov/browse/AB2D-2110) - DB Schema Updates

### Why
These schema updates change the abstractions used for bene-contract mappings to more efficiently store and work with. The other change is to move bene-contract mappings to use the same auditing and status mentality that EOB searches use now.

Key abstractions:

* A CoveragePeriod is a unit of work defined by a <contract id, month, and year>. Instead of storing these three numbers on every Coverage row (billions) we store them as a separate concept with a foreign key (bene_coverage_search table).
* A CoveragePeriod is essentially a membership period for a single contract.
* A CoverageSearchEvent is any change in the state of a search (submitted -> in progress -> etc.). This is supposed to be an audit trail of when jobs ran and what happened to those jobs. There is a foreign key relationship to a CoveragePeriod for every.
* Coverage now references both a CoveragePeriod and CoverageSearchEvent. The search specifies the membership information, the CoverageSearchEvent specifies when this Coverage information was pulled.

***Related Tickets***
 
### What Does This PR Do?

Pushes a large update to the database schema modifying and creating tables and constraints to support storing contract-bene mappings. Further changes are made to JPA Entities and Repositories to reflect the database table changes and support existing services.

* Any breaking database changes. In prod there is no data in the beneficiary or coverage table so I think this issue is probably limited in scope to issues caused by altering/adding constraints.
* Any issues caused by changing the JPA entities that won't be caught by unit tests
* Any design issues with the JPA repos that may cause performance issues in the future

### Database Changes

* Deleted beneficiary table
* Created bene_coverage_period table defining a triple <contract id, month, year> which defines a beneficiary search for a give contract. This table is the link to the contract table which the coverage and event_bene_coverage_search_status table rely on.
* Create event_bene_coverage_search_status_changes table that records the submission, progression, and result of all coverage searches. This table links back to the bene_coverage_search table to tie to a specific contract
* Rework coverage table to only contain beneficiary id instead of a foreign key, to contain a foreign key to a coverage search (<contract id, month, year> triple), and a foreign key to a specific coverage search event that wrote this coverage entry.

### Security Implications

### Submitter Checklist

- [] This PR includes any required documentation changes, (`README`, changelog / release notes, website).


common/src/main/java/gov/cms/ab2d/common/repository/CoverageRepository.java
```diff
@@ -1,46 +1,29 @@
...
 
 import java.util.List;
-import java.util.Optional;
 
 @Repository
 public interface CoverageRepository extends JpaRepository<Coverage, Long> {
 
-
-    @Query(" SELECT c.beneficiary.patientId " +
+    @Query(" SELECT c.beneficiaryId " +
```
nflearl, line 15: With the use of Spring data, I don't see why this query needs to be specifically written.

common/src/main/java/gov/cms/ab2d/common/model/CoverageSearch.java
```diff
@@ -0,0 +1,37 @@
...
+    @EqualsAndHashCode.Include
+    private Integer id;
+
+    @ManyToOne
+    @JoinColumn(name = "contract_id")
+    private Contract contract;
+
+    @Column
+    @NotNull
+    private Integer month;
```
nflearl, line 29: If we are @NotNull, then why not use int instead of Integer?  int can not be null.

common/src/main/java/gov/cms/ab2d/common/service/CacheServiceImpl.java
```diff
@@ -64,6 +76,19 @@ private Integer getMonth(ClearCoverageCacheRequest request) {
         return month;
     }
 
+    private Integer getYear(ClearCoverageCacheRequest request) {
+        final Integer year = request.getYear();
+
+        final int currentYear = OffsetDateTime.now(ZoneOffset.UTC).getYear();
```
nflearl, line 82: Others can confirm but we are using Eastern time as the standard.

common/src/test/resources/application.common.properties
```diff
@@ -9,6 +9,7 @@ spring.datasource.password=${DB_PASSWORD}
 spring.jpa.properties.hibernate.temp.use_jdbc_metadata_defaults=false
 spring.datasource.driver-class-name=org.postgresql.Driver
 spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
+spring.jpa.show-sql=true
```
nflearl, line 12: In my experience, generally this is not done in production all the time.  This is a lower environment setting or for a critical bug.

nflearl (changes requested):
The summary in the PR states the nuts and bolts of the what, but does not give a sense of the why.  The covering JIRA tickets have no detail either.

So, what is the goal here?  Why are these changes being made?

common/src/main/java/gov/cms/ab2d/common/model/CoverageSearch.java
wnyffenegger: Would not null annotation even apply to an integer?

common/src/main/java/gov/cms/ab2d/common/service/CacheServiceImpl.java
adaykin: For storing HPMS data we've been using EST, other data is stored as UTC, which is pretty standard

wnyffenegger:
> The summary in the PR states the nuts and bolts of the what, but does not give a sense of the why. The covering JIRA tickets have no detail either.
> 
> So, what is the goal here? Why are these changes being made?

Tried to answer that by updating the PR summary. Basically, we need to modify the database to treat coverage-bene mapping more like EOB search jobs are treated. We need the ability to audit when mapping data is modified, track the status of mapping jobs to prevent conflicts with EOB searches, and store coverage information more efficiently.

common/src/main/java/gov/cms/ab2d/common/repository/CoverageRepository.java
nflearl: Missed that this is a scalar query, not an object one.

nflearl (approved):

adaykin (approved):

nflearl (approved):

common/src/main/java/gov/cms/ab2d/common/service/CacheServiceImpl.java
wnyffenegger: @adaykin yes but I want to make sure I always have the maximum year it could be so at 12:00 AM January 1st, 2021 UTC it is still 2020 EST which could cause a problem if someone requests a job for January. I don't know if it matters but it seemed safest.

nflearl (dismissed):

lailasharshar (approved):

adaykin (approved):

common/src/main/java/gov/cms/ab2d/common/model/CoveragePeriod.java
```diff
@@ -0,0 +1,34 @@
...
+    private Contract contract;
+
+    @Column
+    private int month;
+
+    @Column
+    private int year;
+
+    @Enumerated(STRING)
+    private JobStatus status;
```
adaykin, line 33: What's jobstatus in here for?

wnyffenegger: A contract beneficiary mapping search is now a job in an almost equivalent sense to the Job class. A mapping search is submitted, in progress, successful/failed. We need to know this because we don't want to begin a regular EOB search job when the mapping is being updated.

adaykin (approved):

common/src/main/java/gov/cms/ab2d/common/model/Coverage.java
```diff
@@ -5,32 +5,28 @@
...
-import javax.persistence.Id;
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
+import javax.persistence.*;
 
 @Entity
 @Getter
 @Setter
-@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
-public class Coverage extends TimestampBase {
```
nflearl, line 19: Why was the extension of TimestampBase dropped?

common/src/test/java/gov/cms/ab2d/common/service/CacheServiceImplTest.java
```diff
@@ -58,17 +56,23 @@
     @BeforeEach
     void setUp() {
         coverageRepo.deleteAll();
+        coverageSearchEventRepo.deleteAll();
+        coverageSearchRepo.deleteAll();
```
nflearl, line 60: deleteAll() should be in the teardown().

nflearl (changes requested):
Fix or justify not extending from TimestampBase.

Move deleteAll() to teardown().

common/src/main/java/gov/cms/ab2d/common/model/Coverage.java
wnyffenegger: This is tied to a CoverageSearchEvent with a foreign key relationship. That event has a specific time on it. Since contract searches return all at once I think that time is fairly accurate. We will also have tens maybe hundreds of millions of these so adding those two columns adds a lot of data.

common/src/test/java/gov/cms/ab2d/common/service/CacheServiceImplTest.java
wnyffenegger: I'll fix that in the next PR coming up

common/src/main/java/gov/cms/ab2d/common/model/Coverage.java
wnyffenegger: @nflearl if you disagree I can add it back in but it just seems like more information than we necessarily need.

nflearl (approved):

PR #435, opened 2020-09-10
AB2D-2090: Adds API endpoints that are admin accessible for creating …

adaykin:
**JIRA Tickets:**

[AB2D-42](https://jira.cms.gov/browse/AB2D-2090) - Description

Adds API endpoints that are admin accessible for creating and deleting IPs. An endpoint also exists for retrieving IPs. Adds a special type to postgres, since the column is an INET type.
 
### What Does This PR Do?

Adds admin API endpoints for management of IPs (add, delete, get)

Whether data is correct in db, make sure failure cases covered for bad IPs, bad sponsors.

### Security Implications

### Submitter Checklist


api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIIPTests.java
```diff
@@ -0,0 +1,214 @@
...
+
+    private static final String IP_URL = "/ip";
+
+    @BeforeEach
+    public void setup() throws JwtVerificationException {
+        contractRepository.deleteAll();
+        jobRepository.deleteAll();
+        userRepository.deleteAll();
+        roleRepository.deleteAll();
+        sponsorRepository.deleteAll();
```
nflearl, line 79: Move cleanup to an @AfterEach method, this improves the chances of not interfering with other tests.

common/src/main/java/gov/cms/ab2d/common/model/PGInet.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (!(obj instanceof PGInet)) {
+            return false;
+        }
+        PGInet other = (PGInet) obj;
+        if (address == null) {
+            if (other.address != null) {
```
nflearl, line 47: Combine in single if statement to avoid nesting.

nflearl (changes requested):

api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIIPTests.java
wnyffenegger: Or just duplicate in an aftereach method

wnyffenegger (approved):

api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIIPTests.java
adaykin: Probably duplicating is the best for now, moving to aftereach just means it may not get cleaned up for this particular suite. Eventually there should be something that just deletes all of the data in each suite in the Before but it's not consistent now between test suites.

nflearl: I don't get why one would ever want to delete things in the before.  Stale data should be fixed where it is created, not in all other places affected by it being left around.

I also don't get how @AfterEach would fail to cleanup in the suite.  One could add the other form of @After, which works on a class level to do the final cleanup.

adaykin: You could pick either one, but it seems easier to do it in the before since there's already setup involved there. As long as every test does it in a before then no test would be affected by data. If there's data remaining when this suite starts then it would effect the current suite.

nflearl: Putting it in the before (in another example) caused other, unrelated tests to fail:  see the solution to https://jira.cms.gov/browse/AB2D-1951.

nflearl: See the checkin here: https://github.com/CMSgov/ab2d/commit/2baca1f153a5a49895c4746401841a77302ce41e

adaykin: Cleaning up and deleting from tables shouldn't cause others to fail unless they're not setting up data properly.

adaykin: Whichever test runs first is somewhat random along with other tests, so it's hard to predict if something else would have cleaned up at this point. Until it's standardized you can't really predict which one will go first

nflearl: My point is that another class should never have leftover data.  Each class ought to clean up after itself.

common/src/main/java/gov/cms/ab2d/common/model/PGInet.java
```diff
@@ -0,0 +1,63 @@
+package gov.cms.ab2d.common.model;
+
+import java.io.Serializable;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+@SuppressWarnings("serial")
+public class PGInet implements Serializable {
```
nflearl, line 8: Why is there a need for this wrapping class?  Why not use InetAddr directly?

You can still have the mapping class to perform the serialization.

api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIIPTests.java
```diff
@@ -0,0 +1,224 @@
...
+    @Autowired
+    private TestUtil testUtil;
+
+    private String token;
+
+    private static final String IP_URL = "/ip";
+
+    @BeforeEach
+    public void setup() throws JwtVerificationException {
+        contractRepository.deleteAll();
```
nflearl, line 76: As per the slack discussion, mark this block with a todo, something like
`// todo Delete me once proper teardown is put in place.`

lailasharshar (approved):

nflearl (changes requested):
Add a todo and respond to the question on the IP wrapper class.

common/src/main/java/gov/cms/ab2d/common/model/PGInet.java
adaykin: Hibernate needs the equals and hashcode, any implementation like this will do that https://vladmihalcea.com/postgresql-inet-type-hibernate/

nflearl (dismissed):

nflearl (approved):

wnyffenegger (approved):

PR #443, opened 2020-09-18
Feature/ab2d 2160 coverage service bulk operations

wnyffenegger:
**JIRA Tickets:**

[AB2D-2160](https://jira.cms.gov/browse/AB2D-2160) - Coverage Service Bulk Read and Writes 

***Related Tickets***

[AB2D-2146](https://jira.cms.gov/browse/AB2D-2146) - Initial coverage service implementation
[AB2D-2110](https://jira.cms.gov/browse/AB2D-2110) - Initial database schema creation

### Why

Coverage-beneficiary mapping information is stored in the `coverage` table. We are expecting to insert as many as 30 million rows into this table for each month and to retrieve all of those rows at least once a month. JPA does not provide adequate support for bulk inserts. Performance testing finds that it takes 4 minutes to insert 100,000 entries with all optimizations available to JPA.

Instead of JPA the insert and select operations are done with plain JDBC `PreparedStatement` logic which can insert one million records in a minute.

Likewise we need to retrieve coverage information. We use plain JDBC for this relying on *index only scans* (B-Tree searches under the hood) to allow searches to be processed almost entirely in memory. One quirk of this is that the search will only be in memory if metadata about the table is updated so a method for updating that metatadata `CoverageService.vacuumCoverage` has also been included.

To prove that the performance is realistic a PerformanceTesting class has been included which allows running millions of inserts and selects with many threads.

Finally, the one controversial thing here is moving us away from using SimpleDateFormat for dates to java.time.LocalDate. The DateRange class is modified to use the Java 8 time api instead of the deprecated api.

### What Does This PR Do?

* Provides efficient implementations for inserting coverage information into the database using plain JDBC.
* Provides an efficient implementation for retrieving data ranges of membership using JDBC.
* Abstracts beneficiary contract membership information into a series of date ranges per contract.
* Add disabled tests which can be used for manually performance testing the speed of inserts and deletes.
* Change the DateRange implementation to the Java 8 time api.
* Modifies already existing but not deployed liquibase changes concerning `modify_coverage_schema.sql`. These changes do not require another file as long as no deployments are done before this branch is merged.

* Instructions to run the performance test locally are included in the class.

### Database Changes

* Schema changes with `modify_coverage_schema.sql`
* Will lead to large volumes of data being pushed to the database

### Security Implications

* PR concerning 2146

### Submitter Checklist


wnyffenegger:
@sb-seanblackford how do I disable gitleaks scanner for the test classes. The test credential is only ever used on our local machines or in Jenkins and doesn't function in any AWS environment?

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
```diff
@@ -84,57 +103,162 @@ private CoverageSearchEvent getLastEvent(CoveragePeriod period) {
...
 
-        CoveragePeriod period = findCoveragePeriod(searchId);
+        // Make sure that coverage period and searchEvent actually exist in the database before inserting
+        findCoveragePeriod(periodId);
         CoverageSearchEvent searchEvent = findCoverageSearchEvent(searchEventId);
 
-        Collection<Coverage> coverages = beneficiaryIds.stream()
-                .map(bene -> buildCoverage(period, searchEvent, bene)).collect(toList());
+        int written = 0;
+        while (written < beneficiaryIds.size()) {
```
nflearl, line 113: This would be better as a for loop, all the information on one line ... including the increment later in the loop.

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
```diff
@@ -84,57 +103,162 @@ private CoverageSearchEvent getLastEvent(CoveragePeriod period) {
...
 
-        JobStatus status = getSearchStatus(searchId);
-        if (status == JobStatus.IN_PROGRESS) {
-            coverageRepo.saveAll(coverages);
-        } else {
-            throw new InvalidJobAccessException("cannot update coverages if coverage search is not in progress (current status " + status + " )");
+            try (Connection connection = dataSource.getConnection();
+                 PreparedStatement statement = connection.prepareStatement(INSERT)) {
+
+                int stopCondition = written + BATCH_INSERT_SIZE > beneficiaryIds.size() ? beneficiaryIds.size() : written + BATCH_INSERT_SIZE;
```
nflearl, line 118: Replace with Math.min()

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
```diff
@@ -84,57 +103,162 @@ private CoverageSearchEvent getLastEvent(CoveragePeriod period) {
...
+
+                for (int index = written; index < stopCondition; index++) {
```
nflearl, line 120: Having a loop surrounding a try with a loop within yields varying **levels of abstraction**.  My preference would be to refactor the contents of the try block to a private method that would be descriptively named aiding in readability.

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
```diff
@@ -84,57 +103,162 @@ private CoverageSearchEvent getLastEvent(CoveragePeriod period) {
...
+
+        // Look up a page of beneficiaries sorted by id
+        Page<String> beneficiaryPage = coverageRepo.findActiveBeneficiaryIds(coveragePeriods, PageRequest.of(pageNumber, pageSize));
+        List<String> beneficiaries = beneficiaryPage.getContent();
+
+        // Look up coverage information for BATCH_SELECT_SIZE beneficiaries at a time
+        // and populate into hash map
+        Map<String, List<Object[]>> coverageMemberships = new HashMap<>(beneficiaries.size());
+        int startIndex = 0;
+        while (startIndex < beneficiaries.size()) {
```
nflearl, line 185: I think this will read better in a for loop.

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
```diff
@@ -84,57 +103,162 @@ private CoverageSearchEvent getLastEvent(CoveragePeriod period) {
...
+        return coverageMemberships;
+    }
+
+    /**
+     * Summarize the coverage of one beneficiary for
+     */
+    private CoverageSummary summarizeCoverageMembership(Contract contract, Map.Entry<String, List<Object[]>> membershipInfo) {
+
+        String beneficiaryId = membershipInfo.getKey();
+        List<Object[]> membershipMonths = membershipInfo.getValue();
```
nflearl, line 204: What are the actual types in the Map?  String?  Integer?

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
```diff
@@ -84,57 +103,162 @@ private CoverageSearchEvent getLastEvent(CoveragePeriod period) {
...
+
+        if (membershipMonths.size() == 1) {
+            LocalDate start = fromRawResults(membershipMonths.get(0));
+            DateRange range = new DateRange(start, start.plusMonths(1));
+            return new CoverageSummary(beneficiaryId, contract, Collections.singletonList(range));
+        } else {
```
nflearl, line 210: With a return above this line, don't need the else which allows the rest of the code in the block to be outdented.

common/src/main/java/gov/cms/ab2d/common/model/CoverageSummary.java
```diff
@@ -0,0 +1,23 @@
...
+ * whatever date ranges are included.
+ */
+@Getter
+@ToString
+@AllArgsConstructor
+public class CoverageSummary {
+
+    private String beneficiaryId;
+    private Contract contract;
+    private List<DateRange> dateRanges;
```
nflearl, line 22: So, there can be gaps in the ranges, correct?

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
```diff
@@ -84,57 +103,162 @@ private CoverageSearchEvent getLastEvent(CoveragePeriod period) {
...
+            startIndex = endIndex;
+        }
+
+        return coverageMemberships;
+    }
+
+    /**
+     * Summarize the coverage of one beneficiary for
+     */
+    private CoverageSummary summarizeCoverageMembership(Contract contract, Map.Entry<String, List<Object[]>> membershipInfo) {
```
nflearl, line 201: This method seems to have quite a bit of complexity that I'm not convinced is needed.

Is the list of **membershipInfo** sorted?  Would it simplify processing if it was?  Since the list refers to months, maximum number is 12, correct?

If so, then performance over processing 12 items can't be a concern and clarity should be the goal.

common/src/test/java/gov/cms/ab2d/common/service/CacheServiceImplTest.java
```diff
@@ -117,7 +118,7 @@ void given_contractNumber_and_month_should_clear_cache() {
         cut.clearCache(request);
 
         CoveragePeriod coveragePeriod = coveragePeriodRepo.getByContractIdAndMonthAndYear(contract.getId(), january, YEAR);
-        final List<String> activePatientIds = coverageRepo.findActiveBeneficiaryIds(coveragePeriod);
+        final List<String> activePatientIds = coverageRepo.findActiveBeneficiaryIds(Collections.singletonList(coveragePeriod));
```
nflearl, line 121: Static imports are really nice here, especially since singletonList is the only method used out of Collections.

common/src/test/java/gov/cms/ab2d/common/service/PerformanceTestingCoverageOperations.java
```diff
@@ -0,0 +1,275 @@
...
+        ExecutorService executor = Executors.newFixedThreadPool(threads);
+
+        Random random = new Random();
+
+        List<Long> timing = new ArrayList<>(queries * threads);
+
+        Runnable select = () -> {
+            int i = 0;
+
+            while (i < queries) {
```
nflearl, line 169: Style-wise, this is a for loop.  Please use **idx** (or something slightly more verbose than **i**).. Searching for **i** in a code base is a real pain.  Searching for **idx** is much easier.

nflearl (changes requested):

common/src/main/java/gov/cms/ab2d/common/model/CoverageSummary.java
wnyffenegger: Yes there can be I am copying the existing DateRange class in the worker module and upgrading it to java.time because java.time is much simpler to work with. This is actually what I think the most controversial change will be. See the class here: filter/src/main/java/gov/cms/ab2d/filter/FilterOutByDate.java

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
wnyffenegger: Both months and years, you can wrap years.

common/src/main/java/gov/cms/ab2d/common/service/CoverageServiceImpl.java
wnyffenegger: No longer relevant converted to an actual class to make the code cleaner.

common/src/test/java/gov/cms/ab2d/common/service/PerformanceTestingCoverageOperations.java
```diff
@@ -0,0 +1,313 @@
...
+        }
+    }
+
+    private List<Long> insertData(CoverageSearchEvent inProgress, int dataPoints, int experiments, boolean erase) {
+        BeneficiaryIdSupplier supplier = new BeneficiaryIdSupplier();
+
+        List<Long> timings = new ArrayList<>();
+
+        int i = 0;
+        while (i < experiments) {
```
nflearl, line 273: Another for loop candidate.  Please use *idx* or some other variable instead of 1 letter ones.

common/src/test/java/gov/cms/ab2d/common/service/PerformanceTestingCoverageOperations.java
```diff
@@ -0,0 +1,313 @@
...
+        }
+        return timings;
+    }
+
+    private void conductBatchInsert(BeneficiaryIdSupplier supplier, long searchEventId, int dataPoints) {
+
+        int written = 0;
+        int chunkSize = 50000;
+
+        while (written < dataPoints) {
```
nflearl, line 294: Another for loop.

common/src/test/java/gov/cms/ab2d/common/service/PerformanceTestingCoverageOperations.java
```diff
@@ -0,0 +1,313 @@
...
+            timings.add(Duration.between(start, end).toMillis());
+            i++;
+        }
+        return timings;
+    }
+
+    private void conductBatchInsert(BeneficiaryIdSupplier supplier, long searchEventId, int dataPoints) {
+
+        int written = 0;
+        int chunkSize = 50000;
```
nflearl, line 292: Make a constant within the class.

nflearl (changes requested):
Still some unresolved nits.

common/src/test/java/gov/cms/ab2d/common/service/PerformanceTestingCoverageOperations.java
adaykin: Using i seems fine to me it's pretty universally used as a index all over the JDK, don't think we need to search for it, plus it's shorter to write

common/src/test/java/gov/cms/ab2d/common/service/PerformanceTestingCoverageOperations.java
wnyffenegger: Converted

common/src/test/java/gov/cms/ab2d/common/service/PerformanceTestingCoverageOperations.java
wnyffenegger: Converted

common/src/test/java/gov/cms/ab2d/common/service/PerformanceTestingCoverageOperations.java
wnyffenegger: Created

common/src/main/java/gov/cms/ab2d/common/service/CoverageRuntimeException.java
```diff
@@ -0,0 +1,12 @@
+package gov.cms.ab2d.common.service;
+
+public class CoverageRuntimeException extends RuntimeException {
```
nflearl, line 3: I did not find anywhere explicit usage of this call (in terms of catching this particular Exception).

It merely wraps RuntimeException and I don't think it pulls its weight.  In other words, I don't see the class providing any value.

nflearl (changes requested):
Much improved!  Only have 1 more item to go.

nflearl (approved):

adaykin (approved):

PR #445, opened 2020-09-21
Update gitleaks_github_actions.yml

sb-seanblackford:
Moving the whitelist file to an external repo so that changes don't require a PR within this project.

**JIRA Tickets:**

[AB2D-42](https://jira.cms.gov/browse/AB2D-42) - Description

***Related Tickets***

[AB2D-41](https://jira.cms.gov/browse/AB2D-41) - Reason it's related
 
### Security Implications

### Submitter Checklist

### DIFF
--- a/.github/workflows/gitleaks_github_actions.yml
+++ b/.github/workflows/gitleaks_github_actions.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       REPO: https://github.com/CMSgov/ab2d
-      REMOTE_EXCLUDES_URL: https://raw.githubusercontent.com/CMSgov/ab2d/master/.github/gitleaks.toml
+      REMOTE_EXCLUDES_URL: https://raw.githubusercontent.com/semanticbits/ab2d-gitleaks-automation/master/gitleaks.toml
       GITLEAKS_VERSION: v3.0.3
     steps:
     - name: Execute Gitleaks


lhanekam (commented):
LGTM

nflearl (approved):
Initially I was concerned that this was a potential end-around having security configuration outside the scope of normal AB2D purview but after a conversation with Lonnie, we are both comfortable in the change.

Changes to this file affect AB2D build process only, not actual runtime.  So if something breaking was introduced, it won't affect consumers of our service, just our ability to build and deploy.

One open question is if there is value in having settings like this in a global project (applicable to all of SemanticBits security concerns) instead of duplicating into additional repos on a per-project basis.

sb-seanblackford:
> 
> 
> Initially I was concerned that this was a potential end-around having security configuration outside the scope of normal AB2D purview but after a conversation with Lonnie, we are both comfortable in the change.
> 
> Changes to this file affect AB2D build process only, not actual runtime. So if something breaking was introduced, it won't affect consumers of our service, just our ability to build and deploy.
> 
> One open question is if there is value in having settings like this in a global project (applicable to all of SemanticBits security concerns) instead of duplicating into additional repos on a per-project basis.

Thanks! If this does become an issue in the future and we need more eyes on changes, we can certainly move this back to the repo to get more eyes on it.

While we do have a "baseline" for what is allowed in each repo, we've found that due to the differences in project coding standards, languages, and various other things that end up in the repo, it's cleaner to have a separate gitleaks file for each repo. Otherwise we might end up with some really large and unmaintainable toml files.

lhanekam (commented):
LGTM

lhanekam (approved):

PR #449, opened 2020-09-29
Bug/ab2d 2247 filter dates default timezone

wnyffenegger:
**JIRA Tickets:**

[AB2D-2247](https://jira.cms.gov/browse/AB2D-42) - FilterOutByDate default timezone

***Related Tickets***

[AB2D-2160](https://jira.cms.gov/browse/AB2D-2160) - Originally made LocalDate based DateRange that needs to be replaced
 
### What Does This PR Do?

Forces all date comparisons to be done in the America/New_York timezone. 

#### Why

Because no timezone was set all date comparisons are being done in a potentially unknown timezone. This change forces the usage of Eastern time to make performance more predictable. Setting a timezone can cause many other issues related to daylight savings time, the flexibility in the hour of the day etc.

#### Changes

* DateRange can only be constructed from the start of a month to the end of the same/other month instead of from random days
* DateRange remains inclusive but not longer uses SimpleDateFormat in the constructor, only accepts provided dates.
* DateRange constructor is private and a DateRange can only be constructed by using factory methods
* DateRange must be constructed using America/New_York timezone
* DateRange now explicitly sets milliseconds for times to catch dates on edge cases
* Tests regarding date range only test month ranges
* FilterOutByDate logic has now been moved to common library instead of filter
* Tests now use an explicitly defined range when the want to have all dates in range

Any issues where time ranges will cut off EOB results at the end of the range by accident. Any issues where the changes to how a `DateRange` is constructed may effect the interval of a date range in production.

### Security Implications

### Submitter Checklist


lailasharshar (approved):

adaykin (dismissed):

worker/src/test/java/gov/cms/ab2d/worker/TestUtil.java
```diff
@@ -0,0 +1,13 @@
+package gov.cms.ab2d.worker;
+
+import gov.cms.ab2d.common.util.FilterOutByDate;
+
+public final class TestUtil {
```
nflearl, line 5: What a horrid name for a class.

Unfortunately, the pattern exists elsewhere but is mitigated by being limited to the test arena.

common/src/main/java/gov/cms/ab2d/common/util/FilterOutByDate.java
```diff
@@ -288,6 +289,14 @@ static boolean withinDateRange(ExplanationOfBenefit ben, DateRange range) {
         if (start == null || end == null) {
             return false;
         }
+
+
+        // May 1st - June 1st
```
nflearl, line 294: This is a member in April?

This whole comment block seems disjointed, perhaps it is just a formatting issue but it makes it harder for me to understand what is going on here.  That's the last thing that a comment should do.

common/src/main/java/gov/cms/ab2d/common/util/FilterOutByDate.java
```diff
@@ -288,6 +289,14 @@ static boolean withinDateRange(ExplanationOfBenefit ben, DateRange range) {
...
+        // They are a member in April
+
+        // April 1st - April 30th
+
+        // DateRange April 1st <= x < May 1st
```
nflearl, line 299: This comment appears to tell a different story than the code.

common/src/main/java/gov/cms/ab2d/common/util/FilterOutByDate.java
```diff
@@ -266,7 +266,8 @@ static boolean afterDate(Date dateVal, ExplanationOfBenefit ben) throws ParseExc
         if (ben == null || ben.getBillablePeriod() == null || dateVal == null) {
             return false;
         }
-        Date attToUse = fullDateFormat.parse(shortDateFormat.format(dateVal) + " 00:00:00");
+
+        Date attToUse = fullDateFormat.parse(shortDateFormat.format(dateVal) + " 00:00:00:000");
         Period p = ben.getBillablePeriod();
```
nflearl, line 271: Why not just add a method on DateRange that takes a Period and returns a boolean if there is a match?

nflearl (changes requested):
TestUtil can stay as is.

The other cleanup adds to clarity.

worker/src/test/java/gov/cms/ab2d/worker/TestUtil.java
nflearl: No need to fix this one.

nflearl (approved):

lailasharshar (approved):

PR #488, opened 2020-10-28
Added post test cleanup hooks.

nflearl:
**JIRA Tickets:**

[AB2D-2481](https://jira.cms.gov/browse/AB2D-2481) - Added post-test cleanup hooks

***Related Tickets***
 
### What Does This PR Do?

Invokes cleanup code so the HPMS tests may be run in any order.

### Security Implications

### Submitter Checklist


lailasharshar (changes requested):
There has to be a better way than having test code in the main source branch.

wnyffenegger (commented):
I don't think this is the right solution because it adds test code to our source code. We aren't ever going to need to expire a token in our source code, is that correct? Instead we should just use ReflectionTestUtils and set those fields manually.
image

nflearl:
Took a different approach and submitted a new PR 490.

PR #498, opened 2020-11-03
Feature/ab2d 2529 eob patient mbi

lailasharshar:
**JIRA Tickets:**

[AB2D-42](https://jira.cms.gov/browse/AB2D-2529) - Add MBI id to extension in EOB

### What Does This PR Do?

The EOBs returned in the batch files do not contain the beneficiaries MBI. This is what the PDPs use to tie the data to their users. It is available in the mapping search if the proper header is specified. This can be used to add the data to the EOB extension data.

### Database Changes

Currently this is not saved in the database but in the next version, it will be.

### Security Implications

### Submitter Checklist


worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
```diff
@@ -227,6 +234,26 @@ private int processHandles(List<Future<EobSearchResult>> futureHandles, Progress
...
+            return;
+        }
+        // Get first EOB Bene ID
+        ExplanationOfBenefit eob = eobs.get(0);
+        String mbi = null;
+        String benId = getPatientIdFromEOB(eob);
+        if (benId != null) {
+            mbi = patients.get(benId).getMbiId();
+        }
+        Extension extension = createExtension(mbi);
```
wnyffenegger, line 248: If benId is null then we should be skipping this correct or is benId never null?

worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
```diff
@@ -227,6 +234,26 @@ private int processHandles(List<Future<EobSearchResult>> futureHandles, Progress
...
+        eobs.stream().forEach(e -> e.addExtension(extension));
+    }
+
+    Extension createExtension(String mbi) {
+        Identifier identifier = new Identifier().setSystem(MBI_ID).setValue(mbi);
+        return new Extension().setUrl(ID_EXT).setValue(identifier);
```
wnyffenegger, line 254: Do we want to move MBI_ID and the other ids to somewhere more central like a constants class?

wnyffenegger (approved):

worker/src/test/java/gov/cms/ab2d/worker/adapter/bluebutton/ContractBeneSearchTest.java
```diff
@@ -119,7 +123,7 @@ void GivenPatientActiveInMonth1And3ButNot2_ShouldReturnTwoRowsInDateRangesUnderC
         Bundle bundle1 = bundle.copy();
         // add 2nd patient
         List<BundleEntryComponent> entries = bundle1.getEntry();
-        entries.add(createBundleEntry("ccw_patient_001", YEAR));
+        entries.add(createBundleEntry("ccw_patient_001", "mbi1", YEAR));
```
nflearl, line 126: Refactor to constant both "ccw_patient_001" and "mbi1" as it is used multiple times.

nflearl (approved):

worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
lailasharshar: beneId should never be null, but if benId is null then there is no way to find the mbi id anyway why try and then have a null pointer exception.

worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
lailasharshar: Probably and I was thinking there may even be values for it in the hapi library but I'm not losing any sleep over this.

PR #515, opened 2020-11-17
Added BULK headers as requested by BFD.

nflearl:
**JIRA Tickets:**

[AB2D-2637](https://jira.cms.gov/browse/AB2D-42) - Added BULK headers as requested by BFD

***Related Tickets***
 
### What Does This PR Do?

It adds 2 headers as requested by BFD.

The job id is located far away from the code that actually builds the HTTP request and adds the headers, make sure that all cases are covered for setting the job id.

### Security Implications

### Submitter Checklist

### DIFF
--- a/bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClient.java
+++ b/bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClient.java
@@ -8,6 +8,10 @@ import java.time.OffsetDateTime;
 
 
 public interface BFDClient {
+    String BFD_HDR_BULK_CLIENTID = "BULK-CLIENTID";
+    String BFD_CLIENT_ID = "AB2D";
+    String BFD_HDR_BULK_JOBID = "BULK-JOBID";
+
     Bundle requestEOBFromServer(String patientID);
     Bundle requestEOBFromServer(String patientID, OffsetDateTime sinceTime);
     Bundle requestNextBundleFromServer(Bundle bundle);
@@ -24,5 +28,7 @@ public interface BFDClient {
     Bundle requestPartDEnrolleesFromServer(String contractNumber, int month);
 
     CapabilityStatement capabilityStatement();
+
+    ThreadLocal<String> BFD_BULK_JOB_ID = new ThreadLocal<>();
 }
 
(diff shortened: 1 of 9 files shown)


bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClient.java
```diff
@@ -24,5 +28,7 @@
     Bundle requestPartDEnrolleesFromServer(String contractNumber, int month);
 
     CapabilityStatement capabilityStatement();
+
+    ThreadLocal<String> BFD_BULK_JOB_ID = new ThreadLocal<>();
```
wnyffenegger, line 32: Do we really need this or can we just pass in the the value every time we make a call? Also nitpick would be to put this up stop with the other static fields

bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClientImpl.java
```diff
@@ -212,13 +214,24 @@ public Bundle requestNextBundleFromServer(Bundle bundle) {
...
 
+    private String getJobId() {
+        var jobId = BFDClient.BFD_BULK_JOB_ID.get();
+        if (jobId == null) {
+            log.warn("BFD Bulk Job Id not set: " + new Throwable());  // Capture the stack trace for diagnosis
+            jobId = "UNKNOWN";
+        }
+        return jobId;
+    }
+
```
wnyffenegger, line 234: If not using thread local then don't need this.

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
```diff
@@ -92,6 +92,7 @@
             } else {
                 sinceTime = request.getSinceTime();
             }
+            BFDClient.BFD_BULK_JOB_ID.set(request.getJob());
```
wnyffenegger, line 95: Why not add as parameter to method instead?

nflearl: I felt that adding parameters everywhere was too disruptive to the code base.  This is a cross-cutting concern, not actually needed by us for the call, more as an aid for BFD to do analysis of their traffic.

bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClient.java
nflearl: I felt that adding parameters everywhere was too disruptive to the code base. This is a cross-cutting concern, not actually needed by us for the call, more as an aid for BFD to do analysis of their traffic.

" put this up stop " => not sure what you mean.  I didn't want to put it in an Impl class since the callers of the Interface need access.

wnyffenegger (approved):

lailasharshar (approved):

PR #535, opened 2020-12-08
Feature/ab2d 2743 refactor coverage mapping processor

wnyffenegger:
**JIRA Tickets:**

[AB2D-2743](https://jira.cms.gov/browse/AB2D-2743) - Refactor Coverage Mapping Processor

***Related Tickets***

[AB2D-2742](https://jira.cms.gov/browse/AB2D-2742) - Making concurrency protection changes
 
### What Does This PR Do?

Makes the CoverageProcessor dumb again. Instead of handling both the logic for processing coverage mapping jobs and being responsible for scheduling those jobs, the CoverageProcessor now only handles the first.

The CoverageDriver handles discovering and handling coverage jobs. It is also where the locks

Whether the isolation is effective or if something is left over breaking the isolation.

### Security Implications

### Submitter Checklist


lailasharshar (approved):

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageProcessor.java
```diff
@@ -34,6 +20,26 @@
...
 
+    /**
+     * Queue a coverage period that failed or was cancelled for any reason
+     * @param coverageMapping coverage mapping that failed
+     * @param prioritize true if coverage mapping needs to be run first before other periods
+     */
+    void queueMapping(CoverageMapping coverageMapping, boolean prioritize);
+
+    /**
+     * Check if processor can except
```
nflearl, line 31: except => accept requests

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageDriverImpl.java
```diff
@@ -0,0 +1,190 @@
...
+        if (propertiesService.isInMaintenanceMode()) {
+            log.debug("waiting to execute queued coverage searches because api is in maintenance mode");
+        }
+
+        if (coverageProcessor.isProcessorBusy()) {
+            log.debug("not starting any new coverage mapping jobs because service is full.");
+            return;
+        }
+
+        while (!coverageProcessor.isProcessorBusy()) {
```
nflearl, line 167: better as a
```
do {
} while (!coverageProcessor.isProcessorBusy());
```
loop as that check has just been made and execution returned if queue is full.

lailasharshar (dismissed):

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageDriverImpl.java
```diff
@@ -0,0 +1,190 @@
...
+        } while (monthsInPast < config.getPastMonthsToUpdate());
+
+        return stalePeriods;
+    }
+
+    @Scheduled(fixedDelay = SIXTY_SECONDS, initialDelayString = "${coverage.update.initial.delay}")
+    public void loadMappingJob() {
+
+        if (propertiesService.isInMaintenanceMode()) {
+            log.debug("waiting to execute queued coverage searches because api is in maintenance mode");
```
nflearl, line 159: Why is there no *return* statement here?  Is this not the same case as busy?

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageDriverImpl.java
```diff
@@ -0,0 +1,190 @@
...
+            OffsetDateTime lastUpdatedAfter = dateTime.minusDays(daysInPast);
+
+            stalePeriods.addAll(coverageService.coveragePeriodNotUpdatedSince(month, year, lastUpdatedAfter));
+            monthsInPast++;
+        } while (monthsInPast < config.getPastMonthsToUpdate());
+
+        return stalePeriods;
+    }
+
+    @Scheduled(fixedDelay = SIXTY_SECONDS, initialDelayString = "${coverage.update.initial.delay}")
```
nflearl, line 155: I love the Scheduled annotation but by itself, will run on every machine in the cluster.  That is why there are all those funky Quartz configuration files lying around the project.

Is it ok if this runs on mulitple nodes?  If yes, please add a comment.

If not, need to guarantee that.

nflearl (changes requested):
Requested changes should be fairly simple to incorporate.

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageDriverImpl.java
wnyffenegger: It is okay that it runs on every machine. This is how we balance the work between nodes.

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageDriverImpl.java
wnyffenegger: There should be a return statement there.

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageDriverImpl.java
wnyffenegger: Can we hold off on addressing this one until the next follow on PR is brought up? Because there is an edge case where the processor might become busy. I also want to get rid of the loop I think because in practice it lets one worker claim all of the coverage work (until it runs out of threads) instead of balancing it between them.

lailasharshar (approved):

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageDriverImpl.java
```diff
@@ -0,0 +1,190 @@
...
+
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import static gov.cms.ab2d.common.util.DateUtil.getAB2DEpoch;
```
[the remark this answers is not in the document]
wnyffenegger: Made the change. It makes sense because the ZonedDateTime object is immutable anyway so there is no need to generate a new one each time it is called.

lailasharshar (dismissed):

nflearl (approved):

lailasharshar (approved):

PR #551, opened 2020-12-30
Feature/ab2d create slack alerts

lailasharshar:
**JIRA Tickets:**

[AB2D-2840](https://jira.cms.gov/browse/AB2D-2840) - Create event-driven design - Integrate with slack.

### What Does This PR Do?

This creates a slack logger which allows the application to integrate AB2D with slack through slack webhooks. There are potentially three different environment variables set up for slack, one for test, one for errors and one for new attesters discovered through HPMS. In the future there could be others, but this is sufficient for now. Because this is hard to test without directly connecting to slack, I added a test endpoint to worker to determine if a test message can be sent. Because we had talked about it earlier, I also added an endpoint to determine if BFD is up.
 
When deployed the webhook values for should be defined for the following environment variables:

AB2D_SLACK_HPMS_WEBHOOK
AB2D_SLACK_ERROR_WEBHOOK
AB2D_SLACK_TEST_WEBHOOK

By default https://slack.com is specified, but obviously will not work, although it will not cause the application not to work. Initially, I imagine the values for all of those variables will be the same.

### Security Implications

- [X] This PR stores or transmits data that was not stored or transmitted before

This application transmits data to Slack for the purpose of notifying AB2D of new attesters, any errors (currently that BFD is down) or test messages to verify that the integration is working.

### Submitter Checklist


nflearl (changes requested):
Generate events of attestation changes too - i.e. gaining attestation or having it revoked.

common/pom.xml
```diff
@@ -82,7 +82,11 @@
             <version>${mockserver.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
```
wnyffenegger, line 85: Any specific reason for this okhttp3? We have so many http clients in our code it is hard to keep track of. Also shouldn't we set the version in the parent pom and then import it?

eventlogger/src/main/resources/application.eventlogger.properties
```diff
@@ -1 +1,6 @@
-execution.env=${AB2D_EXECUTION_ENV:#{'local'}}
\ No newline at end of file
+execution.env=${AB2D_EXECUTION_ENV:#{'local'}}
+
+logger.slack.hpms.url=${AB2D_SLACK_HPMS_WEBHOOK:#{'https://slack.com'}}
+logger.slack.error.url=${AB2D_SLACK_ERROR_WEBHOOK:#{'https://slack.com'}}
+logger.slack.test.url=${AB2D_SLACK_TEST_WEBHOOK:#{'https://slack.com'}}
```
wnyffenegger, line 5: Concerning the webhook, aren't we going to need multiple webhooks? CMS has expressed interest in us sending events to CMS slack as well as SB slack?

eventlogger/pom.xml
```diff
@@ -100,6 +100,16 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.slack.api</groupId>
+            <artifactId>slack-api-client</artifactId>
+            <version>1.5.0-M1</version>
```
wnyffenegger, line 106: Is this an actual release or a pre-release that we should avoid?

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
```diff
@@ -73,6 +77,7 @@ private void updateMaintenanceStatus(Status status, String statusString) {
         PropertiesDTO propertiesDTO = new PropertiesDTO();
         propertiesDTO.setKey(MAINTENANCE_MODE);
         propertiesDTO.setValue(statusString);
+        slackLogger.logErrorMsg("Maintenance Mode status: " + statusString);
```
wnyffenegger, line 80: How often is this run? Might want to make sure this doesn't become spam?

wnyffenegger (approved):

wnyffenegger (commented):
Questions about okhttp and also the Slack client models. How do we support multiple slack workspaces, etc.?

common/pom.xml
lailasharshar: It's required by the slack client library and there is another version of the library that is used in other places so had conflicts.

eventlogger/src/main/resources/application.eventlogger.properties
lailasharshar: You mean for the same event? I created multiple variables for different types of events and currently, they are just sending them to the same one anyway. If it comes up, it shouldn't be too hard to accommodate.

eventlogger/pom.xml
lailasharshar: I just chose the latest one in mavenrepository.com. There really isn't much too it. We are only using one method call.

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
lailasharshar: Only when there is a change to the maintenance mode which doesn't happen unless it receives the same response more than x number of times. I guess we'll find out.

lailasharshar:
@wnyffenegger - There aren't different slack "workspaces". When you create a web hook, you specify a channel and are then given a URL which is considered a secret because that's all you need to post on that that channel. Anyone can use that URL and the channel can be owned by anyone which is why I created environment variables for them.

nflearl (approved):

wnyffenegger (approved):

PR #553, opened 2020-12-31
Techdebt/ab2d 2890 common test cleanup

wnyffenegger:
**JIRA Tickets:**

[AB2D-2890](https://jira.cms.gov/browse/AB2D-2890) - Common Test Cleanup
 
### What Does This PR Do?

Creates a mechanism for registering and deleting domain model objects commonly added to the database for unit testing.

Many of these domain objects are placed into tables that cannot be wiped because they initially contain standard sets of information. So this mechanism only deletes objects created during tests.

The mechanism is added to the DataSetup class and requires all domain objects to be registered with the DataSetup class by each test class. Additionally, each test class must implement an AfterEach method that calls DataSetup.cleanup to complete deletion of those objects between tests.

Any potential side effects that might cause issues in other modules like api and worker.

### Security Implications

### Submitter Checklist

### DIFF
--- a/common/src/test/java/gov/cms/ab2d/common/model/CreateUpdateTimestampTest.java
+++ b/common/src/test/java/gov/cms/ab2d/common/model/CreateUpdateTimestampTest.java
@@ -33,7 +33,7 @@ class CreateUpdateTimestampTest {
 
     @AfterEach
     public void after() {
-
+        dataSetup.cleanup();
     }
 
     @Test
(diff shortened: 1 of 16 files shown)


common/src/test/java/gov/cms/ab2d/common/model/CreateUpdateTimestampTest.java
```diff
@@ -23,9 +25,17 @@
...
 
+    @Autowired
+    private DataSetup dataSetup;
+
     @Autowired
     private ContractRepository contractRepository;
 
+    @AfterEach
+    public void after() {
+
```
nflearl, line 36: Looks like it is missing a call to 
`dataSetup.cleanup()`

nflearl (changes requested):
The only reason that changes are requested is for the comment in CreateUpdateTimestampTest.java.

As discussed, this PR contains 2 very different sets of changes:

1) Small Syntatic/Style changes throughout a large number of files.
	For example, removing public from test classes and methods, reworking of imports, adding static imports.
2) Significant course change in a small number of files.  This is the actual motivating change for the PR.

In the future, please try to separate into 2 different PRs.  The first one affecting many files with small changes can be quickly approved making it much easier to dig into the substantive changes.

lailasharshar (approved):

nflearl (approved):

PR #566, opened 2021-01-07
Techdebt/ab2d 2921 api test cleanup

nflearl:
**JIRA Tickets:**

[AB2D-42](https://jira.cms.gov/browse/AB2D-2921) - API Module Cleanup

***Related Tickets***
 
### What Does This PR Do?

Refactors how data is cleaned up, insuring that every test cleans up after itself.  All the repo.deletAll() calls are removed.

The use of contexts with liquidbase.  I have set the context to "none" for code that runs in real environment and "test" for the test harnesses.  If this is not set, then all liquibase commands are run.  If it is set, the all liquibase commands that don't have a context will run and only those with a matching context will run.

truncate-fake-data.sql does truncates on tables in test environments.  This could be a risk if the liquidbase context is not correctly set in a real environment.  An alternative and more wordy approach, would be to write the specific delete statements that are necessary for each table.

Since jobs are created as a side-effect in a lot of cases, simply added a query and add the job to the cleanup set.

Changed data type in DataSetup to a Set from a list so that if there are inadvertant double adds of the same object, then it does not matter.  There are some cases where jobs are explicitly specified mixed in with implicit test cases.

### Security Implications

### Submitter Checklist

### DIFF
--- a/api/src/test/java/gov/cms/ab2d/api/controller/MaintenanceModeAPITests.java
+++ b/api/src/test/java/gov/cms/ab2d/api/controller/MaintenanceModeAPITests.java
@@ -4,6 +4,7 @@ import gov.cms.ab2d.api.SpringBootApp;
 import gov.cms.ab2d.common.dto.PropertiesDTO;
 import gov.cms.ab2d.common.service.PropertiesService;
 import gov.cms.ab2d.common.util.AB2DPostgresqlContainer;
+import gov.cms.ab2d.eventlogger.reports.sql.DoAll;
 import org.hamcrest.core.Is;
 import org.junit.jupiter.api.*;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -38,6 +39,9 @@ public class MaintenanceModeAPITests {
     @Autowired
     private PropertiesService propertiesService;
 
+    @Autowired
+    DoAll doAll;
+
     @AfterEach
     void tearDown() {
         PropertiesDTO propertiesDTO = new PropertiesDTO();
@@ -45,6 +49,8 @@ public class MaintenanceModeAPITests {
         propertiesDTO.setValue("false");
 
         propertiesService.updateProperties(Collections.singletonList(propertiesDTO));
+
+        doAll.delete();
     }
 
     @Test
--- a/common/src/main/resources/db/changelog/test_only/truncate-fake-data.sql
+++ b/common/src/main/resources/db/changelog/test_only/truncate-fake-data.sql
@@ -1,8 +1,9 @@
 --liquibase formatted sql
 --  -------------------------------------------------------------------------------------------------------------------
 
---changeset AB2D-2921 failOnError:true context:test
+--changeset earl_nolan:AB2D-2921 failOnError:true context:test
 truncate user_role;
 truncate user_account cascade;
 truncate role cascade;
+truncate contract cascade;
 
--- a/common/src/test/resources/application.common.properties
+++ b/common/src/test/resources/application.common.properties
@@ -12,5 +12,6 @@ spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
 
 spring.jpa.hibernate.ddl-auto=none
 spring.liquibase.enabled=true
+spring.liquibase.contexts=none
 
 logging.level.liquibase=ERROR
--- a/api/src/main/resources/application.properties
+++ b/api/src/main/resources/application.properties
@@ -7,6 +7,7 @@ spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
 
 spring.jpa.hibernate.ddl-auto=none
 spring.integration.jdbc.initialize-schema=always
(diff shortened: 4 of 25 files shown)


api/src/test/java/gov/cms/ab2d/api/controller/MaintenanceModeAPITests.java
```diff
@@ -38,13 +39,18 @@
...
+    @Autowired
+    DoAll doAll;
+
     @AfterEach
     void tearDown() {
         PropertiesDTO propertiesDTO = new PropertiesDTO();
         propertiesDTO.setKey(MAINTENANCE_MODE);
         propertiesDTO.setValue("false");
 
         propertiesService.updateProperties(Collections.singletonList(propertiesDTO));
```
lailasharshar, line 51: Does this not log to an event table at any point?

common/src/test/resources/application.common.properties
```diff
@@ -12,5 +12,6 @@ spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
 
 spring.jpa.hibernate.ddl-auto=none
 spring.liquibase.enabled=true
+spring.liquibase.contexts=none
```
lailasharshar, line 15: Shouldn't this be "test" instead of "none"?

common/src/main/resources/db/changelog/test_only/truncate-fake-data.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
```
lailasharshar, line 1: When/how does this get called?

lailasharshar (changes requested):

common/src/test/resources/application.common.properties
nflearl: See "What Reviewers should Watch For" in the PR comments.

api/src/test/java/gov/cms/ab2d/api/controller/MaintenanceModeAPITests.java
nflearl: Virtually all the mockmvc calls result in events being logged.  They can also commonly result in jobs being created.

common/src/main/resources/db/changelog/test_only/truncate-fake-data.sql
nflearl: Upon setup of the Postgres SQL Container.

common/src/test/resources/application.common.properties
lailasharshar: I'm still confused. This properties files is in src/test, not src/main so shouldn't it be test according to the explanation?

common/src/main/resources/db/changelog/test_only/truncate-fake-data.sql
```diff
@@ -0,0 +1,9 @@
...
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset earl_nolan:AB2D-2921 failOnError:true context:test
+truncate user_role;
+truncate user_account cascade;
+truncate role cascade;
+truncate contract cascade;
+
```
lhanekam-sb, line 9: If you want the index to reset to 1 for the tables, you can do this (for example):

TRUNCATE user_account RESTART IDENTITY CASCADE;

lhanekam-sb (approved):

wnyffenegger (commented):
I think I agree with your concerns on using Truncate. I would rather explicitly delete the data that needs to be deleted. However, would a truncate in prod actually work? I assume any violation of constraints would block truncating.

common/src/main/resources/db/changelog/test_only/truncate-fake-data.sql
nflearl: Good to know.  I don't think it really matters for this case.

common/src/test/resources/application.common.properties
nflearl: This will be another ticket to switch to test.  I am focused only on the API module for this PR.

nflearl:
These changes do not apply to prod.  We would have to create a new liquibase script to cleanup the prod and prod-validation environments.  Even though initially, I thought of having this apply elsewhere; for now, I am limiting it just to the auto-test environments.

wnyffenegger:
> These changes do not apply to prod. We would have to create a new liquibase script to cleanup the prod and prod-validation environments. Even though initially, I thought of having this apply elsewhere; for now, I am limiting it just to the auto-test environments.

@nflearl 

I know but I'm saying if somehow someone accidentally applied them to prod. Also do we know if these changes will be registered in liquibase in prod as a changeset even if it isn't applied?

wnyffenegger (approved):
Can we explicitly write out the work that we need to do for these other modules so it doesn't get lost in everything coming our way?

lhanekam-sb (approved):

lailasharshar (approved):

PR #613, opened 2021-02-16
move to new patient endpoint which includes a year parameter

sb-wnyffenegger:
**JIRA Tickets:**

[AB2D-3249](https://jira.cms.gov/browse/AB2D-3249) - Enrollment Year Parameter
 
### What Does This PR Do?

Change the BFD client to use year in addition to month as a parameter when getting enrollment.

The change includes detecting when to use a specific year (0003) when operating on sandbox data.

### What Should Reviewers Watch For?

Whether the parameters are correctly specified according to FHIR and whether

### Usage/Deployment Instructions

To test against BFD I have included an integration test class `BFDClientIntegrationTest` which is disabled by default it. To run it set the environment loaded by the test related to protected keystores.

### Security Implications

- [ ] This PR adds new software dependencies
- [ ] This PR modifies or invalidates our security controls
- [ ] This PR stores or transmits data that was not stored or transmitted before

- [ ] This PR requires additional review of its security implications for other reasons

### Submitter Checklist

- [ ] This PR includes any required documentation changes, (`README`, changelog / release notes, website).
- [ ] All tech debt and/or shortcomings introduced by this PR are detailed in `TODO` and/or `FIXME` comments.
- [ ] Code checked for PHI/PII exposure

PR #614, opened 2021-02-16
Enhancement/ab2 d 2997 integrate r4 with bfd

sharshar-sb:
**JIRA Tickets:**

[AB2D-2997](https://jira.cms.gov/browse/AB2D-2997) - Support for FHIR R4 BFD search
 
### What Does This PR Do?

Adds the code to perform a FHIR R4 search for both Patient & ExplanationOfBenefit endpoints

### Limitations

This does not currently add any v2 endpoints to AB2D or include any documentation changes.

### Security Implications

### Submitter Checklist


bfd/src/test/java/gov/cms/ab2d/bfd/client/BlueButtonClientTestSTU3.java
```diff
@@ -83,25 +77,22 @@
...
     public static class TestConfig {
 
-        @Autowired
-        private FhirContext fhirContext;
-
         @Bean
         @Primary
         public IParser testBeanDefinition() {
-            return fhirContext.newXmlParser();
+            return FhirContext.forDstu3().newXmlParser();
```
sb-wnyffenegger, line 83: Why still an xml parser here?

bfd/src/test/java/gov/cms/ab2d/bfd/client/BlueButtonClientTestSTU3.java
```diff
@@ -110,72 +101,64 @@ public static void setupBFDClient() throws IOException {
         );
 
         for (String patientId : TEST_PATIENT_IDS) {
-            createMockServerExpectation(
+            MockUtils.createMockServerExpectation(
                     "/v1/fhir/Patient/" + patientId,
                     HttpStatus.SC_OK,
-                    getRawXML(SAMPLE_PATIENT_PATH_PREFIX + patientId + ".xml"),
+                    getRawJson(SAMPLE_PATIENT_PATH_PREFIX + patientId + ".json"),
```
sb-wnyffenegger, line 107: A little confused about why this works when the xml parser is specified above.

bfd/src/test/java/gov/cms/ab2d/bfd/client/ConvertXmlToJson.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.commons.io.FileUtils;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.MissingResourceException;
+
+public class ConvertXmlToJson {
```
sb-wnyffenegger, line 15: Maybe some documentation on when to run this and what it is for? I understand based on the context that this is how you convert sample data to JSON but...

filter/src/main/java/gov/cms/ab2d/filter/ExplanationOfBenefitTrimmerR4.java
```diff
@@ -2,108 +2,201 @@
...
     }
 
+    /**
+     * Copy data from the old EOB to the new EOB. Had to do it piecemeal because
+     * ExplanationOfBenefit.copy() resulted in a stack overvlow
+     *
+     * @param benefit - the EOB to copy necessary data
+     * @return the new object
+     */
+    private static ExplanationOfBenefit copyData(ExplanationOfBenefit benefit) {
```
sb-wnyffenegger, line 142: Do we have tests that confirm this is equivalent to the old method? Like the two objects serialize to the same value. I am worried about a regression in here.

filter/src/main/java/gov/cms/ab2d/filter/ExplanationOfBenefitTrimmerR4.java
```diff
@@ -2,108 +2,201 @@
...
-    public static IBaseResource getBenefit(IBaseResource b) {
-        ExplanationOfBenefit benefit = (ExplanationOfBenefit) b;
+    public static IBaseResource getBenefit(IBaseResource resource) {
+        ExplanationOfBenefit benefit = (ExplanationOfBenefit) resource;
         if (benefit == null) {
             return null;
         }
         // Copy it so we don't destroy the original
-        ExplanationOfBenefit newBenefit = benefit.copy();
+        ExplanationOfBenefit newBenefit = copyData(benefit);
```
sb-wnyffenegger, line 128: Why the copy in the first place? It isn't like we need the original for anything. Just seems like an opportunity to miss something and create a bug.

worker/src/main/java/gov/cms/ab2d/worker/HealthAPI.java
```diff
@@ -40,7 +43,12 @@
     @GetMapping(VALIDATE_BFD_ENDPOINT)
     public ResponseEntity<Void> isBfdUp() {
         try {
-            bfdClient.capabilityStatement();
+            bfdClient.capabilityStatement(Versions.FhirVersions.STU3);
```
sb-wnyffenegger, line 46: If this ever fails don't you want to send a message specific to STU3?

sb-wnyffenegger: Also will there ever be a case where one api version is down and the other is up, both are hosted on the same Java Runtime.

worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
```diff
@@ -232,7 +232,7 @@ private int processHandles(List<Future<EobSearchResult>> futureHandles, Progress
     }
 
     private int writeOutResource(List<IBaseResource> eobs, StreamHelper helper, Versions.FhirVersions version) {
-        var jsonParser = fhirContext.newJsonParser();
+        var jsonParser = Versions.getContextFromVersion(version).newJsonParser();
```
sb-wnyffenegger, line 235: Is there any way to create these JSON parsers once? Creating them thousands of times doesn't make sense. Why not just create a parser for each version in the Versions class and pick the parser based on FhirVersion? If the json parsers are thread safe then this shouldn't be an issue.

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
```diff
@@ -27,11 +29,11 @@
 
     // Nothing else should be calling this component except for the scheduler, so keep
     // state here
-    private int consecutiveSuccesses = 0;
+    private Map<Versions.FhirVersions, Integer> consecutiveSuccesses = new HashMap<>();
```
sb-wnyffenegger, line 32: Is there really any difference between one FHIR version being up and another FHIR version? BFD builds them in the same JAR and runs them as a single monolith.

bfd/src/test/java/gov/cms/ab2d/bfd/client/R4Test.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.Set;
+
+import static org.junit.Assert.fail;
+
+/**
+ * Class used to test against the real R4 BFD server (you need to change values in properties file to get it to work)
+ */
+@SpringBootTest(classes = SpringBootApp.class)
+@TestPropertySource(locations = "/application.bfd.properties")
+public class R4Test {
```
sb-wnyffenegger, line 31: Did this class sneak in because none of the tests are enabled and its presence doesn't really make sense.

sb-wnyffenegger (commented):
Couple comments:

1. There are lots of things being created more often than they need to be (FhirContext, json parsers, http clients). Is there any way to cut down on that?
2. The usage of UrlResolver really bugs me because I don't see the point. Why not just have a  config class with a conditional based on FhirVersion and pull those variables using @Value once?
3. Using a static port with MockServer means we can only run one build on a Jenkins agent at a time. This isn't a big deal now but could be in the future when we break things apart into smaller components.

sb-wnyffenegger:
@sharshar-sb 

1. Does CoverageMappingCallable need to know about STU3 vs. R4? We are using it to pull patient enrollment internally and save it. We aren't reporting Patient objects at all. So I don't see why it matters whether we hit STU3 or R4.
2. Do we need to work on e2e-tests for this to confirm it works against sandbox? I have a manual integration test in my branch that you might be able to pattern off of [BFDClientIntegrationTest](https://github.com/CMSgov/ab2d/pull/613/commits/fba311aefa9fdf7637ded29c0dd44a87b358fc37#diff-cc85b33aa59032f9b4691d49ccd3d9e3aeb19882636faad40d6101d769cdf373)

filter/src/main/java/gov/cms/ab2d/filter/ExplanationOfBenefitTrimmerR4.java
sharshar-sb: I'd rather miss adding data than miss removing it. This way you actively copy data that you want to opt in. I should really change it for the STU3 Trimmer.

worker/src/main/java/gov/cms/ab2d/worker/HealthAPI.java
sharshar-sb: Sure, in theory.

bfd/src/test/java/gov/cms/ab2d/bfd/client/R4Test.java
sharshar-sb: It did.

sharshar-sb:
In response to @sb-wnyffenegger comments:
1. It needs to make a choice. Right now, I use STU3 but we will probably change that to R4 in the future.
2. We do, however, we still need to implement our v2 API endpoints before we can do that.

worker/src/main/java/gov/cms/ab2d/worker/bfdhealthcheck/BFDHealthCheck.java
sharshar-sb: That may not always be true.

bfd/src/main/java/gov/cms/ab2d/bfd/client/FhirBfdServer.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Creates the rest client by passing in the Http Client and the URL of the endpoint
+     *
+     * @param httpClient - the client
+     * @param url - the URL of the BFD endpoint
+     * @return
+     */
+    public IGenericClient bfdFhirRestClient(HttpClient httpClient, String url) {
```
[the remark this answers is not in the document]
sharshar-sb: It's called once for each version and saved in a HashMap in BFDClientImpl. The next time someone wants a client, it pulls it from the HashMap.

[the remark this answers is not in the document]
sb-wnyffenegger: Gotcha, I guess that isn't a problem.

bfd/src/test/java/gov/cms/ab2d/bfd/client/BlueButtonClientTestR4.java
```diff
@@ -0,0 +1,154 @@
...
+    private static final String CONTRACT = "Z0012";
+
+    @Autowired
+    private BFDClient bbc;
+
+    private static ClientAndServer mockServer;
+
+    @BeforeAll
+    public static void setupBFDClient() throws IOException {
+        mockServer = ClientAndServer.startClientAndServer(MOCK_SERVER_PORT);
```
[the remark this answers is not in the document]
sharshar-sb: Another product of not injecting the URL. I'll see what I can do.

bfd/src/test/java/gov/cms/ab2d/bfd/client/BlueButtonClientTestSTU3.java
sharshar-sb: Because it doesn't use the above one, it uses the one created in the BFD search client. That's why I changed it. I would have had to support both JSON and XML in the codebase for this to work and I didn't see the reason to add a bunch of unneeded functionality just so the test would run.

bfd/src/test/java/gov/cms/ab2d/bfd/client/ConvertXmlToJson.java
sharshar-sb: It's how I converted the tests so they would work with JSON instead of XML. It's not needed for a test but I thought it could be useful in the future.

filter/src/main/java/gov/cms/ab2d/filter/ExplanationOfBenefitTrimmerR4.java
sharshar-sb: This is for R4. There are no regressions yet. This will spell out in the e2e tests when we implement the v2 endpoints.

worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
sharshar-sb: Yes, it shouldn't be here.

nflearl (changes requested):
FhirVersion should be it's own top level class and most (if not all) the static methods in Version should be instance methods in the enum.  Enums can be very powerful and can be used to delegate behavior directly verses having mapping tables or switch statements.

Example class for one of the methods:
```
package gov.cms.ab2d.fhir;

public enum FhirVersion {

    STU3() {
        public String getClassLocation() {
            return "org.hl7.fhir.dstu3.model";
        }
    },

    R4() {
        public String getClassLocation() {
            return "org.hl7.fhir.r4.model";
        }
    };

    abstract public String getClassLocation();
}
```

If we ever have a 3rd FHIR standard, this pattern is far easier to update.

bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClientImpl.java
```diff
@@ -150,15 +155,14 @@ public IBaseBundle requestEOBFromServer(String patientID, OffsetDateTime sinceTi
...
         String hicnHashVal = hashIdentifier(hicn, bfdHashPepper, bfdHashIter);
         var hicnHashEquals = generateHash(hicnHash, hicnHashVal);
-        return clientSearch(hicnHashEquals);
+        return clientSearch(version, hicnHashEquals);
     }
 
-    private IBaseBundle clientSearch(@SuppressWarnings("rawtypes") ICriterion hashEquals) {
-        Versions.FhirVersions version = getVersion();
-        return client.search()
+    private IBaseBundle clientSearch(Versions.FhirVersions version, @SuppressWarnings("rawtypes") ICriterion hashEquals) {
```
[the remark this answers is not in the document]
sb-wnyffenegger: Thanks

sb-wnyffenegger (dismissed):

bfd/src/test/java/gov/cms/ab2d/bfd/client/BFDMockServerConfigurationUtilSTU3.java
```diff
@@ -3,14 +3,11 @@
...
-import java.io.IOException;
-import java.net.ServerSocket;
-
 import static org.springframework.test.context.support.TestPropertySourceUtils.addInlinedPropertiesToEnvironment;
 
-public class BFDMockServerConfigurationUtil {
+public class BFDMockServerConfigurationUtilSTU3 {
 
-    public static final int MOCK_SERVER_PORT = randomMockServerPort();
+    public static final int MOCK_SERVER_PORT = 8080;
```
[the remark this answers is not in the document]
sharshar-sb: Fixed

sb-wnyffenegger (dismissed):

worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
```diff
@@ -232,7 +233,7 @@ private int processHandles(List<Future<EobSearchResult>> futureHandles, Progress
...
```
sb-wnyffenegger, line 236: I thought this would be removed and instead pull a JSON parser from the already initialized ones.

worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
```diff
@@ -334,7 +335,7 @@ void writeExceptionToContractErrorFile(StreamHelper helper, String data, Excepti
         var errMsg = ExceptionUtils.getRootCauseMessage(e);
         IBaseResource operationOutcome = FHIRUtil.getErrorOutcome(errMsg, version);
 
-        var jsonParser = fhirContext.newJsonParser();
+        var jsonParser = Versions.getContextFromVersion(version).newJsonParser();
```
sb-wnyffenegger, line 338: Same as above comment.

worker/src/main/java/gov/cms/ab2d/worker/processor/ContractProcessorImpl.java
sharshar-sb: Whoops, fixed

sb-wnyffenegger (approved):

sb-wnyffenegger (approved):

nflearl (approved):
BFDHealthCheck.java seems a bit too complicated to me.   And I think the subsequent suggestion might be able to help.

There is a *lot* of passing the version around by argument.  Consider storing that information contextually, either by explicitly being in a Spring Context or via Thread Local or via some other mechanism.  That would save a lot of references to something that should only be employed in a few areas.

PR #627, opened 2021-02-26
Techdebt/ab2d 3353 add org column to user account

sb-wnyffenegger:
**JIRA Tickets:**

[AB2D-3353](https://jira.cms.gov/browse/AB2D-3353) - Add Org Column to User Account
 
#### The most important changes are
1. New SQL files
2. User -> PdpClient
3. UserDTO -> PdpClientDTO
4. Removing all logging of the client id (found several logs)

Everything else is essentially just renaming all over the place.

#### Major changes introduced in this PR
1. Create a `user_account.organization` column which matches `sponsor.org_name`. The table values are copied over from `sponsor.org_name`.
2. Rename the `user_account.username` column to `user_account.client_id`.
3. Rename the User domain object to PdpClient.
4. Rename the UserDTO domain object to PdpClientDTO.
5. Remove all mentions of "user" within the system and replace with "client" because we don't have users (like people) but clients (like systems).
6. Remove all logging

#### In Subsequent PRs We will

1. Change the eventlogger to log organization and not client_id.
2. Remove hashing client id in the eventlogger because it will no longer be necessary.
3. Remove the sponsor table.
4. Rename the user_account table to the client table.

#### Caveats
There is a unique constraint on the `user_account.organization` column that may cause the liquibase scripts to fail if two sponsors have the same org name. Testing in both prod and sandbox has found there are no sponsors with the same org name so this should not be an issue.

* potential failures caused by the liquibase scripts that may break in an environment.
* mistakes changing the User and UserDTO classes to PdpClient and PdpClientDTO classes.
* potential issues with our api caused by changing the PdpClient and PdpClientDTO classes
* **make sure that client ids are no longer logged anywhere. I found several as part of exception messages**

### Security Implications

### Submitter Checklist


sharshar-sb (approved):

common/src/main/resources/db/changelog/v2021/rework_user_account_username.sql
```diff
@@ -0,0 +1,19 @@
+--liquibase formatted sql
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset wnyffenegger:rework_user_account_username failOnError:true
+
+ALTER TABLE user_account ADD COLUMN organization VARCHAR(255) NOT NULL DEFAULT '';
+
+UPDATE user_account
+SET organization = sponsor.org_name
```
nflearl, line 9: I was concerned that if org_name was null, then the set would blow up due to the Null constraint.  I checked in prod and org_name is fully filled for all sponsors.

It would be safer to have a where != null clause but not necessary.

common/src/main/java/gov/cms/ab2d/common/model/PdpClient.java
```diff
@@ -11,12 +11,16 @@
...
+ * Represents a PDP organization which possesses a single set of credentials for
+ * accessing the system.
+ */
 @Entity
 @Table(name = "user_account")
 @Getter
 @Setter
 @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
-public class User extends TimestampBase implements UserDetails {
+public class PdpClient extends TimestampBase implements UserDetails {
```
nflearl, line 23: I love getting rid of one of the **worst** domain nouns ever, replacing it with something that actually matches the domain.

nflearl (approved):

PR #633, opened 2021-02-26
Techdebt/ab2d 3371 remove client id event tables

sb-wnyffenegger:
**JIRA Tickets:**

[AB2D-3371](https://jira.cms.gov/browse/AB2D-3371) - Remove Client Id from Event Tables
 
### What Does This PR Do?

1. Create organization column in all database event tables
2. Populates column with values from `user_account.organization`
3. Deletes `user_id` column in all event tables
4. Create a simple filter to make sure that the `LoggableEvent.organization` field does not contain a client id.
5. Apply filter in both the SQL and Kinesis loggers
5. Change all event classes and mappers to work with organization instead of client id
6. Change all invocations of event class constructors to use organization instead of client id

1. That the SQL changes aren't going to cause liquibase failures in prod
2. That the client id filter is applied correctly in the loggers
3. That the event constructors in the Worker and API nodes all correctly use organization instead of client id
4. That the test cases for the client id filter are sufficient.

### Security Implications

### Submitter Checklist

### DIFF
--- a/common/src/main/java/gov/cms/ab2d/common/util/EventUtils.java
+++ b/common/src/main/java/gov/cms/ab2d/common/util/EventUtils.java
@@ -9,17 +9,17 @@ import java.io.File;
 
 public class EventUtils {
     public static JobStatusChangeEvent getJobChangeEvent(Job job, JobStatus jobStatus, String message) {
-        return new JobStatusChangeEvent(getClientId(job), getJobId(job), getJobStatus(job),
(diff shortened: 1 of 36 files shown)


nflearl (approved):

sharshar-sb:
One small comment - this will effect BFD insights since the Glue table will have had incorrect column names now.

sharshar-sb (approved):

PR #737, opened 2021-08-16
Migrate to new coverage table

sb-wnyffenegger:
**JIRA Tickets:**

[AB2D-3837](https://jira.cms.gov/browse/AB2D-3837) - Migrate to New Coverage Table
 
### What Does This PR Do?

Update the code to insert and query data from the partitioned coverage table. All of the inserts and queries have been performance tested already.

Changes:

1. Convert beneficiary_id from a String to Long
1. Rename "coverage" -> "coverage_deprecated" and "coverage_partitioned" -> "coverage"
1. Allow tests to set the attestation time for a contract as part of creating the contract (guarantees it is saved to DB immediately)
1. Modify all Insert and Select operations in the CoverageServiceRepository to reflect new partitioned database structure
     1. Add contract and year onto all queries to guarantee to avoid full table scans
     1. Modify prepared statement queries to include extra parameters
     1. Modify functions interpreting enrollment paging results
     1. Stop requesting enrollment by coverage period id and start requesting by contract
     1. Simplify enrollment paging requests to a single function
     1. Change how SQL results limit of the paging requests is calculated
1. Modify CoverageService interface and internals to use new CoverageServiceRepository functionality and provide more information via a CoveragePagingRequest to reduce number of arguments.
1. Change JDBC driver string to allow true batch inserts
1. Provide partitions for test contracts for unit and integration testing
1. Update PerformanceTestingCoverageOperations to build up a DB with the appropriate number of records to test inserts and queries

1. String to Long beneficiary_id change will not introduce problems with fhir module, with putting mbis onto claims, or cause unexpected behavior when a beneficiary id is missing
1. CoverageServiceRepository
    1. SQL modifications and changes
    1. Coverage paging functionality changes do not introduce off by one or NPE issues
    1. Delta and other calculations will not be slowed down 
1. JobData and ContractData job tracker changes mesh with job tracking modifications

Tested using staging environment against large number of large contracts.

### Database Changes

1. Rename "coverage" -> "coverage_deprecated" and "coverage_partitioned" -> "coverage"

Switch from a single coverage table to partitioned tables with different indexes optimizing specific query and insert operations.

### Security Implications

### Submitter Checklist


common/src/main/resources/db/changelog/test_only/coverage-fake-partitions.sql
```diff
@@ -6,7 +6,7 @@
 -- May be removed or altered in future once cutover to partitioned coverage table is complete
 -- this is used for testing code changes locally
 
-create table if not exists tst partition of coverage_partitioned
+create table if not exists tst partition of coverage
```
sharshar-sb, line 9: Is changing an sql file going to cause issues with the db?

sb-wnyffenegger: I doubt it because this is only applied in the 'test' context so it is only applied locally when running unit and integration tests.

worker/src/test/resources/application.properties
```diff
@@ -1,5 +1,6 @@
 ## ----------------------------------------------------------------------------  DATA-SOURCE CONFIG
-spring.datasource.url=${DB_URL}
+spring.datasource.url=${DB_URL}?reWriteBatchedInserts=true
+spring.datasource.hikari.data-source-properties.reWriteBatchedInserts=true
```
sb-wnyffenegger, line 3: Need to remove the hikari line here

common/src/test/java/gov/cms/ab2d/common/util/DataSetup.java
```diff
@@ -186,9 +188,14 @@ public void deleteCoverageSearchEvent(CoverageSearchEvent event) {
         coverageSearchEventRepo.delete(event);
     }
 
-    public Contract setupContract(String contractNumber) {
+    public Contract setupContract(String contractNumber, OffsetDateTime attestedOn) {
```
nflearl, line 191: A much cleaner way would be to preserve the original api and add another overloaded method that takes in the specific date string.  

The body of the original method can be nothing more than
return setupContract(contractNumber, OffsetDateTime.now());

api/src/test/java/gov/cms/ab2d/api/controller/AdminAPIPdpClientTests.java
```diff
@@ -363,7 +363,7 @@ public void getClientNotFound() throws Exception {
     }
 
     private void setupClient(String clientId, boolean enabled) {
-        Contract contract = dataSetup.setupContract(ENABLE_DISABLE_CONTRACT);
+        Contract contract = dataSetup.setupContract(ENABLE_DISABLE_CONTRACT, null);
```
nflearl, line 366: Instant red flag, adding a parameter that is now **null** for many cases.  See comments on the setupContract() method for how to fix.  

This will also greatly reduce the number of changes in this PR.

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -267,7 +267,7 @@ void testPatientExportDuplicateSubmissionWithDifferentClient() throws Exception
 
         List<Job> jobs = jobRepository.findAll(Sort.by(Sort.Direction.DESC, "id"));
         PdpClient pdpClient = new PdpClient();
-        Contract contract = dataSetup.setupContract("Test");
+        Contract contract = dataSetup.setupContract("Test", null);
```
nflearl, line 270: Remove all the cases of , null in all files.

I won't comment on any more instances.

bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDClient.java
```diff
@@ -11,8 +11,8 @@
     String BFD_CLIENT_ID = "AB2D";
     String BFD_HDR_BULK_JOBID = "BULK-JOBID";
 
-    IBaseBundle requestEOBFromServer(FhirVersion version, String patientID);
-    IBaseBundle requestEOBFromServer(FhirVersion version, String patientID, OffsetDateTime sinceTime);
+    IBaseBundle requestEOBFromServer(FhirVersion version, Long patientID);
```
nflearl, line 14: Can patient IDs ever be null?  

Otherwise, make it a primitive and one avoids tons of NPE issues.

common/src/test/java/gov/cms/ab2d/common/util/Coverage.java
```diff
@@ -8,23 +8,29 @@
...
-    private final int periodId;
+    private final String contract;
 
-    private final long searchEventId;
+    private final int year;
+
+    private final int month;
 
-    private final String beneficiaryId;
+    private final Long beneficiaryId;
```
nflearl, line 21: This should always be present, correct?  It is not possible to have a mapping to a beneID without a beneId.  

Better to make primitive.

nflearl (changes requested):
Hit a couple of issues that need fixing on just a review of a handful of files.  Once those are taken care of, I'll review the broader changes.

common/src/main/resources/db/changelog/test_only/coverage-fake-partitions.sql
sb-wnyffenegger: I fixed this after realizing what you meant. Yeah it is an error on my part.

sb-wnyffenegger:
@nflearl removed most of the issues related to your comments I think

sharshar-sb (approved):

PR #766, opened 2021-09-21
Added support for Synthea data

sharshar-sb:
**JIRA Tickets:**

[AB2D-3874](https://jira.cms.gov/browse/AB2D-3874) - Added support for Synthea data
 
### What Does This PR Do?

This PR creates new user accounts to support the Synthea contracts. The contracts are:

- Z1001
- Z1002
- Z1003
- Z1004
- Z1005
- Z1006
- Z1007
- Z1008
- Z1009

In the database, user credentials were created to reflect the new accounts provided in test Okta for the purpose of PDP testing searches in the sandbox. These credentials are public and will be available on the website so committing them here is not a security violation.

Because the coverage table is partitioned along contract lines, there was no partition defined for the new contracts. I updated the sandbox partition to accommodate the new contracts and created a default partition so that inserts will not fail if a specific partition is not defined for that contract/year combination. Partitions can be added later as needed and the data will move to those partitions.

To run jobs with this data, the coverage data needs to be loaded for these contracts. You can either create a job or set these properties in the database to these values:

```
update properties set value='engaged' where key='coverage.update.discovery';
update properties set value='engaged' where key='coverage.update.queueing';
update properties set value='true' where key='coverage.update.override';
```

These contracts are dependent on BFD providing these contracts, beneficiaries associated with the contracts and ExplanationOfBenefit objects associated with these beneficiaries. For more on Synthea, check out https://synthea.mitre.org/

### Database Changes

Creation of test contracts and coverage table partitions

### Security Implications

I am committing client IDs to GitHub and the database on purpose! They are part of test credentials that will go on our public website to help PDPs test running jobs in sandbox.

### Submitter Checklist


sb-wnyffenegger (dismissed):
Looks good

common/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -120,3 +119,7 @@ databaseChangeLog:
       file: db/changelog/v2021/drop_old_coverage_table.sql
   - include:
       file: db/changelog/v2021/rework_search_event.sql
+  - include:
```
sb-wnyffenegger, line 122: Can't we just change the original test code? It isn't used in any of our environments so I don't think it'll interfere with liquibase. Might be worth experimenting.

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageMappingCallable.java
```diff
@@ -36,7 +36,6 @@
 
     private final int year;
 
-    private final boolean skipBillablePeriodCheck;
```
sb-wnyffenegger, line 39: Removing this line will trigger changes in the ab2d-ops repo. We need to capture that in a ticket.

sb-wnyffenegger (commented):
Changes look good just wondering whether the extra test sql file is necessary. Need to track changes to ops deployment triggered by this.

common/src/main/resources/db/changelog/db.changelog-master.yaml
sharshar-sb: Worth a try.

worker/src/main/java/gov/cms/ab2d/worker/processor/coverage/CoverageMappingCallable.java
sharshar-sb: It's definitely used other places if that makes a difference. It's just not used in the Callable to determine the year to pass to the contract coverage search.

nflearl (dismissed):

sb-wnyffenegger (approved):

PR #785, opened 2021-10-05
Create default since behavior for R4.

sharshar-sb:
**JIRA Tickets:**

[AB2D-3990](https://jira.cms.gov/browse/AB2D-3990) - Add default since behavior for R4.

### What Does This PR Do?

When a user creates a job and they do not provide a since date, we will change the default behavior (for R4 jobs) to populate the since date.

1. The default behavior of _since will be the date of the PDPs last run job. This would provide an incremental search by default and alleviate the responsibility of PDPs to maintain the accuracy of the time of their last data pull. We know when their last job was requested so we can populate it for them.
1. If the PDP has never pulled data, the default _since behavior would default to the contract’s attestation date.
1. It is important to note that If a _since parameter is provided, it would of course be respected (assuming it was not before their attestation date or before AB2D data services began)

Situations where the default since behavior should not be used.

### Database Changes

Added a field to define how the since value was populated. This way, if we've changed the value in the since column, we'll know that we've done it, not that it was supplied by the PDP.

### Limitations

Currently, this is only defined for the R4 version (V2) as this is a change to an API contract and so we didn't want to effect our current users using STU3

### Security Implications

This can actually limit the amount of data returned

### Submitter Checklist


sb-wnyffenegger (changes requested):
I think you need to add e2e tests related to this for the since date. It would also be nice to see documentation on the SinceSource enum for exactly what each case means. Kind of like what is described in the PR.

common/src/main/resources/db/changelog/v2021/support_default_since.sql
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted sql
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset lsharshar:support_default_since failOnError:true
+
+ALTER TABLE job ADD since_source VARCHAR(256);
```
nflearl, line 6: Give it a default of USER and make this field NOT NULL - makes life much easier.

common/src/main/resources/db/changelog/v2021/support_default_since.sql
```diff
@@ -0,0 +1,8 @@
...
+
+UPDATE job SET since_source = 'USER' where since IS NOT NULL;
```
nflearl, line 8: I think the where clause is backwards.  One wants to initialize where null.

fhir/src/main/java/gov/cms/ab2d/fhir/FhirVersion.java
```diff
@@ -16,25 +16,31 @@
 @Slf4j
 public enum FhirVersion {
     STU3("org.hl7.fhir.dstu3.model", FhirContext.forDstu3(), FhirVersionEnum.DSTU3, "/v1/", "/v1/",
-            org.hl7.fhir.dstu3.model.ResourceType.Patient),
+            org.hl7.fhir.dstu3.model.ResourceType.Patient, false),
```
nflearl, line 19: Instead of adding a variable, add a method that for the versions that support default since, handles the behavior.  This is the OO way of doing things by delegating to the instance the instance specific behavior.  For the versions that do not, does the default:

void computeSince(Job job, JobRepo jobRepo);

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
```diff
@@ -41,6 +46,16 @@ public Job preprocess(String jobUuid) {
...
+        if (job.getSince() != null) {
+            // If the user provided a 'since' value
+            job.setSinceSource(SinceSource.USER);
+            jobRepository.save(job);
+        } else if (job.getFhirVersion().supportDefaultSince()) {
+            // If the user did not, but this version supports a default 'since', populate it
+            job = updateSinceTime(job);
+            jobRepository.save(job);
+        }
+
```
nflearl, line 58: Replace with 

fhirVersion.computeSince(job, jobRepo);

nflearl (changes requested):
I proposed a different way of structuring the code, allowing the FHIRversion enum to delegate the since behavior.  Most folks aren't aware of how powerful enums are.

api/src/main/java/gov/cms/ab2d/api/controller/v2/BulkDataAccessAPIV2.java
```diff
@@ -37,6 +37,7 @@
 import static gov.cms.ab2d.api.controller.common.ApiText.BULK_RESPONSE;
 import static gov.cms.ab2d.api.controller.common.ApiText.BULK_RESPONSE_LONG;
 import static gov.cms.ab2d.api.controller.common.ApiText.BULK_SINCE;
+import static gov.cms.ab2d.api.controller.common.ApiText.BULK_SINCE_DEFAULT;
```
nflearl, line 40: What is wrong with just doing a .* import?  After a certain number of imports, that makes sense to me.  There are several places where that is.  We should actually agree on a standard for the team as I constantly see the effects in code reviews of different IntelliJ settings, making for messier git history.

common/src/main/java/gov/cms/ab2d/common/service/CoverageService.java
```diff
@@ -45,10 +45,10 @@
 
     /**
      * Retrieve list of coverage periods associated with a contract
-     * @param contract contract to retrieve coverage periods for
+     * @param contractId contract to retrieve coverage periods for
      * @return list of coverage periods associated with a contract
      */
-    List<CoveragePeriod> findAssociatedCoveragePeriods(Contract contract);
+    List<CoveragePeriod> findAssociatedCoveragePeriods(Long contractId);
```
nflearl, line 51: Why the change?

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
```diff
@@ -42,6 +58,16 @@ public Job preprocess(String jobUuid) {
             throw new IllegalArgumentException(errMsg);
         }
 
+        if (job.getSince() != null) {
```
nflearl, line 61: Really not fond of null reference to enums.  In general, prefer to have an enum value of NONE.  That enables a lot of nice delegation patters that the null makes too complicated.

For this specific case with existing data, I understand the pushback.

nflearl (approved):

common/src/main/java/gov/cms/ab2d/common/service/CoverageService.java
sharshar-sb: I think this was based on a merge I had to do.

sb-wnyffenegger (approved):

nflearl (approved):

PR #793, opened 2021-10-18
Feature/ab2d 3893 Page Enrollment into Memory

sb-wnyffenegger:
**JIRA Tickets:**

[AB2D-3893](https://jira.cms.gov/browse/AB2D-3893) - Page Enrollment into Memory
 
### What Does This PR Do?

Instead of loading all beneficiaries for a job in memory, only load a subset of beneficiaries at a time to prevent memory pressure issues.

* FhirUtils.addMbis - changed to process single eob which fits Java stream pattern
* ContractData - remove in memory patient map
* PatientClaimsCollector - add mbis to eobs after each individual request to BFD instead of waiting until file write time
* PatientClaimsProcessorImpl - refactor try catch into try catch finally with clearer results
* ProgressTracker - stop counting beneficiaries loaded into memory as a separate part of the job. The only numbers used now to calculate % complete for a job are the number of patients whose eobs have been fully processed from BFD and the number of patients we expect to process for the job.
* application.properties - add env variable that lets you set the number of patients to queue at once defaulting to 1000. This number is also the number of patients between calls to check for completed requests

Changes to the tests:

Because beneficiaries are loaded into memory all of the tests have to mock or, in the case of integration tests, implement that behavior.

* ProgressTrackerIntegrationTest: deleted because percent completion values are now much simpler to calculate
* ContractProcessorInvalidPatientTest, ContractProcessorUnitTest, : provide fake beneficiary information so that requests can be completed
* ProgressTrackerTest: simplify tests now that percentage is calculated based off only two numbers
* JobProcessorIntegrationTest: provide utility methods for adding fake responses to coverage/beneficiary queries so that eob jobs can run
* ContractProcessorImpl:
    * loadEobRequests - load patients in memory x patients at a time and queue eob requests for those patients. Every time x patients are loaded (every iteration of the paging loop) check to see what requests are done. When all patients are loaded check that the expected number of patients (rows in db) matches the actual number iterated over
    * processRemainingRequests - refactor loop handling any leftover requests from loadEobRequests into its own function
    * ContractData removed from most nested functions
    * validPatientInContract no longer checked in contract processor because it has been moved to the PatientClaimsCollector

* Edge cases where patients may not be loaded in loadEobRequests
* Edge cases that do not have unit tests for ContractProcessorImpl
* Unit tests that were changed in a way that alters their meaning

### Security Implications

### Submitter Checklist

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -82,9 +82,9 @@
             <version>${logback.version}</version>
         </dependency>
         <dependency>
-            <groupId>net.rakugakibox.spring.boot</groupId>
+            <groupId>dev.akkinoc.spring.boot</groupId>
             <artifactId>logback-access-spring-boot-starter</artifactId>
-            <version>2.11.0</version>
+            <version>3.0.0</version>
         </dependency>
 
         <dependency>
--- a/eventlogger/pom.xml
+++ b/eventlogger/pom.xml
@@ -83,7 +83,7 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk</artifactId>
-            <version>1.12.89</version>
+            <version>1.12.94</version>
         </dependency>
         <dependency>
             <groupId>com.amazonaws</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-parent</artifactId>
-        <version>2.5.5</version>
+        <version>2.5.6</version>
         <relativePath/> <!-- lookup parent from repository -->
     </parent>
     <groupId>gov.cms.ab2d</groupId>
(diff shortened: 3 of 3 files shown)


worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsCollector.java
```diff
@@ -67,6 +73,7 @@ public void filterAndAddEntries(IBaseBundle bundle) {
                 // Remove Plan D
                 .filter(resource -> !EobUtils.isPartD(resource))
                 .filter(this::matchingPatient)
+                .peek(eob -> FhirUtils.addMbiIdsToEobs(eob, claimsRequest.getCoverageSummary(), claimsRequest.getVersion()))
```
nflearl, line 76: A static import can shorten this a bit.

sb-dkniesteadt (approved):

PR #796, opened 2021-10-21
Feature/ab2d 4033 too many jobs

sb-dkniesteadt:
**JIRA Tickets:**

[AB2D-42](https://jira.cms.gov/browse/AB2D-42) - Description

***Related Tickets***

[AB2D-4033](https://jira.cms.gov/browse/AB2D-4033) - CMS Ticket
 
### What Does This PR Do?
We have limits on how many jobs a user can run at a time. When the user attempts to start a job and the max amount of jobs are running, they get an error message with no indication on what jobs are running. This PR add into the header what jobs are currently running.

### Security Implications

### Submitter Checklist

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
@@ -177,7 +177,7 @@ public class ErrorHandler extends ResponseEntityExceptionHandler {
     public ResponseEntity<JsonNode> handleTooManyRequestsExceptions(final TooManyRequestsException e, HttpServletRequest request) throws IOException {
         HttpHeaders httpHeaders = new HttpHeaders();
         httpHeaders.add(RETRY_AFTER, Integer.toString(retryAfterDelay));
-        if(e.getJobIds() != null) {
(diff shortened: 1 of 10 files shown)


api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
```diff
@@ -176,6 +177,9 @@ private static HttpStatus getErrorResponse(Class clazz) {
     public ResponseEntity<JsonNode> handleTooManyRequestsExceptions(final TooManyRequestsException e, HttpServletRequest request) throws IOException {
         HttpHeaders httpHeaders = new HttpHeaders();
         httpHeaders.add(RETRY_AFTER, Integer.toString(retryAfterDelay));
+        if (e.getJobIds() != null) {
```
sb-wnyffenegger, line 180: @sharshar-sb @nflearl do we want to update our production database to explicitly limit all contracts to one running job? In sandbox this is very problematic so it isn't something we can add to liquibase.

sb-wnyffenegger:
@sb-dkniesteadt Could you also update the Swagger annotations in the API (V1 and V2). That should cover some of the documentation you were worried about.

PR #812, opened 2021-11-03
Remove skipBillablePeriodCheck environment variable

sharshar-sb:
**JIRA Tickets:**

[AB2D-4076](https://jira.cms.gov/browse/AB2D-4076) - Codify test exceptions

### What Does This PR Do?

Removes skipBillablePeriodCheck to the code and as an environment variable. Currently, this created a blanket way to handle contract searches in a particular environment. Originally, this was used to handle test contracts with bad billablePeriod data but now that we have different types of contracts, it isn't always accurate for all tests (Synthea) and it shouldn't apply to an entire environment. We should be able to test test contracts in prod too.

Any place where it was previously checked, verify that it is still being checked for the appropriate thing.

This is part of a similar PR in [ab2d-ops](https://github.com/CMSgov/ab2d-ops/pull/87) which can be deployed after this one as it removed the definition of the variable which should be done after AB2D no longer needs it.

### Limitations

This is used for the test contracts we know up to this date. In theory, there could be more types in the future.

### Security Implications

### Submitter Checklist

### DIFF
--- a/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
+++ b/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
@@ -22,7 +22,8 @@ public class Contract extends TimestampBase {
 
     public static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd H:m:s Z");
 
-    public enum UpdateMode { AUTOMATIC, TEST, MANUAL }
+    public enum UpdateMode { AUTOMATIC, NONE, MANUAL }
+    public enum ContractType { NORMAL, OLD_TEST, SYNTHEA }
 
     @Id
     @GeneratedValue
@@ -47,6 +48,9 @@ public class Contract extends TimestampBase {
     @Enumerated(EnumType.STRING)
     private UpdateMode updateMode = UpdateMode.AUTOMATIC;
 
+    @Enumerated(EnumType.STRING)
+    private ContractType contractType = ContractType.NORMAL;
+
     public Contract(@NotNull String contractNumber, String contractName, Long hpmsParentOrgId, String hpmsParentOrg,
                     String hpmsOrgMarketingName) {
         this.contractNumber = contractNumber;
@@ -63,7 +67,7 @@ public class Contract extends TimestampBase {
     private Set<CoveragePeriod> coveragePeriods = new HashSet<>();
 
     public boolean isTestContract() {
-        return updateMode == UpdateMode.TEST;
+        return contractType == ContractType.OLD_TEST || contractType == ContractType.SYNTHEA;
     }
 
     public boolean hasAttestation() {
@@ -118,26 +122,5 @@ public class Contract extends TimestampBase {
         String dateWithTZ = attestationDate + " " + getESTOffset();
         attestedOn = OffsetDateTime.parse(dateWithTZ, FORMATTER);
         return true;
-    }
-
-
-    public boolean isSyntheaContract() {
-        return isSyntheaContract(this.contractNumber);
-    }
-
-    public boolean hasTestDateIssues() {
-        return hasTestDateIssues(this.contractNumber);
-    }
-
-    public static boolean hasTestDateIssues(String contractNumber) {
-        return isTestContract(contractNumber) && !isSyntheaContract(contractNumber);
-    }
-
-    public static boolean isTestContract(String contractNumber) {
-        return contractNumber.startsWith("Z");
-    }
-
-    public static boolean isSyntheaContract(String contractNumber) {
-        return contractNumber.startsWith("Z1");
     }
 }
--- a/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
+++ b/common/src/main/java/gov/cms/ab2d/common/model/Contract.java
@@ -23,7 +23,13 @@ public class Contract extends TimestampBase {
     public static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd H:m:s Z");
 
     public enum UpdateMode { AUTOMATIC, NONE, MANUAL }
-    public enum ContractType { NORMAL, OLD_TEST, SYNTHEA }
+    public enum ContractType {
+        NORMAL, CLASSIC_TEST, SYNTHEA;
+
+        public boolean isTestContract() {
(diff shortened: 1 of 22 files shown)


sb-wnyffenegger (dismissed):

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
```diff
@@ -64,10 +61,16 @@ public Job preprocess(String jobUuid) {
             // If the user provided a 'since' value
             job.setSinceSource(SinceSource.USER);
             jobRepository.save(job);
-        } else if (job.getFhirVersion().supportDefaultSince() && !skipBillablePeriodCheck) {
-            // If the user did not, but this version supports a default 'since', populate it
-            job = updateSinceTime(job);
-            jobRepository.save(job);
+        } else if (job.getFhirVersion().supportDefaultSince()) {
+            boolean hasDateIssue = false;
+            if (job.getContract() != null) {
```
nflearl, line 66: How can a job not have a contract?  That should not be possible and should just be asserted which then cleans up the logic.

common/src/main/java/gov/cms/ab2d/common/model/Contract.java
```diff
@@ -119,4 +119,25 @@ public boolean updateAttestation(boolean attested, String attestationDate) {
...
+        return isTestContract(contractNumber) && !isSyntheaContract(contractNumber);
+    }
+
+    public static boolean isTestContract(String contractNumber) {
+        return contractNumber.startsWith("Z");
+    }
+
+    public static boolean isSyntheaContract(String contractNumber) {
+        return contractNumber.startsWith("Z1");
+    }
```
nflearl, line 142: There are at least three approaches to determining the "type" of a contract:
1) Naming Conventions
2) updateMode currently flags TEST contracts.  A new enum type of SYNTHEA could just as easily be added and then this field would server two purposes.
3) add a new enum/field in contract.

For this review, just indicate why #1 was preferred.  The downside is a reduced flexibility in what a contract name could be and a bit of tribal knowledge instead of things being quite explicit.

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsCollector.java
```diff
@@ -65,7 +64,7 @@ public void filterAndAddEntries(IBaseBundle bundle) {
         // Perform filtering actions
         BundleUtils.getEobResources(bundleEntries).stream()
                 // Filter by date
-                .filter(resource -> skipBillablePeriodCheck || FilterOutByDate.valid(resource, attestationDate, earliestDate, claimsRequest.getCoverageSummary().getDateRanges()))
+                .filter(resource -> Contract.hasTestDateIssues(claimsRequest.getContractNum()) || FilterOutByDate.valid(resource, attestationDate, earliestDate, claimsRequest.getCoverageSummary().getDateRanges()))
```
nflearl, line 67: Not thrilled about Contract exposing static methods.

PatientClaimsRequest could easily store a reference to a contract (instead of the number), allowing for the call to be simply claimsRequest.getContract().hasTestDateIssues().  That helps in hiding more details.  Let Contract manage the contract number.

nflearl (changes requested):

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
sharshar-sb: Some of the tests did not include contract numbers. It's not an unreasonable thing to check it.

sb-wnyffenegger: It shouldn't be possible anymore so maybe we need to update the tests in a separate story.

sb-wnyffenegger (commented):
@sb-dkniesteadt @nflearl I think that we each need to approve this and then review exactly what this means.

@sharshar-sb I think we need explicit documentation on what business requirement this is meeting and how.

Also, is there any way to test this with an existing contract. For instance running a real contract with a specific since and then deploying this and then running again to make sure we get the exact same results?

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
nflearl: Please add a TODO then.  Bonus points for including the ticket number in the comment.

common/src/main/java/gov/cms/ab2d/common/model/Contract.java
```diff
@@ -22,7 +22,8 @@
 
     public static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd H:m:s Z");
 
-    public enum UpdateMode { AUTOMATIC, TEST, MANUAL }
+    public enum UpdateMode { AUTOMATIC, NONE, MANUAL }
+    public enum ContractType { NORMAL, OLD_TEST, SYNTHEA }
```
nflearl, line 26: Naming suggestion: CLASSIC_TEST instead of OLD.

common/src/main/java/gov/cms/ab2d/common/model/Contract.java
```diff
@@ -63,7 +67,7 @@ public Contract(@NotNull String contractNumber, String contractName, Long hpmsPa
     private Set<CoveragePeriod> coveragePeriods = new HashSet<>();
 
     public boolean isTestContract() {
-        return updateMode == UpdateMode.TEST;
+        return contractType == ContractType.OLD_TEST || contractType == ContractType.SYNTHEA;
```
nflearl, line 70: This is ok.

A fancier suggestion is to delegate to the enum as to whether it is a TEST.  e.g. return contractType.isTest();

nflearl (changes requested):
rename OLD => CLASSIC

Add todo about being able to assume every job has a contract.

sb-wnyffenegger (commented):
Addressing some PR comments a

nflearl (dismissed):

sb-dkniesteadt (approved):

sb-wnyffenegger (dismissed):

nflearl (approved):

PR #883, opened 2022-01-13
Feature/ab2d 4261 coverage module

sb-dkniesteadt:
**JIRA Tickets:**

[AB2D-4261](https://jira.cms.gov/browse/AB2D-4261) - Description

### What Does This PR Do?
Move coverage code to its own module

Anything that might break existing functionality. Also the fact this is part 1 of a multi part process. The next step is to remove references of common module and fix the database to not be as dependent on contracts. 

### Security Implications

### Submitter Checklist


coverage/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
```
sb-wnyffenegger, line 41: This dependency needs to be in the test scope so it isn't built into the JAR.

coverage/src/main/java/gov/cms/ab2d/coverage/service/CoverageService.java
```diff
@@ -76,16 +75,16 @@
...
     boolean canEOBSearchBeStarted(int periodId);
 
     /**
-     * Get search status {@link JobStatus}
+     * Get search status {@link gov.cms.ab2d.common.model.JobStatus}
      * @param periodId {@link CoveragePeriod#getId()}
      * @return search status or null
      */
-    JobStatus getSearchStatus(int periodId);
+    gov.cms.ab2d.common.model.JobStatus getSearchStatus(int periodId);
```
sb-wnyffenegger, line 87: Is the full package path really necessary in the code?

coverage/src/test/java/gov/cms/ab2d/coverage/util/CoverageDataSetup.java
```diff
@@ -0,0 +1,249 @@
...
+        coverageSearchEventRepo.deleteAll();
+        coverageSearchEventRepo.flush();
+
+        coverageSearchRepo.deleteAll();
+        coverageSearchRepo.flush();
+
+        coveragePeriodRepo.deleteAll();
+        coveragePeriodRepo.flush();
+
+        List<Job> jobsToDelete = domainObjects.stream().filter(object -> object instanceof Job)
```
sb-wnyffenegger, line 91: Should this depend on common test for actions that are already available since most of this code is just copy-paste from the common module?

coverage/src/main/java/gov/cms/ab2d/coverage/service/CoverageService.java
sb-dkniesteadt: Yes, because its in a new module It needs the full path. If not then it will fail for javadoc. Importing to keep javadoc causes test check fail due to unused import.

coverage/src/test/java/gov/cms/ab2d/coverage/util/CoverageDataSetup.java
sb-dkniesteadt: I would say yes. Due to issues of the dependencies on the database we need to keep this. Removing it causes foriegn key errors. Once I have edited the database I can remove some of this code and remove the dependencies on common.

sb-wnyffenegger: I'm thinking more to extend DataSetup and then only override/implement the necessary methods.

sharshar-sb (approved):

coverage/src/main/java/gov/cms/ab2d/coverage/service/CoverageService.java
nflearl: Thanks for the explanation.

coverage/src/test/java/gov/cms/ab2d/coverage/util/AB2DPostgresqlContainer.java
```diff
@@ -0,0 +1,27 @@
+package gov.cms.ab2d.coverage.util;
+
+import java.util.Collections;
+import org.testcontainers.containers.PostgreSQLContainer;
+
+public class AB2DPostgresqlContainer extends PostgreSQLContainer<AB2DPostgresqlContainer> {
```
nflearl, line 6: What do you think about renaming this class to something like AB2DCoveragePostgressqlContainer in order to disambiguate from other copies?

nflearl (changes requested):
Please fix the wildcard import by expanding.

Consider renaming the PostGres container class.

coverage/src/test/java/gov/cms/ab2d/coverage/util/CoverageDataSetup.java
sb-dkniesteadt: Yes. My plan as I decouple things more to refactor this code to not be copy paste and decoupled from common.

sb-wnyffenegger (approved):

PR #889, opened 2022-01-19
AB2D-4291 Explicitly Name Slack Alerts

sb-wnyffenegger:
**JIRA Tickets:**

[AB2D-4291](https://jira.cms.gov/browse/AB2D-4291) - Name Slack Alerts
 
### What Does This PR Do?

Give each Slack alert a specific name that corresponds to a table in Confluence. Each alert has instructions to follow and who to contact.

There are diminishing returns to making substantial changes to support this. Because our alerting system supports SQL, Kinesis, and Slack, making any one change can have broad consequences. This implementation is more an MVP to limit the changes necessary to support defined slack alert names.

* A prettier version of this would include the name at the beginning of every message. Instead there is only a guarantee that the alert name will appear in the message.
* Hacky way in JobStatusChangeEvent to guarantee that alert name always appears first.

### Security Implications

### Submitter Checklist


nflearl (changes requested):
Extract these out into their own enum.  This will give folks a 1 stop place to see all the possible values plus avoid a mis-spelling causing two different alerts for the same type of event.

nflearl (approved):

PR #920, opened 2022-02-09
Add ab2d-ops create-analyst-view.sql

sb-dkniesteadt:
**JIRA Tickets:**

[AB2D-4391](https://jira.cms.gov/browse/AB2D-4391) - Description
 
### What Does This PR Do?
Moves sql scripts from ab2d-ops to the main repo

Any thing nneded to add in the sql script

### Security Implications

### Submitter Checklist


common/src/main/resources/db/changelog/v2022/create-analyst-view.sql
```diff
@@ -0,0 +1,142 @@
...
+GRANT SELECT ON event_error TO ab2d_analyst;
+GRANT SELECT ON event_file TO ab2d_analyst;
+GRANT SELECT ON event_job_status_change TO ab2d_analyst;
+GRANT SELECT ON job_output TO ab2d_analyst;
+GRANT SELECT ON contract_view TO ab2d_analyst;
+GRANT SELECT ON job_view TO ab2d_analyst;
+GRANT SELECT ON bcp_view TO ab2d_analyst;
+GRANT SELECT ON coverage_view TO ab2d_analyst;
+
+GRANT SELECT, UPDATE, INSERT ON ab2d_statistics TO ab2d_analyst;
```
nflearl, line 142: Check with Will but I don't think you want anything to do with the analyst in this checkin.

nflearl (changes requested):
What about the bcp_view?

common/src/main/resources/db/changelog/v2022/create-analyst-view.sql
sb-dkniesteadt: @wnyffenegger Comment?

sb-wnyffenegger: Well I can't control permissions from ab2d-ops if those views don't already exist so I would include everything.

sb-wnyffenegger: We're assuming that the ab2d_analyst is already created in the environment and you're just populating the views it can use. It starts with zero view of anything so you're going to have to do the grants.

sb-wnyffenegger: @sb-dkniesteadt one thing though, get rid of the create role part. Assume that the role already exists in the environment.

sb-wnyffenegger: Tricky, I don't think so. Maybe that changes things. I don't want to deal with that mess. It's a chicken and egg problem because during the deployment of the database these views won't exist. It's only when a worker/api start up that it will.

sb-wnyffenegger (commented):
Approved after merge in ab2d-ops

sb-ebukaanene (approved):

PR #1002, opened 2022-04-27
Add dropped max job tets

nflearl:
**JIRA Tickets:**

[AB2D-4449](https://jira.cms.gov/browse/AB2D-4449) - Add job max tests previously dropped.

***Related Tickets***
 
### What Does This PR Do?

Given the separation of JobService into it's own module and the fact that the api module now works through a job client, there were a couple of tests that were no longer appropriate in the api module.  Due to the massive size of an earlier refactoring they were moved into a JIRA ticket to be re-added later when the JobService was moved into its own module.

This PR restores the tests but in a more narrow focus specifically testing JobService features, not the overall MockMVC calls.

Also, noticed that build #141 on master failed which turned out to be a timing issue with comparing timestamps.  Created a custom matcher that can tolerate the test running across a one second boundary.

### Security Implications

### Submitter Checklist

### DIFF
--- a/api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
+++ b/api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
@@ -1149,6 +1149,9 @@ public class BulkDataAccessAPIIntegrationTests {
         };
     }
 
+    /*
+     * Be more accepting of a one-second difference in timestamps when running a test.
+     */
     private ResultMatcher buildTxTimeMatcher() {
         return result -> {
             OffsetDateTime buildTime = OffsetDateTime.now();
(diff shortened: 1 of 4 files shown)


api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
```diff
@@ -1148,4 +1148,20 @@ private ResultMatcher buildExpiresMatcher() {
...
         };
     }
+
+    /*
+     * Be more accepting of a one-second difference in timestamps when running a test.
+     */
+    private ResultMatcher buildTxTimeMatcher() {
+        return result -> {
+            OffsetDateTime buildTime = OffsetDateTime.now();
+            String buildTimeStr = new org.hl7.fhir.dstu3.model.DateTimeType(buildTime.toString()).toHumanDisplay();
```
sb-keane, line 1158: I want to say this is overly convoluted but I wouldn't be surprised if this is the best way to convert between these two datatime types.

sb-keane (approved):

api/src/test/java/gov/cms/ab2d/api/controller/BulkDataAccessAPIIntegrationTests.java
nflearl: I left the original convulsion in place and just added tolerance for deviation by a second.  Matchers are a bit tricky to work with.  At lease having the duplicate code in one builder method is simpler and more contained.

sb-keane (approved):

sb-dkniesteadt (approved):

sharshar-sb (approved):

PR #1069, opened 2022-07-29
Feature/ab2 d 4829 event service integration

sb-dkniesteadt:
Please Note. This branch is NOT GOING INTO MASTER. Until we are ready for an official deployment im putting the code in another project. 

**JIRA Tickets:**

[AB2D-4829](https://jira.cms.gov/browse/AB2D-4829) - Integrates ab2d with the event service Client

### What Does This PR Do?
Integrates ab2d with the event service Client. Replaced the direct logging into the database with sending events to an sqs queue. 

### Database Changes

- [x] All QuickSight impacts have been identified

### Security Implications

### Submitter Checklist


common/src/test/java/gov/cms/ab2d/common/util/AB2DLocalstackContainer.java
```diff
@@ -0,0 +1,27 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+import org.testcontainers.containers.localstack.LocalStackContainer;
+import org.testcontainers.utility.DockerImageName;
+
+
+@Slf4j
+public class AB2DLocalstackContainer extends LocalStackContainer {
+
+    private static final DockerImageName IMAGE_VERSION = DockerImageName.parse("localstack/localstack:latest");
```
sb-keane, line 11: Might be a good idea to set this to a specific version.

common/src/test/java/gov/cms/ab2d/common/util/AB2DLocalstackContainer.java
```diff
@@ -0,0 +1,27 @@
+package gov.cms.ab2d.common.util;
+
+import lombok.extern.slf4j.Slf4j;
+import org.testcontainers.containers.localstack.LocalStackContainer;
+import org.testcontainers.utility.DockerImageName;
+
+
+@Slf4j
+public class AB2DLocalstackContainer extends LocalStackContainer {
```
sb-keane, line 9: Were you not able to use this version in other modules? This class is copied multiple times except it's renamed.

sb-dkniesteadt: no, because its a test class, those don't get applied when using dependents. In the future we might create a future client with all default classes like this.

sb-keane: Oh yeah I see now. API and worker use spring to pull in the common version. The other modules need their own.

hpms/src/test/java/gov/cms/ab2d/hpms/service/MockedFetchTest.java
```diff
@@ -38,6 +39,9 @@ class MockedFetchTest {
     @Container
     private static final PostgreSQLContainer postgreSQLContainer = new AB2DPostgresqlContainer();
 
+    @Container
+    private static final AB2DLocalstackContainer localstackContainer = new AB2DLocalstackContainer();
```
nflearl, line 43: Is this really needed in all the places it has been added?  HPMS doesn't seem to have a need for using LocalStack.

nflearl (changes requested):
Seems to be way too much duplication in this submission.  Push up more to the ab2d pom.  Does every single service now require LocalStack in order to test?  That seems like overkill and contributes to the code base being an even bigger ball of mud.

sb-dkniesteadt:
> Seems to be way too much duplication in this submission. Push up more to the ab2d pom. Does every single service now require LocalStack in order to test? That seems like overkill and contributes to the code base being an even bigger ball of mud.

@nflearl 
The same reason we have postgres everywhere we need the localstack. It connects to sqs and will fail if there is no environment. If the project uses logmanager it needs the container with it. Once we have sqs up and running we can go back and mabye remove some of the postgres containers.

hpms/src/test/java/gov/cms/ab2d/hpms/service/MockedFetchTest.java
sb-dkniesteadt: The same reason we have postgres everywhere we need the localstack. It connects to sqs and will fail if there is no environment. If the project uses logmanager it needs the container with it. Once we have sqs up and running we can go back and mabye remove some of the postgres containers.

sb-dkniesteadt: Everything uses events and they need sqs to have that functionality working

eventlogger/src/test/resources/application.eventlogger.properties
```diff
@@ -22,3 +22,11 @@ eventlogger.kinesis.enabled=send_events
...
 slack.trace.webhooks=${AB2D_SLACK_TRACE_WEBHOOKS:#{''}}
 
+cloud.aws.credentials.access-key:${AWS_SQS_ACCESS_KEY_ID:#{'EXAMPLEKEY'}}
+cloud.aws.credentials.secret-key:${AWS_SQS_SECRET_ACCESS_KEY:#{'EXAMPLEKEY'}}
+cloud.aws.end-point.uri:${AWS_SQS_URL:#{'http://localhost:4566'}}
+cloud.aws.stack.auto=false
+cloud.aws.region.static=us-east-1
+
+feature.sqs.enabled=false
+
```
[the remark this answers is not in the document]
sb-dkniesteadt: I modified the rest of the code to use common. Eventlogger still needs it because it does not use the common module.

nflearl (approved):
What we really ought to do is to factor out the properties and the base AB2DLocalstackContainer into a test-localstack library that is a dependency where-ever it is needed.

Then, all the aws properties can be located there and we can reduce DRY.

See the embedded comment about how the POMS should really work.

Overall, this checkin is a step backwards in terms of complexity.  A lot of additional stuff was added that it seems everything is dependent upon.  This is an example of leaky abstractions, where implementation details are bleeding all over the project.

There has gotta be cleaner ways of doing things.  Perpetuating the big ball of mud really won't get us that far.

However, I recognize that this is a step on the journey and hopefully, as pieces are carved out, we can resume reducing accidental complexity.

sb-keane (approved):

PR #1135, opened 2022-10-28
Enhancement/ab2d 5134 integrate property service with ab2d

sharshar-sb:
**JIRA Tickets:**

[AB2D-5134](https://jira.cms.gov/browse/AB2D-5134) - Integrate property service with AB2D

### What Does This PR Do?

Creates a feature flag for the properties service, defaulting to false to use the property service or to use the database directly.

The environment variable PROPERTIES_SERVICE_FEATURE_FLAG should be set to true or false and the Properties URL should be defined in the PROPERTIES_SERVICE_URL environment URL.

### Database Changes

- [ ] All QuickSight impacts have been identified

### Security Implications

### Submitter Checklist


properties/src/main/java/gov/cms/ab2d/properties/service/PropertiesAPIServiceImpl.java
```diff
@@ -1,34 +1,84 @@
...
 
     @Override
     public String getProperty(String property) {
-        Properties properties = propertiesService.getPropertiesByKey(property);
-        return properties.getValue();
+        String value;
+        if (usePropertyService) {
+            try {
+                Property prop = propertiesClient.getProperty(property);
+                value = prop.getValue();
```
sb-dkniesteadt, line 35: can't these two be one line. Don't need 2 variables. Or at least a check if properties is null

properties/src/main/java/gov/cms/ab2d/properties/service/PropertiesAPIServiceImpl.java
```diff
@@ -54,6 +118,15 @@ public boolean deleteProperty(String key) {
         if (StringUtils.isEmpty(key)) {
             return false;
         }
+        if (usePropertyService) {
+            try {
+                propertiesClient.deleteProperty(key);
+                return true;
+            } catch (Exception ex) {
+                log.error(ERROR_MESSAGE, ex);
+                return propertiesService.deleteProperty(key);
```
sb-dkniesteadt, line 127: Might be better to have an event call instead. Especially if in not too long we are deleting propertiesService soon.

sb-dkniesteadt (changes requested):

properties/src/main/java/gov/cms/ab2d/properties/service/PropertiesAPIServiceImpl.java
sharshar-sb: This is temporary because this is the fallback to the local database if the service isn't working.

sb-keane (approved):

sb-dkniesteadt (approved):

PR #1234, opened 2023-04-20
Add gitleaks precommit config

gfreeman-navapbc:
**JIRA Tickets:**

[PLT-3](https://jira.cms.gov/browse/PLT-3) - Add pre-commit config for gitleaks to all repos
 
### What Does This PR Do?
Introduces pre-commit config for gitleaks

### Security Implications

- [x] This PR adds new software dependencies


gsf (approved):
We should add something to the README about usage but that can be in another PR

PR #1242, opened 2023-05-01
update readme to include precommit install

gfreeman-navapbc:
**JIRA Tickets:**

[PLT-3](https://jira.cms.gov/browse/PLT-3) - Add pre-commit config for gitleaks to all repos


gsf:
I'd like to make the gitleaks hook more of a default for the repo. Could we add a `.pre-commit-config.yaml` with gitleaks in the [default_install_hook_types](https://pre-commit.com/#top_level-default_install_hook_types)? Then in the readme the instructions can be to run `pre-commit install` without the hook-type flag.

Also, since most of us should be on macs, we might recommend `brew install pre-commit` with a link to https://pre-commit.com/#install for other options.

.pre-commit-config.yaml
```diff
@@ -3,3 +3,5 @@ repos:
     rev: v8.16.1
     hooks:
       - id: gitleaks
+
+default_install_hook_types: [pre-commit, gitleaks]
```
gsf, line 7: Looking into this further, I think I misunderstood it before. The gitleaks hook is already a hook type of "pre-commit", and since that's already the default, we can drop this definition of `default_install_hook_types`.

README.md
```diff
@@ -50,3 +50,25 @@ API Setup
 
 
 [AB2D Deploy](Deploy/README.md)
+
+## Installing and Using Pre-commit with Gitleaks
```
gsf, line 54: Let's leave gitleaks out of this. Also this should be added to the Table of Contents above.

```suggestion
## Installing and using pre-commit
```

README.md
```diff
@@ -50,3 +51,25 @@ API Setup
 
 
 [AB2D Deploy](Deploy/README.md)
+
+## Installing and Using Pre-commit
+
```
gsf, line 56: Since this is a compliance measure, could we add a line here along the lines of "Anyone committing to this repo must use the pre-commit hook to lower the likelihood that secrets will be exposed."

gsf (approved):
Looks good! Something like this should work in other repos as well

gsf (approved):
Looks good!

PR #1249, opened 2023-05-18
PLT-79: remove semanticbits gitleaks config

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-79

## 🛠 Changes

Removes external semanticbits config. 

## ℹ️ Context for reviewers

Most of the external configuration is now baked into gitleaks by default. We expect to add new rules as needed.

## 🔒 Security Implications

### DIFF
--- a/.github/workflows/gitleaks.yml
+++ b/.github/workflows/gitleaks.yml
@@ -7,7 +7,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       REPO: https://github.com/CMSgov/ab2d
-      REMOTE_EXCLUDES_URL: https://raw.githubusercontent.com/semanticbits/ab2d-gitleaks-automation/master/ab2d/gitleaks.toml
       GITLEAKS_VERSION: v7.2.2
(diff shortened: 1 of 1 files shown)


.github/workflows/gitleaks.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       REPO: https://github.com/CMSgov/ab2d
-      REMOTE_EXCLUDES_URL: https://raw.githubusercontent.com/semanticbits/ab2d-gitleaks-automation/master/ab2d/gitleaks.toml
+      REMOTE_EXCLUDES_URL: https://raw.githubusercontent.com/gitleaks/gitleaks/master/config/gitleaks.toml
```
gsf, line 10: We shouldn't need to reference the default config, right? Also, could we upgrade to latest?

gfreeman-navapbc: so, i'll have to add an extra unpack step in the script

gsf (approved):
Great work

PR #1266, opened 2023-06-21
Docker compose fix

oluwolenpbc:
Doubling ram and increasing timeouts for API test.


Sadibhatla:
Solved so no need to keep this open

PR #1270, opened 2023-06-29
PLT-143: Enable and configure dependabot for security updates across all repos

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-143

## 🛠 Changes

Update dependabot to only alert on security changes.

## 🔒 Security Implications


gsf (approved):
We could switch these to daily but weekly is also fine

PR #1283, opened 2023-09-13
PLT-136 Run e2e tests on impl environment

oluwolenpbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-136

## 🛠 Changes

Added workflow for running e2e tests on the impl environment.

## ℹ️ Context for reviewers

This e2e-test workflow should allow us to gain confidence in deploying while simplifying our test pipelines by running tests directly against resources in the impl environment.

## ✅ Acceptance Validation

We may need to merge this before we can test running the workflow.

## 🔒 Security Implications

None, these tests are run on self-hosted runners within our internal network.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -15,6 +15,7 @@ jobs:
       AWS_DEFAULT_REGION: "us-east-1"
       WORKSPACE: ${{ github.workspace }}
       AB2D_BFD_KEYSTORE_LOCATION: "${{ github.workspace }}/opt/ab2d/ab2d_bfd_keystore"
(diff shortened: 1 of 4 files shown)


.github/workflows/github-actions-impl-test.yml
```diff
@@ -0,0 +1,100 @@
+name: GitHub Actions Workflow
```
gsf, line 1: If this is for PLT-136 then we need two workflows:

- deploy.yml to deploy code to an environment (in this case, impl)
- e2e-test.yml to run e2e tests on an environment (also on the impl in this case)

The deploy and test on impl should run on push to master and manual dispatch.

Sadibhatla (approved):
LGTM

PR #1297, opened 2023-11-30
PLT-189 github actions workflow for unit & integration testing

oluwolenpbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-189

## 🛠 Changes

Added workflow to run unit and integration tests on PRs.

## ℹ️ Context for reviewers

This is a port of the Jenkins "AB2D-Repo" job to GitHub Actions. 

## ✅ Acceptance Validation

The workflow has been run many times to validate. See checks.

## 🔒 Security Implications

No new security implications.

### DIFF
--- a/.github/workflows/unit-integration-test.yml
+++ b/.github/workflows/unit-integration-test.yml
@@ -50,10 +50,10 @@ jobs:
           aws-region: us-east-1
           role-to-assume: arn:aws:iam::${{ vars.DEV_ACCOUNT_NUMBER }}:role/delegatedadmin/developer/github-actions-runner-role            
 
-      - name: Set Environment Variables
-        run: |
-          AWS_ACCOUNT_ID=${{ vars.DEV_ACCOUNT_NUMBER }}
-          AWS_REGION=${{ vars.AWS_REGION }}
+      - name: Set env vars from AWS params
+        uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
+        env:
+          AWS_REGION: ${{ vars.AWS_REGION }}
         with:
           params: |
             OKTA_CLIENT_ID=okta/client-id
--- a/.github/workflows/unit-integration-test.yml
+++ b/.github/workflows/unit-integration-test.yml
@@ -64,7 +64,7 @@ jobs:
       - name: Install Maven 3.6.3
         run: |
           if mvn -v; then echo "Maven already installed" && exit 0; else echo "Installing Maven"; fi
-          curl -LsS https://archive.apache.org/dist/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz | tar xf -
+          curl -LsS https://archive.apache.org/dist/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz | tar xzf -
           sudo mv apache-maven-3.6.3 /opt/maven
           echo "PATH=$PATH:/opt/maven/bin" >> $GITHUB_ENV
 
--- a/.github/workflows/unit-integration-test.yml
+++ b/.github/workflows/unit-integration-test.yml
@@ -8,14 +8,6 @@ jobs:
   build:
     runs-on: self-hosted
 
-    env:
-      CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
-      E2E_ENVIRONMENT: 'CI'
-      AB2D_V2_ENABLED: 'true'
-      SEARCH_BENE_BATCH_SIZE: '10'
-      AWS_DEFAULT_REGION: "us-east-1"
-      WORKSPACE: ${{ github.workspace }}
-
     steps:
       - name: Checkout Code
         uses: actions/checkout@v3
@@ -65,7 +57,7 @@ jobs:
           sudo rm -rf /opt/maven
           sudo mv "$tmpdir/apache-maven-3.6.3" /opt/maven
 
-      - name: Create ab2d workspace directory 
+      - name: Create ab2d workspace directory
         run: mkdir -p opt/ab2d
 
       - name: Download Code Coverage
--- a/.github/workflows/unit-integration-test.yml
+++ b/.github/workflows/unit-integration-test.yml
@@ -32,6 +32,10 @@ jobs:
           java-version: '17'
           maven-version: '3.6.3'
 
+      - name: Download bfd keystore
+        run: |
+          aws s3 cp s3://bcda-opensbx-access-logs/ab2d_sbx_keystore /tmp/ab2d_sbx_keystore 
+
       - name: Set env vars from AWS params
         uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
         env:
@@ -77,11 +81,7 @@ jobs:
           tar xzvf apache-maven-3.6.3-bin.tar.gz
           sudo mv apache-maven-3.6.3 /opt/maven
           echo "PATH=$PATH:/opt/maven/bin" >> $GITHUB_ENV
-        
-      - name: Download bfd keystore
-        run: |
-          aws s3 cp s3://bcda-opensbx-access-logs/ab2d_sbx_keystore /tmp/ab2d_sbx_keystore 
-
+  
 
       - name: Create ab2d workspace directory and copy in keystore
         run: sudo mkdir -p "$WORKSPACE/opt/ab2d"
--- a/.github/workflows/unit-integration-test.yml
+++ b/.github/workflows/unit-integration-test.yml
@@ -11,6 +11,16 @@ jobs:
     runs-on: self-hosted
 
 
+    env:
+      CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+      E2E_ENVIRONMENT: 'CI'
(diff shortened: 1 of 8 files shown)


.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,107 @@
...
+          OKTA_CLIENT_PASSWORD: ${{ secrets.OKTA_CLIENT_PASSWORD }}
+          SECONDARY_USER_OKTA_CLIENT_ID: ${{ secrets.SECONDARY_USER_OKTA_CLIENT_ID }}
+          SECONDARY_USER_OKTA_CLIENT_PASSWORD: ${{ secrets.SECONDARY_USER_OKTA_CLIENT_PASSWORD }}
+          CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+          E2E_ENVIRONMENT: 'CI'
+          AB2D_V2_ENABLED: 'true'
+          SEARCH_BENE_BATCH_SIZE: '10'
+          ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
+          ECR_REPO_ENV_AWS_ACCOUNT_NUMBER: ${{ vars.SBX_ACCOUNT_NUMBER }}
+          WORKSPACE: ${{ github.workspace }}
```
gsf, line 39: Any env we aren't setting from AWS parameter store should still be set in the top-level "env" as you were doing before. So this would include E2E_ENVIRONMENT, AB2D_V2_ENABLED, etc. The only variable you need to pass into this step is AWS_REGION.

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,107 @@
...
+        with:
+          params: |
+            ARTIFACTORY_URL=/artifactory/url
+            ARTIFACTORY_USER=/artifactory/user
+            ARTIFACTORY_PASSWORD=/artifactory/password 
+
+
+      - name: Install Maven 3.2.5
+        run: |
+          wget https://archive.apache.org/dist/maven/maven-3/3.2.5/binaries/apache-maven-3.2.5-bin.tar.gz
```
gsf, line 49: Run this without showing progress in the logs

gsf: Also, why are you installing 3.2.5 here when it's 3.6.3 above and in Jenkins? It really seems like the setup-java action should handle this but I found (as you probably did as well) this issue where it seems they've opted to pin to 3.8.7 instead of making it configurable: https://github.com/actions/setup-java/issues/457

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,107 @@
...
+          tar xzvf apache-maven-3.2.5-bin.tar.gz
+          sudo mv apache-maven-3.2.5 /opt/maven
+          echo "PATH=$PATH:/opt/maven/bin" >> $GITHUB_ENV
+        
+      - name: Create ab2d workspace directory and copy in keystore
+        run: sudo mkdir -p "$WORKSPACE/opt/ab2d"
```
gsf, line 55: I would love to drop this dependency from these unit and integration tests.

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Download Code Coverage
+        run: |
+          mkdir -p codeclimate
+          if [ ! -f ./codeclimate/cc-test-reporter ]; then
+            curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 \
```
gsf, line 61: Run this without progress output in the logs (with `--no-progress-meter` if this version of curl supports it).

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,107 @@
...
+#            ARTIFACTORY_URL=/artifactory/url
+#            ARTIFACTORY_USER=/artifactory/user
+#            ARTIFACTORY_PASSWORD=/artifactory/password
+#            SONAR_HOST_URL=/sonarqube/url
+#            SONAR_TOKEN=/sonarqube/token
+
+
+      - name: Run unit and integration tests
+        run: |
+          mvn -s settings.xml -X -Dusername=${ARTIFACTORY_USER} -Dpassword=${ARTIFACTORY_PASSWORD} -Drepository_url=${ARTIFACTORY_URL} test -pl common,job,coverage,api,worker
```
gsf, line 94: Drop DEBUG and INFO levels from the logs here.

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,107 @@
...
+          docker volume ls -qf dangling=true | xargs -I name docker volume rm name
+          docker ps -aq | xargs -I name docker rm --force name
+
+      # ... (Similar conversions for other cleanup steps)
+
+      - name: Post-Cleanup
+        if: always()
+        run: |
+          rm -rf "$WORKSPACE/opt/ab2d" 2> /dev/null
+          rm -rf "$WORKSPACE/.m2/repository/gov/cms/ab2d" 2> /dev/null
```
gsf, line 107: I think we can drop these cleanup steps given none of the runners will be around for more than a day.

.github/workflows/unit-integration-test.yml
oluwolenpbc: they keystore dependency? or the actual directory we make

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,109 @@
...
+          mvn -U clean
+
+      - name: Run unit and integration tests
+        run: |
+          mvn -s settings.xml -X -Dusername=${ARTIFACTORY_USER} -Dpassword=${ARTIFACTORY_PASSWORD} -Drepository_url=${ARTIFACTORY_URL} test -pl common,job,coverage,api,worker
+        working-directory: ${{ github.workspace }}
+
+      - name: Cleanup -
+        run: |
+          sudo rm -rf /opt/actions-runner/_work/ab2d/ab2d/opt/ab2d
```
gsf, line 109: Use `${{ github.workspace }}/opt/ab2d` instead of hardcoding the full path.

Also, this step should include an `if: ${{ !cancelled() }}` to run on both success and failure. See https://docs.github.com/en/actions/learn-github-actions/expressions#always

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,109 @@
...
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Setup Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: '17'
+          maven-version: '3.6.3'
```
gsf, line 29: Drop maven-version since it does nothing in setup-java.

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,109 @@
...
+        uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
+        env:
+          AWS_REGION: ${{ vars.AWS_REGION }}
+          ECR_REPO_ENV_AWS_ACCOUNT_NUMBER: ${{ vars.SBX_ACCOUNT_NUMBER }}
+        with:
+          params: |
+            ARTIFACTORY_URL=/artifactory/url
+            ARTIFACTORY_USER=/artifactory/user
+            ARTIFACTORY_PASSWORD=/artifactory/password
+            CC_TEST_REPORTER_ID=cc-test-reporter-id
```
gsf, line 45: This should be in aws parameter store under a path like /cc/test-reporter-id, and if it's ab2d-specific it should be in the parameters we grab in the step below.

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: temurin
+          cache: maven
+
+      - name: Download bfd keystore
+        run: |
+          aws s3 cp s3://bcda-opensbx-access-logs/ab2d_sbx_keystore /tmp/ab2d_sbx_keystore
```
gsf, line 33: I'm confused by the `${{ github.workspace }}/opt/ab2d/ab2d_bfd_keystore` in the AB2D_BFD_KEYSTORE_LOCATION above, then this downloads to /tmp/ab2d_sbx_keystore. And below, the opt/ab2d directory gets created, but when does the file get moved?

Also, why is the ab2d_sbx_keystore in the bcda-opensbx-access-logs bucket?

.github/workflows/unit-integration-test.yml
```diff
@@ -0,0 +1,97 @@
...
+
+      - name: Set env vars from AWS params
+        uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
+        env:
+          AWS_REGION: ${{ vars.AWS_REGION }}
+          ECR_REPO_ENV_AWS_ACCOUNT_NUMBER: ${{ vars.SBX_ACCOUNT_NUMBER }}
```
gsf, line 39: The ECR_REPO_ENV_AWS_ACCOUNT_NUMBER env var isn't needed here.

gsf:
I think we're about good to go. I've filled in the PR template.

gsf (approved):

PR #1300, opened 2023-12-12
WIP -[AB2D-5870] Update postgres container version to 15

Rwolfe-Nava:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-5870

## 🛠 Changes

Updates testcontainers and docker image to use postgres 15.
Installs pg_cron inside of the docker images and testcontainers.

## ℹ️ Context for reviewers

(Background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers.)

## ✅ Acceptance Validation

(How were the changes verified? Did you fully test the acceptance criteria in the ticket? Provide reproducible testing instructions and screenshots if applicable.)

## 🔒 Security Implications


common/src/test/java/gov/cms/ab2d/common/util/AB2DPostgresqlContainer.java
```diff
@@ -1,24 +1,70 @@
...
+                .run("touch /docker-entrypoint-initdb.d/init_pg_cron.sh")
+                .run("echo \"#!/bin/sh\"\n >> /docker-entrypoint-initdb.d/init_pg_cron.sh")
+                .run("echo \"# Configure PG Cron\"\n >> /docker-entrypoint-initdb.d/init_pg_cron.sh")
+                .run("echo \"echo shared_preload_libraries = 'pg_cron' >> /var/lib/postgresql/data/postgresql.conf\"\n >> /docker-entrypoint-initdb.d/init_pg_cron.sh")
+                .run("echo \"echo cron.database_name = 'test' >> /var/lib/postgresql/data/postgresql.conf\"\n >> /docker-entrypoint-initdb.d/init_pg_cron.sh")
+                .run("echo \"# Required to load pg_cron\"\n >> /docker-entrypoint-initdb.d/init_pg_cron.sh")
+                .run("echo \"pg_ctl restart\"\n >> /docker-entrypoint-initdb.d/init_pg_cron.sh")
+                .run("chmod +x /docker-entrypoint-initdb.d/init_pg_cron.sh")
+                .run("apt-get install -y curl")
+                .run("apt-get install -y postgresql-15-cron")
```
gsf, line 26: Can these lines be moved to the postgres Dockerfile, then that file be referenced with withFileFromPath?

api/Dockerfile
```diff
@@ -11,6 +11,11 @@ ENV JVM_MAX_RAM=80.0
 #Update Packages
 RUN yum update -y --security
 
+#Install PG_Cron 15
+# RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+# RUN yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
+# RUN yum install -y pg_cron_15
+
```
gsf, line 18: Are these lines needed?

gsf (commented):
Coming back to this, I worry this will have a big impact on tests -- will this result in dozens of image builds during testing? Could the postgres image instead be built and uploaded to ECR, then referenced there?

common/src/test/java/gov/cms/ab2d/common/util/AB2DPostgresqlContainer.java
Rwolfe-Nava: I am not sure. I originally attempted to have init_pg_cron.sh already exist inside the repo, and then just copy it to the container with a:

.copy("<path_to_init_pg_cron.sh>", "/docker-entrypoint-initdb.d/");

The problem I ran into was a null pointer exception with reference to the init_pg_cron.sh path. I tried both a relative path and an absolute path but neither worked. I am probably missing something, but for testing purposes I gave up and just created the script inside the container on the fly using the commands above.

Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM postgres:15.2
+RUN apt-get -y update
+RUN apt-get install -y curl 
+RUN apt-get install -y postgresql-15-cron
```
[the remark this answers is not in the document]
Rwolfe-Nava: Before this was in place, the docker-compose for ab2d used:

`image: postgres:11`
Is my understanding correct this is functionally identical to having FROM postgres:<version_number> in a Dockerfile?

I agree that we could move the Dockerfile to a subdirectory. I will implement this change.

> Also, files should end with a newline. Update than in your text editor settings.
Noted. Thank you!

Rwolfe-Nava:
> Coming back to this, I worry this will have a big impact on tests -- will this result in dozens of image builds during testing? Could the postgres image instead be built and uploaded to ECR, then referenced there?

I think it already does spin up dozens of images during testing. At least locally, when we run testcontainers it spins up a LOT of containers for all of the tests that run. In that regard, this PR doesn't change that so it should have the same impact that it currently does.

That said, I am not sure what is meant by building the image and uploading it to ECR (I don't know what ECR is so I would need to research this). If that would eliminate the current behavior of spinning up a lot of dummy test-containers that would be a great benefit in my opinion.

Dockerfile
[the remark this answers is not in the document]
Rwolfe-Nava: Done

api/Dockerfile
Rwolfe-Nava: This has been cleaned up

PR #1308, opened 2024-01-25
[AB2D-5918] Add Null Attestation Handling

Rwolfe-Nava:
## 🎫 Ticket
https://jira.cms.gov/browse/AB2D-5908
https://jira.cms.gov/browse/AB2D-5918

## 🛠 Changes

Adds a Null Check for AttestationDate in JobPreprocessorImpl. 

If Null Check Triggers:

Logs a warning to AWS
Updates job to a failed status and updates events job_status_changed table with relevant information.

## ℹ️ Context for reviewers

This is to implement a check so we can handle the case where the attestation date of a contract was null in our system and a job was started against that contract and fail the job before it starts. 

Alleviates the current issue where the job would be stuck in the submitted status, and our logs on AWS would be bombarded and clogged with NullPointer Exceptions.

## ✅ Acceptance Validation

This has been deployed to the IMPL environment and tested there:

AWS_Event_IMPL
Job_Status_Change_Screenshot
No_NullPointer_Message

## 🔒 Security Implications

### DIFF
--- a/worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
+++ b/worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
@@ -77,8 +77,8 @@ public class JobPreProcessorImpl implements JobPreProcessor {
         if (contract.getAttestedOn() == null) {
             log.warn("JobPreProcessorImpl > preprocess: job FAILED because the contract attestation date is null.");
 
-            eventLogger.logAndAlert(job.buildJobStatusChangeEvent(FAILED, EOB_JOB_FAILURE + " Job for "
-                + contract.getContractNumber() + " in progress"), PUBLIC_LIST);
+            eventLogger.logAndAlert(job.buildJobStatusChangeEvent(FAILED, EOB_JOB_FAILURE + " Job " + jobUuid
+                + "failed for contract: " + contract.getContractNumber() + " because contract attestation date is null"), PUBLIC_LIST);
 
             job.setStatus(FAILED);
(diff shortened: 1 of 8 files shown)


smirnovaae (approved):

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
```diff
@@ -72,6 +73,20 @@ public Job preprocess(String jobUuid) {
         if (contract == null) {
             throw new IllegalArgumentException("A job must always have a contract.");
         }
+
```
Sadibhatla, line 76: add here comments about the ticket number and what we are trying to achieve

worker/src/test/java/gov/cms/ab2d/worker/processor/JobPreProcessorUnitTest.java
```diff
@@ -131,12 +132,30 @@ void whenJobIsInSubmittedStatus_ThenJobShouldBePutInProgress() throws Interrupte
         verify(jobRepository).save(Mockito.any());
     }
 
+    @DisplayName("Job is FAILED if contract is not attested")
+    @Test
```
Sadibhatla, line 136: just a few line comments about mock tests also would be great for readability in future. Everything else looks great and Great job on finishing this super quickly!

worker/src/main/java/gov/cms/ab2d/worker/processor/JobPreProcessorImpl.java
Rwolfe-Nava: Will do before merge

PR #1339, opened 2024-05-31
Adding current mbi join for matching counts

Sadibhatla:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6134

## 🛠 Changes

Updated the SQL's to include join with Current MBI

## ℹ️ Context for reviewers

Database retrieved and queued counts are mismatching most likely due to some missing joins with current MBIs for opt outs

## ✅ Acceptance Validation

No pipeline is broken and jobs still run fine

## 🔒 Security Implications

### DIFF
--- a/coverage/src/main/java/gov/cms/ab2d/coverage/repository/CoverageServiceRepository.java
+++ b/coverage/src/main/java/gov/cms/ab2d/coverage/repository/CoverageServiceRepository.java
@@ -85,8 +85,9 @@ public class CoverageServiceRepository {
      *
      * The contract and year must be included to take advantage of the partitions and prevent a table scan
      */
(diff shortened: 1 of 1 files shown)


Rwolfe-Nava (approved):
This is really picky because the "AS" keyword is optional, but we use it in other places (like CoverageSearchEventRepository.java). So for consistency I would suggest adding it:

`SELECT COUNT(*) FROM coverage AS c` ...

That said, I am approving because this is fine as is.

Sadibhatla:
https://github.com/CMSgov/ab2d/blob/f6738fe245fb786296560b59741fa99ff2f170a1/coverage/src/main/java/gov/cms/ab2d/coverage/repository/CoverageServiceRepository.java 

In this file where I made the changes "AS" is not used in other places and its completely optional Richard

PR #1350, opened 2024-06-18
Ab2d-6106/tibq job

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6106
https://jira.cms.gov/browse/AB2D-6105

## 🛠 Changes

TIBQ changes for ab2d Job, Worker and API

ToDo: I will also update pom file with new bfd version, when ab2d-libs will be ready to merge

## ℹ️ Context for reviewers

Currently, AB2D API only accepts _since as a date parameter. The goal is for AB2D to be able to support the use case where a PDP could run jobs with time interval based queries so that they could download their data in batches. To achieve that, we will add a _until parameter to define the upper ceiling of the date range. 

Both the since and _until parameter will be applied to the lastupdated_Date. 

## ✅ Acceptance Validation

Tested manually on impl and prod-validation
https://impl.ab2d.cms.gov/swagger-ui/index.html

## 🔒 Security Implications


api/src/main/java/gov/cms/ab2d/api/controller/common/ApiCommon.java
```diff
@@ -87,14 +81,27 @@ public void checkSinceTime(OffsetDateTime date) {
         if (date.isAfter(OffsetDateTime.now())) {
             throw new InvalidClientInputException("You can not use a time after the current time for _since");
         }
-        try {
-            OffsetDateTime ed = OffsetDateTime.parse(SINCE_EARLIEST_DATE, ISO_DATE_TIME);
-            if (date.isBefore(ed)) {
-                log.error("Invalid _since time received {}", date);
-                throw new InvalidClientInputException("_since must be after " + ed.format(ISO_OFFSET_DATE_TIME));
-            }
-        } catch (Exception ex) {
```
smirnovaae, line 96: This try catch makes no sense

api/src/main/java/gov/cms/ab2d/api/controller/common/ApiCommon.java
```diff
@@ -87,14 +81,27 @@ public void checkSinceTime(OffsetDateTime date) {
         if (date.isAfter(OffsetDateTime.now())) {
             throw new InvalidClientInputException("You can not use a time after the current time for _since");
         }
-        try {
-            OffsetDateTime ed = OffsetDateTime.parse(SINCE_EARLIEST_DATE, ISO_DATE_TIME);
```
smirnovaae, line 91: I added new constant SINCE_EARLIEST_DATE_TIME with OffsetDateTime type. In this case, we don't need to parse string every time

pom.xml
```diff
@@ -45,7 +45,7 @@
 
         <!-- AB2D lib versions -->
         <events-client.version>1.12.6</events-client.version>
-        <bfd-lib.version>2.1.3</bfd-lib.version>
+        <bfd-lib.version>2.2.0</bfd-lib.version>
```
smirnovaae, line 48: version will be 2.3.0

Sadibhatla (approved):
Looks great and lets test it thorougly

PR #1351, opened 2024-06-21
AB2D-6157 Update mismatch job failure to pass in slack alerts

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6157

## 🛠 Changes

Added threshold for the alert

## ℹ️ Context

Currently slack shows jobs have failed due to a mismatch in data that is cause by the Null MBI issue. 

We'd like to update the slack alert to only show a failure if the mis matches reach a threshold higher that 10. Otherwise the job should show as a success as the job does complete for the PDP to receive the data.


coilysiren (commented):
The PR description and ticket both mention `10`, but the code has a `35`?

smirnovaae:
> The PR description and ticket both mention `10`, but the code has a `35`?

yes, you're right. I'm going to double check with Anna H.

coilysiren (approved):
Approved per convo in ab2d slack 🙏🏼

PR #1354, opened 2024-06-25
AB2D-6157 Update mismatch job failure to pass in slack alerts

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6157

## 🛠 Changes

Disabled mismatch alert if threshold > 35

## ℹ️ Context

Follow up for https://github.com/CMSgov/ab2d/pull/1351

Currently slack shows jobs have failed due to a mismatch in data that is cause by the Null MBI issue.

We'd like to update the slack alert to only show a failure if the mis matches reach a threshold higher that 10. Otherwise the job should show as a success as the job does complete for the PDP to receive the data.

## 🧪 Validation
 
Unit tests passed


coilysiren (approved):
👍🏼 !

PR #1363, opened 2024-07-11
[AB2D-6188] increase `common` test coverage

coilysiren:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6188

## 🛠 Changes

Adds tests for getting test coverage on the `PdpClientServiceImpl` class

## ℹ️ Context

This is a part of my journey to achieve 90% test coverage everywhere

### DIFF
--- a/common/src/main/resources/db/changelog/v2021/add_synthea_contracts.sql
+++ b/common/src/main/resources/db/changelog/v2021/add_synthea_contracts.sql
@@ -12,6 +12,7 @@ insert into contract (id, contract_number, contract_name, attested_on, update_mo
 insert into contract (id, contract_number, contract_name, attested_on, update_mode, created, modified) values ((select nextval('hibernate_sequence')), 'Z1007', 'Z1007', '2020-03-01', 'TEST', NOW(), NOW());
 insert into contract (id, contract_number, contract_name, attested_on, update_mode, created, modified) values ((select nextval('hibernate_sequence')), 'Z1008', 'Z1008', '2020-03-01', 'TEST', NOW(), NOW());
 insert into contract (id, contract_number, contract_name, attested_on, update_mode, created, modified) values ((select nextval('hibernate_sequence')), 'Z1009', 'Z1009', '2020-03-01', 'TEST', NOW(), NOW());
+insert into contract (id, contract_number, contract_name, attested_on, update_mode, created, modified) values ((select nextval('hibernate_sequence')), 'Z1010', 'Z1010', '2020-03-01', 'TEST', NOW(), NOW());
 
 insert into user_account(id, client_id, enabled, max_parallel_jobs, contract_id, organization) values((select nextval('hibernate_sequence')), '0oa9jyx2w9Z0AntLE297', true, 1000, (select id from contract where contract_number='Z1001' limit 1), 'PDP-1001');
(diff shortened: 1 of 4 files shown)


common/src/main/resources/db/changelog/v2021/add_synthea_contracts.sql
```diff
@@ -32,9 +34,10 @@ insert into user_role(user_account_id, role_id) values ((select id from user_acc
 insert into user_role(user_account_id, role_id) values ((select id from user_account where client_id='0oa9jyzerqYlfXtsQ297' limit 1), (select id from role where name='SPONSOR' limit 1));
 insert into user_role(user_account_id, role_id) values ((select id from user_account where client_id='0oa9jyy4ou5c1MJ3N297' limit 1), (select id from role where name='SPONSOR' limit 1));
 insert into user_role(user_account_id, role_id) values ((select id from user_account where client_id='0oa9jyw49f9U19vvB297' limit 1), (select id from role where name='SPONSOR' limit 1));
+insert into user_role(user_account_id, role_id) values ((select id from user_account where client_id='0oa9jyw49f9U20vvB297' limit 1), (select id from role where name='ADMIN' limit 1));
```
coilysiren, line 37: Setting this new line's role name to `ADMIN` gives us test coverage on the `return false` case inside of the [PdpClientServiceImpl.hasSponsorRole](https://github.com/CMSgov/ab2d/blob/3e23a031c72373523c4e79191b69f35604642b0d/common/src/main/java/gov/cms/ab2d/common/service/PdpClientServiceImpl.java#L53-L61) function.

The problem is... I don't know where this data is loaded, or how it is supposed to be used. It is only test data? If so, then why isn't it inside of the test folder? Is it perhaps sandbox data? If so, we me adding fake data in here possibly break something in sandbox? I unfortunately really don't know much about this system, so I can't project the impact of this change.

All I know, is that adding this line gives us more test coverage. And its the only way to get test coverage on that line, as far as I can tell. So here I am, trying to figure out if this change is safe or not 🤷🏼

common/src/main/resources/db/changelog/v2024/add_test_data.sql
```diff
@@ -0,0 +1,10 @@
+--liquibase formatted sql
+--  -------------------------------------------------------------------------------------------------------------------
+
+--changeset lsharshar:AB2D-3874 failOnError:true
+
+insert into Contract.contract (id, contract_number, contract_name, attested_on, update_mode, created, modified) values ((select nextval('hibernate_sequence')), 'Z1010', 'Z1010', '2020-03-01', 'TEST', NOW(), NOW());
+
+insert into user_account(id, client_id, enabled, max_parallel_jobs, contract_id, organization) values((select nextval('hibernate_sequence')), '0oa9jyw49f9U20vvB297', true, 1000, (select id from Contract.contract where contract_number='Z1010' limit 1), 'PDP-1010');
+
+insert into user_role(user_account_id, role_id) values ((select id from user_account where client_id='0oa9jyw49f9U20vvB297' limit 1), (select id from role where name='ADMIN' limit 1));
```
coilysiren, line 10: This client_id is totally random

Sadibhatla (approved):

PR #1385, opened 2024-08-08
Rename current mbi  opt out column name

Sadibhatla:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6151

## 🛠 Changes

We should update the column name to data_sharing or share_data:
column name of opt_out is updated to share_data
no reduction in test coverage
 

## ℹ️ Context

1-800 opt out was rolled out on 4/6. The data sharing preference field in ICD can be true or false (or blank): When it is true, it means the beneficiary has chosen to share their data; when it is false, it means beneficiary has chosen to not share their data. The data sharing preference (specific in ICD) field is currently mapped to opt_out, which causes confusion as the column name contradicts the meaning of the field. column name of opt_out is updated to share_data

## 🧪 Validation

Verify to make sure that changes update the table column name .
image

### DIFF
--- a/common/src/main/resources/db/changelog/v2024/ab2d_6151_rename_column.sql
+++ b/common/src/main/resources/db/changelog/v2024/ab2d_6151_rename_column.sql
@@ -1,0 +1,11 @@
+DO $$
+BEGIN
+    IF EXISTS (SELECT 1
+     FROM information_schema.columns
+     WHERE table_schema = 'public' AND table_name = 'current_mbi' AND
+column_name = 'opt_out_flag')
+    THEN
+ALTER TABLE public.current_mbi RENAME opt_out_flag TO share_data;
+END IF;
+END
+$$;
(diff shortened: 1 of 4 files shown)


coilysiren (approved):

common/src/main/resources/db/changelog/v2024/ab2d_6151_rename_column.sql
```diff
@@ -0,0 +1,11 @@
...
+BEGIN
+    IF EXISTS (SELECT 1
+     FROM information_schema.columns
+     WHERE table_schema = 'public' AND table_name = 'current_mbi' AND
+column_name = 'opt_out_flag')
+    THEN
+ALTER TABLE public.current_mbi RENAME opt_out_flag TO share_data;
+END IF;
+END
+$$;
```
smirnovaae, line 11: It also need to be renamed for `update_current_mbi` procedure. May be manually?

Sadibhatla: We dont use "update_current_mbi " anymore remember?  proc_insert_mbi_to_table is what we have the pg_cron set for. that doesnt care for opt_out_flag at all

smirnovaae: ah, right! Impl still has the good old "update_current_mbi" and this confused me

smirnovaae (approved):

PR #1391, opened 2024-08-27
Update SecurityConfig.java

feralearthman:
Add "/metadata" to authExceptions so that you don't need a bearer token to get a CapabilityStatement.

[## 🎫 Ticket](https://jira.cms.gov/browse/AB2D-6253)

## 🛠 Changes
Changed one file: SecurityConfig.java. Add "/metadata" to authExceptions.
This makes it so that a bearer token is not required to access the metadata endpoint.

## ℹ️ Context
The Inferno and Touchstone both try to access the /metadata endpoint in their bulk IG test suite. Currently this only works if you pass in a bearer token, which they don't. This change should make those conformance tests work.

## 🧪 Validation
Once this is deployed to the ab2d sandbox, see if a request to the metadata endpoint works without a bearer token. This will be addressed in https://jira.cms.gov/browse/AB2D-6286


Sadibhatla (approved):

feralearthman:
A "/metadata" to authExceptions in SecurityConfig.java.

PR #1393, opened 2024-08-31
PLT-228 Add build and deploy workflows

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-228

## 🛠 Changes

Added deploy and pull request workflows and updated test workflows for reuse.

## ℹ️ Context

These workflows should set up pull requests to trigger this sequence of workflows:

- unit and integration tests
- deploy api and worker to test (impl) environment
- run e2e tests against the test (impl) environment

## 🧪 Validation

See checks on this PR. Running deploys and e2e tests against other environments (dev and sbx) will need to be further tested once this is merged.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -82,7 +82,7 @@ jobs:
       - name: Create opt/ab2d directory and download keystore
         run: |
           mkdir -p opt/ab2d
-          KEYSTORE_FILE_NAME="ab2d_${{ inputs.environment == "test" && "imp" || inputs.environment }}_keystore"
+          KEYSTORE_FILE_NAME="ab2d_${{ inputs.environment == 'test' && 'imp' || inputs.environment }}_keystore"
           aws s3 cp s3://${{ vars[format('{0}_DEPLOYMENT_ENV', inputs.environment)] }}-main/$KEYSTORE_FILE_NAME $AB2D_BFD_KEYSTORE_LOCATION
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -70,7 +82,8 @@ jobs:
       - name: Create opt/ab2d directory and download keystore
         run: |
           mkdir -p opt/ab2d
-          aws s3 cp s3://ab2d-east-impl-main/ab2d_imp_keystore $AB2D_BFD_KEYSTORE_LOCATION
+          KEYSTORE_FILE_NAME="ab2d_${{ inputs.environment == 'test' && 'imp' || inputs.environment }}_keystore"
```
Sadibhatla, line 85: Shoudlnt we be consistent with "Impl" usage.. in som places "Imp" and some other place "Impl" is a lil confusing. Minor change if thats okay with you

Sadibhatla (approved):
Would like to test running e2e tests especially before disabling the Jenkins pipeline for AB2D

.github/workflows/e2e-test.yml
Sadibhatla: Also  what do you think about a check for valid/passed E2E pipeline is a pre-requisite to deploy?

gsf: Yes we could improve consistency in the naming of that keystore in a future ticket.

We need to deploy to impl/test before running the e2e test, so it can't be a pre-requisite. We should have a passing e2e test before running the release pipeline. We could discuss ways to require that.

PR #1400, opened 2024-10-04
PLT-679 Tag commits to main for promotion

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-679

## 🛠 Changes

Added workflow to trigger bulids on push to main and updated build workflow to push "main" tags to ECR.

## ℹ️ Context

We need images we can count on being around for a bit from each to push to main for promotion to prod on release. This allows us to set a lifecycle policy specific to tags with the "main" prefix. 

## 🧪 Validation

See runs at https://github.com/CMSgov/ab2d/actions/runs/11195118631, https://github.com/CMSgov/ab2d/actions/runs/11195323158, and https://github.com/CMSgov/ab2d/actions/runs/11202145951 for successful tests of the release and promote workflows. The push-main workflow (and expected tag updates in ECR) must be checked following merge of this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,7 +84,7 @@ jobs:
 
           # Push to special tag for promotion if this is run on a push to main
           if [ "$GITHUB_REF" == "refs/heads/main" ]; then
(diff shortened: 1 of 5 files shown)


ildesenesence (approved):
Reviewed - I like how you cleaned up the environment variables in here on top of the functionality changes, much more legible!

PR #1402, opened 2024-10-16
Adding verify scripts and deployment verification workflow

oluwolenpbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-642

## ℹ️ Context

This is a new addition to our github actions workflows to verify our deployments

## 🧪 Validation

The test will be validated by github actions test runs.

### DIFF
--- a/.github/workflows/verify-deploy.yml
+++ b/.github/workflows/verify-deploy.yml
@@ -1,0 +1,47 @@
+name: Verify deployment
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        required: true
+        type: string
+      module:
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      environment:
+        required: true
+        type: choice
+        options:
+          - sbx-sandbox
+          - east-prod
+          - east-impl
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+
+      - name: Set script file based on environment
+        run: |
+          if [[ "${{ github.event.inputs.environment }}" == "sbx-sandbox" ]]; then
+            script="verify-deployment-ab2d-sbx-sandbox.sh"
+          elif [[ "${{ github.event.inputs.environment }}" == "east-prod" ]]; then
+            script="verify-deployment-ab2d-east-prod.sh"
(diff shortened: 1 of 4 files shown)


gsf (commented):
Seems like a lot of duplication to have a separate verify script for each environment. Can it be parameterized?

.github/workflows/verify-deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Assume role in AB2D account for this environment
+        uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502
+        env:
+          ACCOUNT: ${{ inputs.environment == '' && 'test' || inputs.environment }}
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets[format('{0}_ACCOUNT_ID', env.ACCOUNT)] }}:role/delegatedadmin/developer/ab2d-${{ env.ACCOUNT }}-github-actions
+          
+      - name: Set environment-specific variables
```
gsf, line 43: I don't understand why these env vars are set in a separate step from the two below. They could all be one step and it would be more readable. But if this works then fine to leave as is.

.github/workflows/verify-deploy.yml
```diff
@@ -0,0 +1,77 @@
+name: Verify deployment
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
gsf, line 7: This job should be added to the pull-request and push-main orchestration jobs rather than triggering here. It should run last and depend on the e2e-test in both cases.

.github/workflows/verify-deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+  workflow_call:
+    inputs:
+      environment:
+        required: false
+        type: string
+      module:
+        required: true
+        type: string
```
gsf, line 15: No need for the module input

gsf:
I'm still seeing what looks like an error in the latest run: https://github.com/CMSgov/ab2d/actions/runs/12837809052/job/35803223841?pr=1402#step:6:76

gsf (approved):
Looks good! I'd still like to see the output of those curl commands in the logs (via tee or something), and I'd prefer we pull BASIC_AUTH from the parameter store, but that can be changed later.

PR #1434, opened 2025-01-16
PLT-683 publish ab2d-libs

oluwolenpbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-683

## 🛠 Changes

Creating a new github workflow for AB2D-Libs build

## ℹ️ Context

We are moving AB2D libs build workflow to github actions from jenkins

## 🧪 Validation

This will be tested in this PR

### DIFF
--- a/.github/workflows/publish-libs.yml
+++ b/.github/workflows/publish-libs.yml
@@ -1,0 +1,55 @@
+name: publish
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        required: true
+        type: string
+      module:
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      environment:
+        required: true
+        type: choice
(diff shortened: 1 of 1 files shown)


gsf (approved):
Looks great! Would you mind adding some more detail on issues you ran into and what the trick was to get past them? Either in the context on this PR or in comments on the ticket.

oluwolenpbc:
There were issues with hitting the correct repository. The job was defaulting to maven repo which does not have our CMS jars. To fix this I had to add the below env calls. The job was not picking it up by default from the github env.
```
            -Dusername="${ARTIFACTORY_USER}" \
            -Dpassword="${ARTIFACTORY_PASSWORD}" \
            -Drepository_url="${ARTIFACTORY_URL}"
```

PR #1450, opened 2025-03-12
remove on pull request

oluwolenpbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-683

## 🛠 Changes

This runs on pull request we only want it to run on dispatch

## ℹ️ Context

We do not want this to run every time a pull request is made. Follow-up on #1434.

## 🧪 Validation

It will not run on this pull request

### DIFF
--- a/.github/workflows/publish-libs.yml
+++ b/.github/workflows/publish-libs.yml
@@ -1,7 +1,6 @@
 name: publish
 
 on:
-  pull_request:
   workflow_call:
     inputs:
       environment:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-libs.yml
```diff
@@ -1,7 +1,6 @@
 name: publish
```
gsf, line 1:
```suggestion
name: publish libs
```

gsf (approved):

PR #1440, opened 2025-02-14
Merge 'file-compression' branch into main (rebased after Spring Boot upgrade)

bennavapbc:
## 🎫 Ticket

- https://jira.cms.gov/browse/AB2D-6462
- https://jira.cms.gov/browse/AB2D-6505

## Related PR

- https://github.com/CMSgov/ab2d/pull/1429
- https://github.com/CMSgov/ab2d/pull/1456 (Branch merged into this PR instead of `main`)

## 🛠 Changes

- Add utilities for compressing job output files to `gzip` format
- Allow bulk download API to return compressed or uncompressed files (per request header)
- Merge **file-compression** branch into branch based on **AB2D-6340/Upgrade_spring**

## ℹ️ Context
- Compress job output files to reduce file size by ~90% and (slightly) improve transfer times for clients requesting compressed files

## SonarQube
image

## 🧪 Validation
TODO -- Need to re-test

PR #1456, opened 2025-03-25
Fix error response

bennavapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6579

## 🛠 Changes

Fix error handling related to the download API so that error messages are displayed to the user. Currently, a blank response is returned to the user. 

## ℹ️ Context

Currently, the download API fails to return error message to the user. The server sends a blank response and HTTP 500. 

For reference, the status API returns a helpful error message if e.g. a job is not found:
image

If an exception is thrown from the status API, the `ErrorHandler` catches the exception and returns an error response as expected. 

If an exception is thrown from the download API, the `ErrorHandler` does **not** catch the exception. The exception bubbles up until a `ServletException` is thrown. 
- The reason for this is not clear. The download and status APIs have similar configuration. The only difference is the download API method accepts a `HttpServletResponse response` which is written to directly. Spring might exclude methods like this from error handling?   

## 🧪 Validation

Deployed to IMPL and verified the following download scenarios return helpful error messages:
- Valid job ID but non-existent file
- Non-existent job ID
- File downloaded the max number of times

Screenshot 2025-03-25 at 11 39 29 AM

Screenshot 2025-03-25 at 11 39 54 AM

Screenshot 2025-03-25 at 11 42 53 AM

### DIFF
--- a/api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
+++ b/api/src/main/java/gov/cms/ab2d/api/controller/ErrorHandler.java
@@ -26,7 +26,6 @@ import java.util.stream.Collectors;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import lombok.extern.slf4j.Slf4j;
-import lombok.val;
 import org.apache.commons.lang3.exception.ExceptionUtils;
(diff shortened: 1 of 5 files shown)


Sadibhatla (approved):

bennavapbc:
Instead of merging to `main`, this code is now part of PR #1440

PR #1458, opened 2025-04-03
Upgrade to spring boot 3.4.4

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6625

## 🛠 Changes

Upgraded:

- Spring boot version from 3.2.8 to 3.4.4
- Spring cloud aws from 3.1.1 to 3.3.0
- Liquibase to 4.31.1
- TestContainers to 1.20.6

## ℹ️ Context

Changes are required as part of the work to upgrade the FHIR HAPI version across all AB2D repositories.

### DIFF
--- a/common/src/main/resources/db/changelog/db.changelog-master.yaml
+++ b/common/src/main/resources/db/changelog/db.changelog-master.yaml
@@ -168,4 +168,6 @@ databaseChangeLog:
   - include:
       file: db/changelog/v2024/ab2d_6151_rename_column.sql
   - include:
-      file: db/changelog/v2025/add_sequence.sql
+      file: db/changelog/v2025/add_sequences.sql
+  - include:
+      file: db/changelog/v2025/alter_sequences.sql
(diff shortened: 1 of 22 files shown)


common/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -168,4 +168,6 @@ databaseChangeLog:
   - include:
       file: db/changelog/v2024/ab2d_6151_rename_column.sql
   - include:
-      file: db/changelog/v2025/add_sequence.sql
+      file: db/changelog/v2025/add_sequences.sql
```
bennavapbc, line 171: Will renaming the file / changing file contents cause problems for environments where this SQL change was already applied?

smirnovaae: Changing content with renaming is ok. It will be just a new record in databasechangelog table.
If we change content without renaming, yes, we will have liquibase exception after deployment (which is annoying, but easy to fix)

bennavapbc (approved):

PR #1483, opened 2025-05-19
PLT-1083 Drop gitleaks workflow

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1083

## 🛠 Changes

Dropped the gitleaks workflow.

## ℹ️ Context

@SJWalter11 Tagging you in here since this does drop one layer of gitleaks checking, but as noted in the JIRA ticket, it should not be needed considering the commit hook and push protection. Note also that this workflow would only catch sensitive data after it had been committed and pushed to the public repo, not beforehand as is the case with the commit hook and push protection.

## 🧪 Validation

Gitleaks workflow should no longer run on push.


bennavapbc (approved):

SJWalter11 (approved):
The removal an movement to an inline process to catch prior to post sounds good. The code update looks to remove the post catch.

PR #1521, opened 2025-07-01
 PLT-1164 adding aurora db resources to db module

oluwolenpbc:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1164

## 🛠 Changes

We are updating the AB2D datastore module to use amazon aurora

## ℹ️ Context

This change is being made to save money on our database cost. Our RDS database instances were bloated and inefficient moving to aurora will give us about a 40% saving.

## 🧪 Validation

This has been tested, planned, and applied in both dev and test environments.

### DIFF
--- a/ops/services/20-microservices/main.tf
+++ b/ops/services/20-microservices/main.tf
@@ -126,6 +126,8 @@ data "aws_iam_policy_document" "this" {
   }
 }
 
+data "aws_caller_identity" "current" {}
+
 resource "aws_lb" "internal_lb" {
   name               = "${local.service_prefix}-${local.service}"
   internal           = true
@@ -138,8 +140,7 @@ resource "aws_lb" "internal_lb" {
   drop_invalid_header_fields       = true
 
   access_logs {
-    bucket  = local.network_access_logs_bucket
-    prefix  = "${local.service_prefix}-${local.service}"
+    bucket  = "cms-cloud-${data.aws_caller_identity.current.account_id}-us-east-1"
     enabled = true
   }
 }
--- a/ops/services/30-api/main.tf
+++ b/ops/services/30-api/main.tf
@@ -341,6 +341,8 @@ resource "aws_cloudwatch_metric_alarm" "health" {
   }
 }
 
+data "aws_caller_identity" "current" {}
+
 resource "aws_lb" "ab2d_api" {
   #TODO Consider using name_prefix for ephemeral environments... thhey may only be up to 6-characters
   name               = "${local.service_prefix}-api"
@@ -362,8 +364,7 @@ resource "aws_lb" "ab2d_api" {
   drop_invalid_header_fields       = true
 
   access_logs {
-    bucket  = local.network_access_logs_bucket
-    prefix  = "${local.service_prefix}-${local.service}"
+    bucket  = "cms-cloud-${data.aws_caller_identity.current.account_id}-us-east-1"
     enabled = true
   }
 }


juliareynolds-nava (dismissed):

ops/services/10-core/database.tf
```diff
@@ -244,6 +244,21 @@ resource "aws_cloudwatch_metric_alarm" "db_swap_usage" {
   }
 }
 
+resource "aws_cloudwatch_metric_alarm" "aurora_cpu_utilization" {
```
mjburling, line 247: Now that we're beyond merely transitioning resources from legacy to greenfield, we can be more discerning about alerts that yield to an alarm and ensure that they are actionable. What do we expect an operator or on-call to do when they encounter this alarm?

ops/tf-modules/ds/main.tf
```diff
@@ -110,3 +110,48 @@ resource "aws_vpc_security_group_egress_rule" "this" {
   ip_protocol       = "-1"
   security_group_id = aws_security_group.this.id
 }
+
+resource "aws_rds_cluster" "aurora" {
```
mjburling, line 114:
```suggestion
resource "aws_rds_cluster" "this" {
```

ops/tf-modules/ds/main.tf
```diff
@@ -110,3 +110,48 @@ resource "aws_vpc_security_group_egress_rule" "this" {
...
+  cluster_identifier      = "${local.service_prefix}-aurora"
+  engine                  = "aurora-postgresql"
+  engine_version          = "16.2"
```
mjburling, line 117:
```suggestion
  engine_version          = "16.8"
```

ops/tf-modules/ds/main.tf
```diff
@@ -110,3 +110,48 @@ resource "aws_vpc_security_group_egress_rule" "this" {
...
+  preferred_backup_window = var.backup_window
+  apply_immediately       = true
+  skip_final_snapshot     = true
+  deletion_protection     = var.deletion_protection
+  tags = {
+    AWS_Backup = "4hr7_w90"
+  }
+}
+
+resource "aws_rds_cluster_instance" "aurora_instance" {
```
mjburling, line 140:
```suggestion
resource "aws_rds_cluster_instance" "this" {
```

ops/tf-modules/ds/main.tf
```diff
@@ -110,3 +110,48 @@ resource "aws_vpc_security_group_egress_rule" "this" {
...
+  identifier              = "${local.service_prefix}-aurora-instance-1"
+  cluster_identifier      = aws_rds_cluster.aurora.id
```
mjburling, line 142:
```suggestion
  cluster_identifier      = aws_rds_cluster.this.id
```

ops/tf-modules/ds/main.tf
```diff
@@ -110,3 +110,48 @@ resource "aws_vpc_security_group_egress_rule" "this" {
...
+  apply_immediately       = true
+  skip_final_snapshot     = true
+  deletion_protection     = var.deletion_protection
+  tags = {
+    AWS_Backup = "4hr7_w90"
+  }
+}
+
+resource "aws_rds_cluster_instance" "aurora_instance" {
+  identifier              = "${local.service_prefix}-aurora-instance-1"
```
mjburling, line 141: Merely a suggestion, but this probably ought to be 0-indexed.
```suggestion
  identifier              = "${local.service_prefix}-aurora-instance-0"
```

ops/tf-modules/ds/main.tf
```diff
@@ -110,3 +110,48 @@ resource "aws_vpc_security_group_egress_rule" "this" {
...
+  cluster_identifier      = aws_rds_cluster.aurora.id
+  instance_class          = var.instance_class
+  engine                  = aws_rds_cluster.aurora.engine
+  engine_version          = aws_rds_cluster.aurora.engine_version
+  db_subnet_group_name    = aws_db_subnet_group.this.name
+  publicly_accessible     = false
+  monitoring_interval     = var.monitoring_interval
+  apply_immediately       = true
```
mjburling, line 149: Seems like this should generally be set to `false`, but potentially overridable. I'm not sure what kinds of changes we'd want/need to make, but thinking of the production cluster, it might be wise to wait for a maintenance window to apply some settings? 🤔

ops/tf-modules/ds/main.tf
```diff
@@ -110,3 +110,48 @@ resource "aws_vpc_security_group_egress_rule" "this" {
...
+  auto_minor_version_upgrade = true
+  performance_insights_enabled = true
+  performance_insights_kms_key_id = null
+  monitoring_role_arn     = null
```
mjburling, line 153: This needs some additional thought. This can't be adjusted after the cluster is instantiated. I think there should be a specific monitoring role for AB2D established _somewhere_ and I think it should be using the CMK for the given AB2D environment, rather than deferring to the default key.

ops/tf-modules/ds/main.tf
```diff
@@ -110,3 +110,48 @@ resource "aws_vpc_security_group_egress_rule" "this" {
...
+    AWS_Backup = "4hr7_w90"
+  }
+}
+
+resource "aws_rds_cluster_instance" "aurora_instance" {
+  identifier              = "${local.service_prefix}-aurora-instance-1"
+  cluster_identifier      = aws_rds_cluster.aurora.id
+  instance_class          = var.instance_class
+  engine                  = aws_rds_cluster.aurora.engine
+  engine_version          = aws_rds_cluster.aurora.engine_version
```
mjburling, line 145:
```suggestion
  engine                  = aws_rds_cluster.this.engine
  engine_version          = aws_rds_cluster.this.engine_version
```

mjburling (changes requested):
👋 Looked this over and it's trending in the right direction. In addition to some of the in-line comments, we're missing a few other details. Here's where we can start:
* we'll need a defined monitoring role (potentially here but probably something defined as an input?) that can use the CMK for a given application/environment allowing us to define the `performance_insights_kms_key_id` here
* this needs both an [aws_rds_cluster_parameter_group](https://registry.terraform.io/providers/hashicorp/aws/5.100.0/docs/resources/rds_cluster_parameter_group) and [aws_db_parameter_group](https://registry.terraform.io/providers/hashicorp/aws/5.100.0/docs/resources/db_parameter_group) defined and used within the `aws_rds_cluster` and `aws_rds_cluster_instance`, respectively.
* `aws_rds_cluster` must accept a [storage_type](https://registry.terraform.io/providers/hashicorp/aws/5.100.0/docs/resources/rds_cluster#storage_type-1) variable and `10-core` probably ought to use 
* This solution needs to be able to accommodate a snapshot as in input for instantiating the cluster.

Alternatively, you might look at BFD's [02-database](https://github.com/CMSgov/beneficiary-fhir-data/blob/master/ops/services/02-database/main.tf) terraservice for a complete example. However, that solution enables some dynamic scaling which this application can't accommodate quite yet and really... may never need.

bennavapbc (approved):

PR #1522, opened 2025-07-03
Fix impl db name

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/...

## 🛠 Changes

Changed db name from test to impl and from sandbox to sbx in action.yml

## ℹ️ Context

Fix for psql: error: connection to server at "ab2d-sandbox.ckhq6co204yw.us-east-1.rds.amazonaws.com" (10.157.214.105), port 5432 failed: FATAL:  database "sandbox" does not exist
Unable to retrieve organization


Sadibhatla (approved):
Thank you for the fix

PR #1554, opened 2025-08-04
AB2D-6842: Update "Uptime metrics"

Sadibhatla:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6842

## 🛠 Changes

Update QuickSight dataset to use an acceptable alternative to `nextval()` against the Aurora reader endpoint.

## ℹ️ Context

Fix: Avoid nextval() usage on Aurora read-only replicas
Aurora read-only replicas do not support nextval() calls, as sequence operations require write access to the underlying sequence object. Attempting to call nextval() on a replica  causing errors in dataset:

ERROR: cannot execute nextval() in a read-only transaction
This PR modifies the application logic to avoid invoking nextval() in read-only environments. 
Change:
Identified and removed/redirected all calls to nextval() on read-only connections.

## 🧪 Validation

The SQL queries run as expected after the changes made to remove any nextVal() references

### DIFF
--- a/ops/services/40-insights/README.md
+++ b/ops/services/40-insights/README.md
@@ -10,7 +10,7 @@
 
 | Name | Version |
 |------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | 5.100.0 |
+| <a name="provider_aws"></a> [aws](#provider\_aws) | ~> 5 |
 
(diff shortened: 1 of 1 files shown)


Sadibhatla:
@mjburling : Updated the README,  Thank you for applying the changes already !  Asking for re-review

PR #1605, opened 2025-09-10
AB2D-6904 Encrypt ECS cluster storage with CMK

mjburling:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-6904

## 🛠 Changes
This change makes use of the CDAP-managed cluster module instead of directly configuring `aws_ecs_cluster` resources and effectively configures the clusters to use the prescriptive, environment-specific CMK for each cluster.

## ℹ️ Context
Post-greenfield migration, post-fargate migration cleanup.

## 🧪 Validation
This was applied to `dev` and `test` environments. We encountered some unrelated issues, particularly in `test`, but those have since been remedied. The necessary imports have already been executed which makes this a very low risk change taking effect on the next deployment to `sandbox` and `prod` environments.

### DIFF
--- a/.github/workflows/unit-integration-test.yml
+++ b/.github/workflows/unit-integration-test.yml
@@ -74,6 +74,8 @@ jobs:
           mvn -ntp -U clean
 
       - name: Run unit and integration tests
+        env:
+          MAVEN_OPTS: "-Xmx1G -Xms1G"
         run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/unit-integration-test.yml
```diff
@@ -74,6 +74,8 @@ jobs:
           mvn -ntp -U clean
 
       - name: Run unit and integration tests
+        env:
+          MAVEN_OPTS: "-Xmx1G -Xms1G"
```
mjburling, line 78: This is entirely unrelated to the change, but I did receive at least one OOM as part of the regular CI checks. I'd be happy to pull this out of the changeset if this or something similar will land elsewhere.

Sadibhatla (approved):

PR #1749, opened 2026-04-14
AB2D-7180 lambda arm64 migration

keeyanghoreshi-rgb:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7180

## 🛠 Changes

Updates architecture for the lambdas to use arm64, and also fixes a minor issue with a handler.

## ℹ️ Context

Part of a greater initiative to migrate coderunners to run/build on arm64 architecture. See [here](https://confluence.cms.gov/spaces/ODI/pages/1527722964/DevSecOps_Strategy_ARM64) for more info.

The lambdas run natively on arm64.

## 🧪 Validation

https://github.com/CMSgov/ab2d/actions/runs/24409711016
https://github.com/CMSgov/ab2d/actions/runs/24408903342

The deploy works, and the architecture of the lambdas is converted to arm64, checked with:
aws lambda get-function-configuration --function-name <lambda name> --query "Architectures"

### DIFF
--- a/.github/workflows/deploy-lambda.yml
+++ b/.github/workflows/deploy-lambda.yml
@@ -38,7 +38,7 @@ env:
 
 jobs:
   deploy:
-    runs-on: codebuild-ab2d-${{github.run_id}}-${{github.run_attempt}}
+    runs-on: codebuild-ab2d-${{ contains(fromJSON('["dev", "test"]'), env.AB2D_ENV) && 'non-prod' || 'prod'}}-${{github.run_id}}-${{github.run_attempt}}
     env:
       AWS_REGION: ${{ vars.AWS_REGION }}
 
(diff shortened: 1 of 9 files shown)


bennavapbc:
@keeyanghoreshi-rgb Can you bump the version from `1.1.3` to `1.1.4`, run the lambda publish workflow, and then run deploy lambda workflow with `1.1.4`?

.github/workflows/deploy-lambda.yml
```diff
@@ -38,7 +38,7 @@ env:
 
 jobs:
   deploy:
-    runs-on: codebuild-ab2d-${{github.run_id}}-${{github.run_attempt}}
+    runs-on: codebuild-ab2d-${{ contains(fromJSON('["dev", "test"]'), inputs.environment) && 'non-prod' || 'prod'}}-${{github.run_id}}-${{github.run_attempt}}
```
bennavapbc, line 41: What's the reason for this? Some of the updated workflows lack the conditional based on environment. Wondering if we also need to update these. 

Image

keeyanghoreshi-rgb: These workflows don't need the prod runner, since they're hardcoded to use non-prod credentials. For example in `build.yml` the account is set as `AWS_ACCOUNT: ${{ secrets.NON_PROD_ACCOUNT }}`, so the runner will always be non-prod. This is also true of other workflows that restrict the environment input to dev/test.

Sadibhatla (approved):

Sadibhatla:
Please update the branch Keeyan

smirnovaae (approved):

bennavapbc (approved):
Looks good, though we want to make sure there are no runtime errors when the lambda execute. It'll be a good idea to keep an eye on the CloudWatch logs.

PR #1750, opened 2026-04-14
AB2D-7205 Add necessary changes for ephemeral environments

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7205

## 🛠 Changes

Adds the following:
- BFD truststore parameters
- Updated naming for IDR IAM roles
- Added security group management of orphaned microservices into `30-lambda` as it still had a dependency

## ℹ️ Context

These changes are necessary for standing up an ephemeral environment

## 🧪 Validation

Environment was stood up and stable, and E2E tests passed pointed at the ephemeral ALB.
https://github.com/CMSgov/ab2d/actions/runs/24367112711

Will need to import `aws_security_group.microservices_lb` on tofu-apply of `30-lambda` in non-ephemeral environments as it already exists in those environments
```
tofu import aws_security_group.microservices_lb <security_group_id>
tofu import aws_vpc_security_group_ingress_rule.lambda_sg_ingress_access <security_group_rule_id>
```

### DIFF
--- a/ops/services/10-core/main.tf
+++ b/ops/services/10-core/main.tf
@@ -34,7 +34,7 @@ locals {
   region_name        = module.platform.primary_region.name
   vpc_id             = module.platform.vpc_id
   splunk_alert_email = lookup(module.platform.ssm.splunk, "alert-email", { value : null }).value
-  slack_queue_env    = local.parent_env
+  slack_queue_env    = local.parent_env == "test" || local.parent_env == "dev" ? "test" : "prod"
 }
 
 resource "aws_s3_bucket" "main_bucket" {
--- a/ops/services/10-core/main.tf
+++ b/ops/services/10-core/main.tf
@@ -34,7 +34,7 @@ locals {
   region_name        = module.platform.primary_region.name
   vpc_id             = module.platform.vpc_id
   splunk_alert_email = lookup(module.platform.ssm.splunk, "alert-email", { value : null }).value
-  slack_queue_env    = local.env == "test" || local.env == "dev" ? "test" : "prod"
+  slack_queue_env    = local.parent_env
 }
 
 resource "aws_s3_bucket" "main_bucket" {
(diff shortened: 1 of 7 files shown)


bennavapbc (dismissed):

ops/services/10-core/main.tf
```diff
@@ -34,7 +34,7 @@ locals {
   region_name        = module.platform.primary_region.name
   vpc_id             = module.platform.vpc_id
   splunk_alert_email = lookup(module.platform.ssm.splunk, "alert-email", { value : null }).value
-  slack_queue_env    = local.env == "test" || local.env == "dev" ? "test" : "prod"
+  slack_queue_env    = local.parent_env
```
mianava: is parent_env for "dev" test and parent_env for sandbox "prod"?

ops/services/30-lambda/main.tf
```diff
@@ -84,6 +84,22 @@ provider "artifactory" {
...
+  vpc_id = module.platform.vpc_id
+
+  tags = { Name = "${local.service_prefix}-microservices-lb" }
+}
+
+resource "aws_vpc_security_group_ingress_rule" "lambda_sg_ingress_access" {
+  from_port                    = 80
+  to_port                      = 80
+  ip_protocol                  = "tcp"
+  description                  = "inbound access for lambda to microservices"
```
mianava, line 98: Is this the other direction? Meaning the lambda gets called by microservices?

ops/services/10-core/main.tf
gfreeman-navapbc: Good catch, reverted the change and updated `local.env` to `local.parent_env` to account for ephemeral envs

juliareynolds-nava (approved):

ops/services/30-lambda/main.tf
gfreeman-navapbc: Could be, this was copied from a very old commit because this code is reliant on an orphaned resource from before the split to contracts and events

bennavapbc (approved):

PR #1757, opened 2026-04-21
AB2D-7158-Patch codeql 

keeyanghoreshi-rgb:
## 🎫 Ticket

Based on https://jira.cms.gov/browse/AB2D-7158

## 🛠 Changes

Remove build from codeql, switch runner to github runner. This prevents Artifactory access, but we don't need it anyway, since codeql doesn't build anyway. The package creation step is unnecessary for codeql.

If we still want to run the package step just to test it, we can do that seperately. If coverage of generated code is needed, we'll need to figure out something else.

## ℹ️ Context
 
Codeql has no ARM support. It has to run on a codebuild with the correct architecture support. 

## 🧪 Validation

https://github.com/CMSgov/ab2d/actions/runs/24745178298/job/72394853501

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,7 +2,7 @@ name: CodeQL
 
 on:
   push:
-    branches: [ "main", "codeql-fix" ]
+    branches: [ "main" ]
     
 jobs:
   analyze:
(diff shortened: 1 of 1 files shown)


bennavapbc (approved):

bennavapbc:
Will want @Sadibhatla to review this as she's more familiar with this workflow

bennavapbc (approved):

Sadibhatla (approved):

PR #1763, opened 2026-04-22
AB2D-7242: Rotate password securely 

mianava:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7242

## 🛠 Changes

Modifies the credentials for the IDR task definition so secrets are retrieved directly from SSM parameters by the task. 
## ℹ️ Context

These changes ensure secure coverage of secrets and credentials. 
 

## 🧪 Validation
These changes were validated by running them in the dev environment. We subsequently prioritized the provisioning of these changes into prod and ephemeral to ensure continuity of service.


keeyanghoreshi-rgb (dismissed):

ops/services/30-idr-db-importer/main.tf
```diff
@@ -34,15 +34,33 @@ resource "aws_ecs_task_definition" "idr_db_importer" {
...
+    cpu_architecture        = "ARM64"
+    operating_system_family = "LINUX"
   }
   container_definitions = nonsensitive(jsonencode([
     {
       image                  = "${local.image_repo_uri}:${var.image_tag}"
       name                   = local.service
       readonlyRootFilesystem = true
 
+      secrets = concat([
```
bennavapbc, line 46: Can we define the secrets outside of the call to `nonsensitive`? My understanding is SSM parameters are sensitive by default, but the `nonsensitive` negates that. 

I know this works using `data.aws_ssm_parameter.ab2d_db_password.arn`, but if someone mistakenly updated this to `data.aws_ssm_parameter.ab2d_db_password.value` it would print the password due to the `nonsensitive` keyword. If this were defined outside of `nonsensitive`, the SSM parameter would be treated as sensitive by default.

Sadibhatla (approved):

smirnovaae (approved):

ops/services/30-idr-db-importer/main.tf
bennavapbc: To be addressed in https://jira.cms.gov/browse/AB2D-7247, marking resolved

bennavapbc (approved):

PR #1766, opened 2026-04-27
AB2D-7248: Don't request PDE from BFD in v1/v2

alex-dzeda:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7248

## 🛠 Changes

Adds "type" parameter to calls to BFD to ensure that only Part A + B claims are requested

## ℹ️ Context

To improve performance, optimizations to AB2D are underway from both the AB2D side + BFD. AB2D is not allowed to share part D claims, so there is no point to request them from BFD, with all of the overhead that is entailed. This will reduce the volume of claims received by AB2D from BFD by approx ~50%, which should improve AB2D's performance in v1/v2. Additionally, this change improves parity with v3, where _source=NCH is functionally equivalent to the addition of "type" to this query.


Sadibhatla (approved):

bennavapbc (approved):

libs/ab2d-bfd/src/main/java/gov/cms/ab2d/bfd/client/BFDSearchImpl.java
```diff
@@ -67,6 +67,7 @@ public IBaseBundle searchEOB(BFDSearchDTO searchDTO) throws IOException {
         }
         else {
             url.append("&excludeSAMHSA=true");
+            url.append("&type=carrier,dme,hha,hospice,inpatient,outpatient,snf");
```
smirnovaae, line 70: Could you please update version `bfdVersion='4.5.2'` in build.gradle?

PR #1791, opened 2026-06-02
AB2D-7251  Add all active contracts to coverage table

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7251

## 🛠 Changes

  - Purge inactive contracts from the V3 history summary 
  - Track contract end dates from HPMS 
  - Skip non-attested contracts in staging copy 

## ℹ️ Context

Currently, the historic coverage database table only contains attribution data for contracts that have submitted an API request in the past year. Coverage for all active contracts per HPMS needs to be available

## 🧪 Validation

Deployed and tested in ephemeral prod 

```
"message": "[V3] Deleted 3692244 rows from coverage_v3_history_summary for unattested contracts or contracts ended > 2 years ago",
    "logger_name": "gov.cms.ab2d.coverage.service.v3.CoverageV3SyncServiceImpl"

```


coverage/src/main/java/gov/cms/ab2d/coverage/service/v3/CoverageV3SyncServiceImpl.java
```diff
@@ -131,12 +133,27 @@ WHERE make_date(year, month, 1) < (date_trunc('month', CURRENT_DATE) - interval
...
+    with deleted_rows as (
+        DELETE FROM v3.coverage_v3_history_summary
+        WHERE contract IN (
+            SELECT contract_number FROM contract.contract
+            WHERE attested_on IS NULL
+               OR (hpms_end_date IS NOT NULL AND hpms_end_date < :cutoff)
+        )
+        RETURNING *
+    )
+    select count(*) from deleted_rows;
```
bennavapbc: The PR looks good as is, but it would be useful for auditing/debugging purposes if we could log the inactive contracts that were deleted. 

I don't know what the chances are, but if HPMS marks a contract as unattested (which we've seen before) or otherwise updates the contract information that leads to us deleting coverage data, it would be helpful to have a record of that in the logs. I'd just suggest adding a method `getInactiveContracts()` and logging the output, and then proceeding to call `deleteInactiveContractsFromHistorySummary()`

bennavapbc (dismissed):

contracts/build.gradle
```diff
@@ -18,7 +18,7 @@ ext {
     testContainerVersion='1.21.4'
     liquibaseVersion = '5.0.0'
     eventClientVersion='3.3.14'
-    contractClientVersion='2.1.4'
+    contractClientVersion='2.2.0'
```
Sadibhatla, line 21: @keeyanghoreshi-rgb : tagging here specifically for commonsLangVersion

Sadibhatla (approved):

bennavapbc (approved):

PR #1809, opened 2026-06-22
Ab2d-7222/eob processing allocations

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7222

## 🛠 Changes

- The PR still includes metric collection to compare GC behavior and memory allocation between the legacy approach and the in-place implementation.

- Created a new getBenefitInPlace(...) method that cleans up the EOB directly, without creating a .copy() for each field.
- Top-level EOB fields are controlled by an allowlist (KEPT_ELEMENTS)
## ℹ️ Context

With the [investigation for the in-place trimmer](https://confluence.cms.gov/pages/viewpage.action?pageId=1564311544&spaceKey=ADN&title=Refactor+ExplanationOfBenefit+Processing+to+Eliminate+Excess+Allocations), the logic that was presented had a key difference from the existing trimmer: it changed from an allowlist of elements to add to a denylist of elements to remove.

## 🧪 Validation

Deployed to test, ran both the legacy and in-place jobs, and compared the results.


worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
```diff
@@ -95,6 +113,11 @@ public Future<ProgressTrackerUpdate> process(PatientClaimsRequest request) {
     }
 
     String writeOutData(PatientClaimsRequest request, FhirVersion fhirVersion, ProgressTrackerUpdate update) throws IOException {
+        boolean useInPlace   = propertiesService.isToggleOn(EOB_V3_IN_PLACE, false);
```
bennavapbc: @smirnovaae What do you think about adding an E2E test to verify the regular trimmer and in-place trimmer produce the same output?

libs/ab2d-filters/src/main/java/gov/cms/ab2d/filter/ExplanationOfBenefitTrimmerR4V3.java
```diff
@@ -158,6 +193,80 @@ public static IBaseResource getBenefit(IBaseResource resource) {
...
+
+        List<ExplanationOfBenefit.SupportingInformationComponent> filteredSupportingInfo =
+                filterSupportingInfo(benefit.getSupportingInfo());
+
+        List<ExplanationOfBenefit.InsuranceComponent> filteredInsurance = buildInsuranceInPlace(benefit);
+
+        benefit.setContained(filteredContained);
+        benefit.setSupportingInfo(filteredSupportingInfo);
+        benefit.setCareTeam(filteredCareTeam);
+        benefit.setInsurance(filteredInsurance);
```
colby-seyferth-nava: For these, why are we creating a new object, adding the sub-elements to the object, then setting the element to the new object?

This appears to be building up new objects instead of deleting in-place, but i might just be missing something

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
smirnovaae: done!

libs/ab2d-filters/src/main/java/gov/cms/ab2d/filter/ExplanationOfBenefitTrimmerR4V3.java
smirnovaae: Updated!

worker/src/main/java/gov/cms/ab2d/worker/processor/PatientClaimsProcessorImpl.java
bennavapbc: Looks great! Thanks for adding that

bennavapbc (dismissed):

colby-seyferth-nava (dismissed):

colby-seyferth-nava (dismissed):

bennavapbc (approved):

bennavapbc (approved):

colby-seyferth-nava (approved):

PR #1816, opened 2026-07-01
AB2D-7319 pause resume prototype

keeyanghoreshi-rgb:
## 🎫 Ticket

[https://jira.cms.gov/browse/AB2D-7319](https://jira.cms.gov/browse/AB2D-7319)

## 🛠 Changes

Adds the prototype code for the pause/resume feature. Creates a separate version of the processor, and when the feature flag is on, hijacks incoming jobs and lets the prototype processor partition it and create BATCH metadata for the job.

Essentially, Spring Batch creates a ledger of work to be done, and also saves its work with checkpoints every so often. On a worker shutting down, the job is paused, and when the worker picks it up again, it will restart from where it left off.

## ℹ️ Context

These changes prevent having to redo an entire job if the worker needs to be redeployed or taken down for any reason. 

## 🧪 Validation

Tested locally - more validation to come in following tickets.


coverage/src/main/java/gov/cms/ab2d/coverage/query/GetAggregatedCoverageMembership.java
```diff
@@ -129,6 +129,23 @@ SELECT DISTINCT json_array_elements(to_json(historical_coverage_summaries))::TEX
     FETCH FIRST :limit ROWS WITH TIES
     """;
 
+    private static final String FETCH_FROM_AGGREGATED_TABLE_BY_ROW_RANGE =
```
bennavapbc: Because this table can contain multiple records with the same `patient_id`, this query could return an incomplete set of records for a given beneficiary. This is briefly explained in [Handling of opt-outs and multiple records](https://confluence.cms.gov/spaces/ADN/pages/1636797221/Support+Troubleshooting+Runbook+v1+v2+v3#Support%26TroubleshootingRunbook(v1%2Fv2%2Fv3)-Handlingofopt-outsandmultiplerecords). In order to properly handle duplicates and opt-outs, the query would need to return all records for a given beneficiary.

Sadibhatla: To note: If bene_sk based data sharing preferences gets finished before this.. There will be changes to this porcess as optout functionality will be IDR based and will be stored in diff tables for v3

keeyanghoreshi-rgb:
This isn't being merged anytime soon, so I'm closing it for now

PR #1821, opened 2026-07-06
AB2D-7287 Create automation for Datadog Monitors and Dashboards

gfreeman-navapbc:
## 🎫 Ticket

AB2D-7287

## 🛠 Changes

Adds workflows to handle plan and apply for `60-monitors` and `65-dashboard`.

## ℹ️ Context

We have the terraform to handle the infrastructure, now we need automation to apply the terraform at a regular cadence.

## 🧪 Validation

Workflows need to be merged to be tested thoroughly, but this is based on the existing plan and apply workflow, just specific for these modules.


.github/workflows/dashboard.yml
```diff
@@ -0,0 +1,50 @@
...
+
+concurrency:
+  group: tofu-plan-or-apply
+
+permissions:
+  contents: read
+  id-token: write
+
+env:
+  TENV_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
keeyanghoreshi-rgb: My assumption is that this is included here for the call to `tofu-run`? But env variables in github actions don't propagate down to called workflows. This should be moved to `tofu-run`

.github/workflows/dashboard.yml
```diff
@@ -0,0 +1,50 @@
...
+
+jobs:
+  setup:
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: codebuild-ab2d-prod-${{ github.run_id }}-${{ github.run_attempt }}
```
keeyanghoreshi-rgb: this could be non-prod. Not that it matters much, but `tofu-run` will choose its own runner anyway and `monitors.yml` uses non-prod, so it would be more consistent.

.github/workflows/monitors.yml
```diff
@@ -0,0 +1,77 @@
...
+
+concurrency:
+  group: tofu-plan-or-apply
+
+permissions:
+  contents: read
+  id-token: write
+
+env:
+  TENV_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
keeyanghoreshi-rgb: same as above

.github/workflows/dashboard.yml
gfreeman-navapbc: Yea I think my reasoning here was that dashboard will only ever be run in prod, but no harm in changing the caller workflow to non-prod

.github/workflows/dashboard.yml
gfreeman-navapbc: Good catch

.github/workflows/dashboard.yml
```diff
@@ -0,0 +1,51 @@
+name: 65-dashboard
+run-name: "[${{ github.event_name == 'push' && 'APPLY' || 'PLAN' }}] 65-dashboard → ${{ github.event_name == 'push' && github.ref_name || github.head_ref }}"
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "ops/services/65-dashboard/**"
+      - ".github/workflows/tofu-dashboard.yml"
```
keeyanghoreshi-rgb: Is this the right file name?

```suggestion
      - ".github/workflows/dashboard.yml"
```

.github/workflows/monitors.yml
```diff
@@ -0,0 +1,78 @@
+name: 60-monitors
+run-name: "[${{ github.event_name == 'push' && 'APPLY' || 'PLAN' }}] 60-monitors → ${{ github.event_name == 'push' && github.ref_name || github.head_ref }}"
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "ops/services/60-monitors/**"
+      - ".github/workflows/tofu-monitors.yml"
```
keeyanghoreshi-rgb:
```suggestion
      - ".github/workflows/monitors.yml"
```

.github/workflows/dashboard.yml
gfreeman-navapbc: whoops, copied `tofu-run`'s path, good catch

keeyanghoreshi-rgb (approved):

smirnovaae (approved):

PR #1822, opened 2026-07-07
AB2D-7247 Refactor IDR Importer Module

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7247

## 🛠 Changes

- Moved ab2d_db_user, idr_snowflake_user, and idr_snowflake_role from the environment block to the ECS secrets block, referencing them via valueFrom = <arn> instead of value = <...>.value.
- Removed the nonsensitive(...) wrapper from container_definitions, since no sensitive .value remains inline — restoring Tofu's sensitive-value masking.
- Non-sensitive params (ab2d_db_database, ab2d_db_host, idr_db_importer_bucket) remain as plain environment values.

## ℹ️ Context

We want to refactor the IDR importer module to define secrets outside of the nonsensitive block so that sensitive values like database passwords are not inadvertently exposed during a tofu apply.

## 🧪 Validation

Task completed successfully 

Screenshot 2026-07-07 at 2 20 28 PM


mianava (changes requested):
This importer can be moved to use the service module.

ops/services/10-core/iam.tf
```diff
@@ -389,7 +363,11 @@ data "aws_iam_policy_document" "idr_db_importer_additional_bucket_policy" {
       variable = "aws:PrincipalArn"
       values = [
         aws_iam_role.idr_db_importer.arn,
-        aws_iam_role.idr_db_importer_task.arn
+        # The importer ECS task role is created by the `service` module in the
+        # 30-idr-db-importer terraservice. It has no IAM path, so its ARN is
+        # constructed here as a string to avoid a reverse cross-state dependency
+        # (10-core applies before 30-idr-db-importer).
+        "arn:aws:iam::${local.aws_account_number}:role/${local.service_prefix}-idr-db-importer-task-role"
```
mianava: Is this for access to an S3 bucket?

ops/services/30-idr-db-importer/main.tf
```diff
@@ -25,104 +25,49 @@ module "platform" {
...
+  memory               = 2048
+  desired_count        = 0
+  enable_datadog_agent = false
+  execution_role_arn   = data.aws_iam_role.idr_db_importer_task_execution.arn
+  image                = "${local.image_repo_uri}:${var.image_tag}"
+  platform             = module.platform
+  security_groups      = [data.aws_security_group.idr_db_importer.id]
+  subnets              = keys(module.platform.private_subnets)
+
+  additional_task_role_policies = { s3 = data.aws_iam_policy.idr_db_importer_task.arn }
```
mianava: where is this policy generated?

smirnovaae: Good catch. It was looking up the policy from 10-core by name, which was more complicated than needed. I moved the policy into 30-idr-db-importer as aws_iam_policy.idr_db_importer_task and now reference it directly.

ops/services/10-core/iam.tf
smirnovaae: yes

mianava: Can this bucket be moved into its service idr-db-importer? I assume it's required for the IDR importer and is not "core"

ops/services/30-idr-db-importer/main.tf
```diff
@@ -25,104 +25,84 @@ module "platform" {
...
+  container_secrets = concat(
+    [
+      { name = "AB2D_DB_PASSWORD", valueFrom = data.aws_ssm_parameter.ab2d_db_password.arn },
+      { name = "AB2D_DB_USER", valueFrom = data.aws_ssm_parameter.ab2d_db_user.arn }
+    ],
+    module.platform.parent_env == "prod" ? [
+      { name = "IDR_SNOWFLAKE_PRIVATE_KEY", valueFrom = data.aws_ssm_parameter.idr_private_key[0].arn },
+      { name = "IDR_SNOWFLAKE_WAREHOUSE", valueFrom = data.aws_ssm_parameter.idr_snowflake_warehouse[0].arn },
+      { name = "IDR_SNOWFLAKE_USER", valueFrom = data.aws_ssm_parameter.idr_snowflake_user[0].arn },
+      { name = "IDR_SNOWFLAKE_ROLE", valueFrom = data.aws_ssm_parameter.idr_snowflake_role[0].arn }
```
mianava: Are these stored in SOPS? If so, use platform module with ssm_root_map 

  ssm_root_map = { this = "/ab2d/${local.env}/id-db-importer/" }
Ideally, these get stored in a pattern that matches their relevant service (hence "idr-db-importer" is the path because that's how this terraservice has been named) 

And then you can use `module.platform.ssm.this.snowflake_role.arn`

ops/services/30-idr-db-importer/main.tf
```diff
@@ -25,104 +25,84 @@ module "platform" {
...
+        Effect = "Allow"
+        Action = [
+          "s3:AbortMultipartUpload",
+          "s3:GetObject",
+          "s3:PutObject",
+          "s3:DeleteObject",
+          "s3:ListBucket"
+        ]
+        Resource = [
+          "arn:aws:s3:::${data.aws_ssm_parameter.idr_db_importer_bucket.value}",
```
mianava: Recommend getting the arn directly https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/s3_bucket -- though I'd favor moving the s3 bucket to be colocated with idr-db-importer as it is critical to this service

ops/services/10-core/iam.tf
```diff
@@ -347,8 +285,8 @@ resource "aws_iam_policy" "idr_db_importer" {
           "s3:ListBucket"
         ]
         Resource = [
-          "${module.idr_db_importer_bucket.arn}",
-          "${module.idr_db_importer_bucket.arn}/*"
+          "arn:aws:s3:::${module.platform.app}-${module.platform.env}-idr-db-importer-*",
+          "arn:aws:s3:::${module.platform.app}-${module.platform.env}-idr-db-importer-*/*"
```
mianava: Can this logic and the bucket be moved to live alongside the service that it's associated with?

ops/services/30-idr-db-importer/iam.tf
```diff
@@ -0,0 +1,63 @@
...
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          Service = [
+            "rds.amazonaws.com"
+          ]
+        }
+        Condition = {
+          "StringLike" = {
+            "aws:SourceAccount" = "${local.aws_account_number}"
```
mianava: I am not sure if this introduces further risk regarding printing additional information than what is already printed in a git workflow run.

ops/services/10-core/database.tf
```diff
@@ -1,5 +1,7 @@
+# FIXME Move conditional logic into a config/<env>.yml or workspace vars file and process with a locals block
+# TODO change the commit hash once these aurora changes are merged in CDAP
 module "db" {
-  source = "github.com/CMSgov/cdap//terraform/modules/aurora?ref=b0e43ec5b235d8fddc9cf2aaf3440b73e185d47f"
+  source = "github.com/CMSgov/cdap//terraform/modules/aurora?ref=1b4ae3a625562fd9317f047f34bc2047ec949bed"
```
mianava: This needs to be changed upon the merge of CDAP PR.

ops/services/30-idr-db-importer/iam.tf
```diff
@@ -0,0 +1,157 @@
...
+}
+
+data "aws_iam_policy_document" "idr_db_importer_bucket_extended_deny_policy" {
+  statement {
+    sid    = "DenyReadAccessExceptIDRDBImporterRole"
+    effect = "Deny"
+
+    actions = [
+      "s3:GetObject"
+    ]
```
mianava, line 122: Should this be restricted beyond Get? Would it serve to also prevent PUT?

ops/services/10-core/iam.tf
```diff
@@ -186,225 +186,6 @@ resource "aws_iam_policy" "chatbot_guardrail_policy" {
...
-  policy = jsonencode({
-    Version = "2012-10-17"
-    Statement = [
-      {
-        Sid    = "S3Import"
-        Effect = "Allow"
-        Action = [
-          "s3:GetObject",
-          "s3:ListBucket"
-        ]
```
mianava, line 348: I think this was all replaced in iam.tf in the terraservice. Though there is not an exact policy that looks like this one, it has more API calls.

ops/services/30-idr-db-importer/iam.tf
smirnovaae: done

ops/services/30-idr-db-importer/iam.tf
smirnovaae: done

mianava (approved):
LGTM - Hoping that having this more sorted makes iterating and updating easier.

bennavapbc (approved):

PR #1838, opened 2026-08-12
PLT-1743 : Create code-checks-snyk-scheduled.yml

jasonvinson:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1743

## 🛠 Changes

Adds a scheduled Snyk Code run every other day in addition to our Snyk runs on PRs.

## ℹ️ Context

Suggested by Brian Hodges to further increase code security.

## 🧪 Validation

Verified by running workflow in CDAP repo.

### DIFF
--- a/.github/workflows/code-checks-snyk-scheduled.yml
+++ b/.github/workflows/code-checks-snyk-scheduled.yml
@@ -1,0 +1,19 @@
+name: code-checks (snyk code scheduled)
+
+on:
+  schedule:
+    - cron: '0 0 */2 * *'
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  scheduled-code-checks:
+    name: 'Scheduled Code Checks for AB2D'
+    uses: CMSGov/CDAP/.github/workflows/code-checks-snyk.yml@028524afd97e26370e0dd812faad64b9a2ad558b
+    with:
+      app: ab2d
+      ref: main
+    secrets: inherit


keeyanghoreshi-rgb (approved):

.github/workflows/code-checks-snyk-scheduled.yml
```diff
@@ -0,0 +1,19 @@
...
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  scheduled-code-checks:
+    name: 'Scheduled Code Checks for AB2D'
+    uses: CMSGov/CDAP/.github/workflows/code-checks-snyk.yml@028524afd97e26370e0dd812faad64b9a2ad558b
```
bennavapbc, line 15: The command used in this [workflow](https://github.com/CMSgov/cdap/blob/main/.github/workflows/code-checks-snyk.yml#L69) might need to be updated to include **--all-projects**. Our repo is a monorepo with a mix of gradle and maven projects.

jasonvinson: Hi Ben! This workflow is running snyk code static application security testing (SAST). Snyk code scans all the code in the entire repo. The --all-projects flag I believe is for Snyk Open Source dependency scanning, which we also do separately.

bennavapbc: ~@jasonvinson We've experimented with this before, and I believe we must use `--all-projects` because we have multiple, separate projects in our repo. Without `--all-projects`, I recall that only pom.xml is inspected.~

~Tagging @keeyanghoreshi-rgb as he's also used this CLI before. Either way, I think we will want to modify the CDAP workflow to allow additional arguments so we can provide a list of projects to scan.~

@jasonvinson Re-reading your comment and experimenting with the CLI, I see what you mean. Thanks for clarifying. This PR looks good and I will be adding a `.snyk` file to ignore some projects from snyk SAST scans. Thanks!

bennavapbc (approved):

PR #1839, opened 2026-08-17
AB2D-7391 Skip v3 cron jobs during BFD coverage sync 

bennavapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7391

## 🛠 Changes

Avoid running v3-related scheduled tasks in when the v1/v2 coverage sync is in progress. 

## ℹ️ Context

Too many database-intensive operations can cause query timeouts in both v1/v2 and v3 tasks as discussed [here.](https://cmsgov.slack.com/archives/G01M1A45ASX/p1786477855597279) leading to a series of retries, more timeouts, and longer time to complete.


coverage/src/main/java/gov/cms/ab2d/coverage/service/v3/CoverageV3SyncServiceImpl.java
```diff
@@ -191,6 +191,11 @@ ON CONFLICT (contract, year, month)
     DO NOTHING
     """;
 
+    private static final String QUERY_BFD_COVERAGE_SYNC_IN_PROGRESS =
+    """
```
Sadibhatla, line 195: Add current year to this ? Sometimes we see old coverages getting scheduled . Wonky behavior I have noticed once in awhile when worker crashes . Nitpicking ofcourse

Sadibhatla (approved):

coverage/src/main/java/gov/cms/ab2d/coverage/service/v3/CoverageV3SyncServiceImpl.java
bennavapbc: Hmm... even if old coverage periods are running, I still think we want to avoid running the BFD sync and v3 cron sync at the same time? Before a v3 job is started, the worker will sync v3 coverage data for that contract even if the BFD sync is in progress. Thoughts?

Sadibhatla: Ya old or new the load on the worker will remain the same . In ideal case it shud never try to run coverage sync for months older than 3 . Which is why suggesting a year check .

keeyanghoreshi-rgb (approved):

PR #1855, opened 2026-08-28
AB2D-7379 AB2D-7425 Add pause/resume liveness check 

bennavapbc:
## 🎫 Ticket

- https://jira.cms.gov/browse/AB2D-7379
- https://jira.cms.gov/browse/AB2D-7425

## 🛠 Changes

Add liveness check for `PrototypeJobLeaseRenewer`

## ℹ️ Context

`PrototypeJobLeaseRenewer` renews a lease periodically (every 20 seconds by default). It's possible for a worker to be dead and/or not making any progress in which case we don't want to continue renewing the lease. 

An in-memory heartbeat is used to check for liveness, and if the worker appears to be alive, its lease will continue to be renewed by `PrototypeJobLeaseRenewer`. If too much time elapses without a heartbeat (whether it be from a SpringBatch listener or called manually before creating the aggregated table), its lease will not be renewed. 

## 🧪 Validation

Added integration test with two test cases


worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/PrototypeJobProcessorImpl.java
```diff
@@ -186,13 +189,15 @@ public Job process(String jobUuid) {
             JobExecution last = batchJobRepository.getLastJobExecution(PROTOTYPE_JOB_NAME, parameters);
             if (last == null) {
                 log.info("no prior batch execution for job {} - creating aggregated attribution table", jobUuid);
+                leaseRenewer.postHeartbeat(jobUuid, fenceToken, BEFORE_CREATE_AGGREGATED_TABLE);
```
bennavapbc: Creating the aggregated table can take several minutes (for large production clients). The worker will block until it is completed and will appear to not be making progress. This posts a heartbeat and expects the next heartbeat to be within a given window (10 minutes as configured currently).

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/PrototypeJobProcessorImpl.java
```diff
@@ -239,6 +244,7 @@ public Job process(String jobUuid) {
                     // The batch finished but too many beneficiaries errored, this is a terminal failure
                     applyThresholdFailure(job, contractNumber, jobUuid);
                 } else {
+                    leaseRenewer.postHeartbeat(jobUuid, fenceToken, BEFORE_ASSEMBLE_FILES);
```
bennavapbc: It's unclear whether we need to care about this scenario. If assembling files **does** take a while, then it makes sense to give a larger window before expecting the next heartbeat

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/PrototypeProperties.java
```diff
@@ -52,4 +54,9 @@ public class PrototypeProperties {
      * Whether hard recovery restarts partitions from scratch or from the last good chunk
      */
     private boolean copyForwardEnabled = true;
+
+    private long maxDurationSecondsAfterCreateLease = Duration.ofMinutes(1).toSeconds();
+    private long maxDurationSecondsCreateAggregatedTable = Duration.ofMinutes(10).toSeconds();
```
keeyanghoreshi-rgb: Minor note, the name `maxDurationSecondsCreateAggregatedTable` doesn't match the `max-duration-seconds-before-create-aggregated-table` in the properties. `maxDurationSecondsAssembleFiles` is the same.

worker/src/main/resources/application.properties
```diff
@@ -45,6 +45,12 @@ pause-resume.prototype.copy-forward-enabled=true
 job.lock.ttl=40
 # how often a worker renews its job_lease heartbeat, in milliseconds (must be well under job.lock.ttl)
 pause-resume.prototype.lease-renew-ms=20000
+
+pause-resume.prototype.max-duration-seconds-after-create-lease=60
+pause-resume.prototype.max-duration-seconds-before-create-aggregated-table=600
+pause-resume.prototype.max-duration-seconds-after-write-callback=120
+pause-resume.prototype.max-duration-seconds-before-assemble-files=600
```
keeyanghoreshi-rgb: We can be pretty generous with the time windows here, the case we're guarding against should be pretty rare. 

For the after-write, I think it should be dynamic based on chunk size. We can calculate the maximum time we should wait before dropping the lease by doing something like chunkSize * bfdTimeout. In the absolute worst case, every item in a chunk finishes just before it times out. Anything over that would be an indication that the worker isn't making progress.

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/PrototypeJobProcessorImpl.java
```diff
@@ -479,6 +485,12 @@ private org.springframework.batch.core.job.Job buildPartitionedJob(String contra
                 // each chunk commit comes with a fence token which prevents
                 // us from committing work if we've lost the job
                 .listener(new PrototypeFenceGuard(jobLease, jobUuid, fenceToken))
+                .listener(new ItemWriteListener<>() {
+                    @Override
+                    public void afterWrite(Chunk<?> items) {
+                        leaseRenewer.postHeartbeat(jobUuid, fenceToken, AFTER_WRITE_CALLBACK);
+                    }
+                })
```
keeyanghoreshi-rgb: This is a minor concern, but the very first chunk will have a max heartbeat of either the `create_lease` or `table_creation` steps instead of the `after_write`. One solution is to just extend the max duration of all the steps. Another is to do this `postHeartbeat` in `beforeChunk` instead of `afterWrite` which should accomplish the same effect but cover the first chunk.

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/PrototypeJobProcessorImpl.java
keeyanghoreshi-rgb: I think extending the window before assembly is a good idea.

worker/src/main/resources/application.properties
bennavapbc: Updated this to be dynamic based on a configured value (max duration to process an item on average) and chunk size.

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/PrototypeJobProcessorImpl.java
bennavapbc: Good point. Added an event `BEFORE_LAUNCH_OR_RESUME_JOB` to handle this

worker/src/test/java/gov/cms/ab2d/worker/processor/prototype/PrototypeJobLeaseExpirationIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+		    });
+    }
+
+    @Test
+    @DisplayName("Job lease is removed due to being fenced out")
+    void jobLeaseRemovedDueToBeingFencedOut(CapturedOutput out) throws Exception {
+        val job = createSubmittedV3Job("test");
+		val jobUuid = job.getJobUuid();
+        val owner = startWorkerUntilOnePartitionDone(jobUuid, "test-owner");
+        owner.run();
```
keeyanghoreshi-rgb: The tests should call `owner.killHard()` when they're done with the worker, otherwise the worker thread stays running into the next test which might cause flakiness.

bennavapbc: Done

keeyanghoreshi-rgb (approved):

PR #1858, opened 2026-09-03
AB2D-7395 Use batched inserts for daily IDR <-> worker sync

bennavapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7395

## 🛠 Changes

Use batched inserts for daily IDR <-> worker sync

## ℹ️ Context

Our largest contract has 8.8 million beneficiaries for each year-month period in the staging table. 

For syncing attribution data from staging to the recent coverage table, inserts are now done for each year-month period at a time (8.8 million) instead of all (26.7 million).

image

## 🧪 Validation

Existing tests pass and added a new test to simulate a sync failure.

### DIFF
--- a/coverage/src/main/java/gov/cms/ab2d/coverage/service/v3/CoverageV3SyncServiceImpl.java
+++ b/coverage/src/main/java/gov/cms/ab2d/coverage/service/v3/CoverageV3SyncServiceImpl.java
@@ -296,10 +296,12 @@ public class CoverageV3SyncServiceImpl  implements CoverageV3SyncService {
             audit.log(action, null, contract, null, Map.of("rowsInserted", rowsInserted));
         } catch (Exception e) {
             /**
-             * if {@link #batchCopyFromStagingToCoverage} throws an exception, it will record an audit event
-             * rethrow exception to rollback transaction
+             * If {@link #batchCopyFromStagingToCoverage} throws an exception, it will record an audit event
+             * If this step fails, there may be incomplete attribution data copied from staging to the recent coverage
+             * table. The sync will be re-attempted the next time the cron job fires OR when a job is run, so this
+             * will correct itself eventually
              */
-           throw e;
+            return SYNC_FAILED_FOR_CONTRACT;
         }
 
         val rowsInCoverageAfterCopy = executeTimedQuery(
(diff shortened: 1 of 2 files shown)


Sadibhatla (dismissed):

coverage/src/main/java/gov/cms/ab2d/coverage/service/v3/CoverageV3SyncServiceImpl.java
```diff
@@ -284,12 +287,20 @@ private CoverageV3SyncResult copyFromStagingToRecentCoverageTable(final String c
...
+            val rowsInserted = executeTimedQuery(
+                format("[V3] batchCopyFromStagingToCoverage contract=%s", contract),
+                () -> batchCopyFromStagingToCoverage(contract)
+            );
+            log.info("[V3] Copied {} rows from staging to coverage for contract {}", rowsInserted, contract);
+            audit.log(action, null, contract, null, Map.of("rowsInserted", rowsInserted));
+        } catch (Exception e) {
+            /**
+             * if {@link #batchCopyFromStagingToCoverage} throws an exception, it will record an audit event
+             * rethrow exception to rollback transaction
```
keeyanghoreshi-rgb: Which transaction is this trying to roll back? This will toss the audit events as well. If the intention is to keep the audit logs on exception, I think it won't work as it is now.

bennavapbc: That's a good point. We would need to update the audit log method with `@Transactional(propagation = Propagation.REQUIRES_NEW)` in order for the audit log writes to succeed unconditionally. Eventually we should probably refactor this to batch audit logs and then commit in a separate transaction in a `finally` block. 

I removed the line that rethrows the exception, so the transaction is not rolled back. As a result there might exist partially synced attribution data if there's an error, but that isn't so bad because it will self correct (added a comment regarding this).

coverage/src/main/java/gov/cms/ab2d/coverage/service/v3/CoverageV3SyncServiceImpl.java
```diff
@@ -301,11 +314,11 @@ private CoverageV3SyncResult copyFromStagingToRecentCoverageTable(final String c
         if (rowsInStaging != rowsInCoverageAfterCopy) {
             log.error("[V3] Row count in staging ({}) != row count in coverage ({}) for contract {}",
                     rowsInStaging,
-                    rowsInCoverageBeforeCopy,
+                    rowsInCoverageAfterCopy,
                     contract
             );
             result = SYNC_FAILED_FOR_CONTRACT;
-            audit.log(action, result, contract, "rowsInStaging does not match rowsInCoverageBeforeCopy", Map.of("rowsInStaging", rowsInStaging, "rowsInCoverageBeforeCopy", rowsInCoverageBeforeCopy));
+            audit.log(action, result, contract, "rowsInStaging does not match rowsInCoverageAfterCopy", Map.of("rowsInStaging", rowsInStaging, "rowsInCoverageAfterCopy", rowsInCoverageAfterCopy));
```
bennavapbc, line 321: Just noticed this bug and in the few lines above

smirnovaae (approved):

Sadibhatla (approved):

PR #1861, opened 2026-09-03
AB2D-7374 Crash/fault injection for the pause-resume prototype worker

Sadibhatla:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7374

## 🛠 Changes

Adds crash/fault injection to the pause-resume prototype worker so recovery pathways can be validated on a real deployed container, not just in-process tests.

- CrashInjector - a small, config-driven component that halts the worker (Runtime.halt(137)) at a chosen pipeline point. Replaces the single hardcoded crash hook that was inline in EobItemProcessor. Controlled by:
  - pause-resume.prototype.crash-probability (0 = off)
  - pause-resume.prototype.crash-at - a CrashPoint enum value: process, read, write, or assemble
- Wired into four points: EobItemProcessor (process), BeneficiaryItemReader (read), NdjsonCompositeWriter (between the data and error writes), and PrototypeOutputAssembler (mid promoteWinners, so a restart has to finish a half-promoted assembly idempotently).
- Safety: validates crash-at against the enum, and only arms in dev/test/local by reading execution.env. Prod, sandbox, and any unrecognised env are denied, so a bad value fails closed.
- Tests: CrashInjectorTest (off by default, fires only at the configured point, case-insensitive config, unknown value disarms, and a parameterized guard proving prod/sandbox/unknown never arm), plus a new crashDuringReadingRecovers case filling the one previously-uncovered stage.
- Runbook: docs/prototype-crash-testing.md - how to crash a deployed worker (env-var injector or AWS FIS / ecs stop-task), the log markers to watch, and what to verify after recovery.

Default behaviour is unchanged: crash-probability=0 means every injection point is a no-op unless explicitly armed in a dev/test worker.

## ℹ️ Context

We want confidence that a worker can safely restart a job in a deployed environment without data corruption or faulty output, including crashes mid-file-write and mid-assembly. Unit and integration tests already cover many recovery cases in-process; this adds a way to exercise recovery on a real container and follow the crash then recovery then clean-completion story in the logs (the processor already logs the fresh/soft/hard claim on pickup).

No new dependencies, security controls, or data flows. Crash injection is off by default and can only ever run in dev/test/local.

## 🧪 Validation

image
image

### DIFF
--- a/docs/prototype-crash-testing.md
+++ b/docs/prototype-crash-testing.md
@@ -1,0 +1,92 @@
+# Prototype crash testing runbook
+
+The pause/resume prototype worker has recovery paths for crashes at each stage of a job
+(partitioning, reading, processing, file writing, assembly). Unit and integration tests cover these
+in-process. This runbook is for the other half: proving a **deployed** worker container recovers from a
+real crash without corrupting or duplicating output.
+
+Crash injection is off by default and is only meant for `dev`/`test`. **Never arm it in prod.**
+
+## How the injector works
+
+`CrashInjector` halts the JVM (`Runtime.getRuntime().halt(137)`) at a chosen pipeline point, so the
+container dies exactly like a real crash - no shutdown hooks, no cleanup. It is controlled by two
+properties:
+
+| Property | Env var | Meaning |
+| --- | --- | --- |
+| `pause-resume.prototype.crash-probability` | `PAUSE_RESUME_PROTOTYPE_CRASH_PROBABILITY` | Chance (0-1) of crashing each time the armed point is hit. `0` = off. |
+| `pause-resume.prototype.crash-at` | `PAUSE_RESUME_PROTOTYPE_CRASH_AT` | Which point crashes: `process`, `read`, `write`, or `assemble`. |
+
+A small probability (e.g. `0.01`) crashes somewhere in the middle of a large job, which is the
+interesting case for recovery. `1.0` crashes at the first opportunity.
+
+## Option A: crash the worker from the inside (injector)
+
+Use this to test a crash at a specific pipeline stage.
+
+1. Point the worker task at a crash stage and probability. In `dev` you can set these as env vars on the
+   worker ECS task definition (they flow into Spring via relaxed binding - no code change needed):
+
+   ```
+   PAUSE_RESUME_PROTOTYPE_CRASH_AT=write
+   PAUSE_RESUME_PROTOTYPE_CRASH_PROBABILITY=0.02
+   ```
+
+2. Redeploy the worker so the new task definition is live.
+3. Submit a job large enough to span several chunks.
+4. Watch the task die and ECS start a replacement. The replacement picks the job back up (hard recovery:
+   new lease token, new files) and finishes it.
+5. **Reset `PAUSE_RESUME_PROTOTYPE_CRASH_PROBABILITY=0` and redeploy** when you are done.
+
+Repeat with `crash-at=read`, `process`, and `assemble` to exercise each stage.
+
+## Option B: crash the container from the outside (AWS FIS / ECS)
+
+Use this to test an infrastructure-level kill (task stopped mid-job) rather than a specific stage.
+
+- **AWS FIS**: run an experiment with the `aws:ecs:stop-task` action targeting a running worker task
+  while a job is in progress.
+- **Manual equivalent**: find the running task and stop it:
+
+  ```
+  aws ecs list-tasks --cluster <worker-cluster> --service-name <worker-service>
+  aws ecs stop-task --cluster <worker-cluster> --task <task-arn> --reason "crash recovery test"
+  ```
+
+ECS replaces the task; the replacement recovers the in-flight job the same way.
+
+## Watch it in the logs
+
+The crash/recovery lifecycle is logged with greppable markers so you can follow it end to end. In order,
+you should see:
+
+| Marker | What it means |
+| --- | --- |
+| `CRASH-INJECTION ARMED at '<point>'` | the worker started in crash-test mode (logged once at startup) |
+| `CRASH-INJECTION firing at '<point>'` | the crash just happened - the worker is halting (exit 137) |
+| `RECOVERY for job <uuid>: HARD RECOVERY` | a replacement worker fenced the dead one and is redoing incomplete partitions |
+| `finished with status COMPLETED` | the job's batch run completed on the replacement |
+| `assembly for job <uuid>: ... new JobOutput row(s)` | the final output was assembled and delivered |
+
+A quick way to pull the whole story for a job:
+
+```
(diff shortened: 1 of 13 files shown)


worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/PrototypeJobRecovery.java
```diff
@@ -82,10 +82,14 @@ public Ownership acquire(String jobUuid, String owner) {
         Optional<Long> adopted = jobLease.tryAdoptCleanSuspend(jobUuid, owner);
         if (adopted.isPresent()) {
             // The token did not move, so the files and checkpoint keys already line up. Nothing to repair.
+            log.info("RECOVERY for job {}: SOFT RESUME on token {} - clean suspend adopted, nothing to redo",
+                    jobUuid, adopted.get());
             return new Ownership(adopted.get(), true);
         }
 
         long fenceToken = jobLease.bump(jobUuid, owner);
+        log.info("RECOVERY for job {}: HARD RECOVERY on new token {} - a crashed or stale worker was fenced, "
```
keeyanghoreshi-rgb: Technically `acquire` gets run every time a job gets picked up, including the first time. So this will log a hard recovery on job pickup. Honestly, we should probably just tighten up the fresh/soft/hard acquisition because fresh/soft are bleeding into each other a little bit. 

Hard/soft recovery is already logged, these logs (this hard recovery one and the soft resume one above it) could be edits to the existing ones in the processor.

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/BeneficiaryItemReader.java
```diff
@@ -68,6 +71,7 @@ public void open(ExecutionContext executionContext) throws ItemStreamException {
 
     @Override
     public CoverageSummary read() {
+        crashInjector.maybeCrash("read");
```
keeyanghoreshi-rgb: Would be good to make the injection point names `read/write/process/assembly` an enum.

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/CrashInjector.java
```diff
@@ -0,0 +1,48 @@
...
+    private final double crashProbability;
+
+    public CrashInjector(
+            @Value("${pause-resume.prototype.crash-at:process}") String crashAt,
+            @Value("${pause-resume.prototype.crash-probability:0}") double crashProbability) {
+        this.crashAt = crashAt == null ? "" : crashAt.trim();
+        this.crashProbability = crashProbability;
+        if (crashProbability > 0) {
+            // Loud on purpose: if you see this in a deployed worker's logs, it is going to crash itself.
+            log.warn("CRASH-INJECTION ARMED at '{}' with probability {} - this worker will halt itself, "
```
keeyanghoreshi-rgb, line 43: Going off the enum idea, this would be a good place to validate that the property matches a value in the enum

docs/prototype-crash-testing.md
```diff
@@ -0,0 +1,92 @@
+# Prototype crash testing runbook
+
+The pause/resume prototype worker has recovery paths for crashes at each stage of a job
+(partitioning, reading, processing, file writing, assembly). Unit and integration tests cover these
+in-process. This runbook is for the other half: proving a **deployed** worker container recovers from a
+real crash without corrupting or duplicating output.
+
+Crash injection is off by default and is only meant for `dev`/`test`. **Never arm it in prod.**
```
keeyanghoreshi-rgb: The crash injector should check `execution.env` and refuse to arm if it's prod/sandbox.

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/NdjsonCompositeWriter.java
```diff
@@ -48,6 +51,7 @@ public void close() throws ItemStreamException {
 
     @Override
     public void write(@NonNull Chunk<? extends SerializedEobs> chunk) throws Exception {
+        crashInjector.maybeCrash("write");
```
keeyanghoreshi-rgb: The crash is synchronous so this always crashes before we start writing. We'd want to crash mid-write.

docs/prototype-crash-testing.md
```diff
@@ -0,0 +1,96 @@
+# Prototype crash testing runbook
+
+The pause/resume prototype worker has recovery paths for crashes at each stage of a job
+(partitioning, reading, processing, file writing, assembly). Unit and integration tests cover these
```
keeyanghoreshi-rgb: There is no partitioning recovery path, partitioning has no resume

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/CrashInjector.java
```diff
@@ -0,0 +1,68 @@
...
+        this.armed = requested && point.isPresent() && !protectedEnv;
+        if (armed) {
+            log.warn("CRASH-INJECTION ARMED at '{}' with probability {} - this worker will halt itself, "
+                    + "FOR RECOVERY TESTING ONLY", this.crashAt.configValue(), crashProbability);
+        }
+    }
+
+    private static boolean isProtectedEnv(String executionEnv) {
+        String env = executionEnv == null ? "" : executionEnv.toLowerCase();
+        return env.contains("prod") || env.contains("sandbox");
```
keeyanghoreshi-rgb: This blacklists prod/sandbox but it should really whitelist dev/test/local and deny everything else to be safe.

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/NdjsonCompositeWriter.java
```diff
@@ -48,6 +51,7 @@ public void close() throws ItemStreamException {
 
     @Override
     public void write(@NonNull Chunk<? extends SerializedEobs> chunk) throws Exception {
+        crashInjector.maybeCrash(CrashPoint.WRITE);
```
keeyanghoreshi-rgb: This is minor but it would be better to crash the write in between the data write and error write, a few lines down.

worker/src/main/java/gov/cms/ab2d/worker/processor/prototype/PrototypeOutputAssembler.java
```diff
@@ -66,6 +68,8 @@ public void assemble(Job job, String jobUuid, String contractNumber) throws IOEx
             return;
         }
 
+        crashInjector.maybeCrash(CrashPoint.ASSEMBLE);
```
keeyanghoreshi-rgb: I think it would be more interesting to have this crash in the middle of `promoteWinners`, so that the assembly step is left unstable and on restart proves its idempotency.

keeyanghoreshi-rgb (approved):

PR #1865, opened 2026-09-14
AB2D-7435  Investigate why total-benes-served-update workflow is not stable

smirnovaae:
## 🎫 Ticket

https://jira.cms.gov/browse/AB2D-7435

## 🛠 Changes

I rewrote `find_benes_served()` so it scans the coverage data once instead of scanning the same partitions again for each contract. This reduced the runtime from 291 minutes to 31 minutes and returned the same result: 34,421,221.

This fixes the intermittent workflow failures, which happened because the job sometimes ran longer than the runner’s hard limit of about 310 minutes and was killed.

## ℹ️ Context

The total-benes-served-update workflow fails intermittently. We need to find the root cause and fix the procedure.

## 🧪 Validation

Old: https://github.com/CMSgov/ab2d/actions/runs/34560629957
Updated: https://github.com/CMSgov/ab2d/actions/runs/34663760386

### DIFF
--- a/.github/workflows/resources/total-benes-served.sql
+++ b/.github/workflows/resources/total-benes-served.sql
@@ -1,1 +1,1 @@
-SET statement_timeout='7h'; CALL find_benes_served();
+SET statement_timeout='7h'; CALL find_benes_served_optimized();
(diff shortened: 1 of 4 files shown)


bennavapbc (approved):
Nice!

Sadibhatla (approved):

